PR #16, opened 2020-02-28
update deployment strategies of several services to "replace"

opadron:
Changes the deployment strategies of several services from the default "rollout" to "replace".

Necessary for deployments with `replicas=1` and who mount persistent volumes.  With `replace` strategy, the pod running the service is completely removed before being recreated.  This strategy avoids issues where the "replacement" pod in a rollout cannot successfully start because they need resources held by the original pod.

Note that to achieve higher availability and/or zero downtime deployments require that a service be redefined in terms of replicasets and dynamic volumes.  This change only affects services that either cannot be deployed using a rolling strategy (e.g.: databases, because they do not support it) or for which we haven't gone through the trouble of redeploying them in a configuration that would support it (e.g.: gitlab).

### DIFF
--- a/kubernetes/spack/cdash-spack-io/db/deployments.yaml
+++ b/kubernetes/spack/cdash-spack-io/db/deployments.yaml
@@ -13,6 +13,8 @@ spec:
       app: cdash
       svc: db
   replicas: 1
+  strategy:
+    type: Recreate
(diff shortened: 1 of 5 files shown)


zackgalbreath:
LGTM

PR #18, opened 2020-03-05
Make a couple changes to the documentation

scottwittenburg:
Make sure the workflow still works as documented.


scottwittenburg:
Seems I don't have permission to merge, @opadron, can you do it?

PR #21, opened 2020-03-25
Gitlab Runner Communication Hardening

opadron:


kubernetes/spack/gitlab-spack-io/runner/medium/deployments.yaml
```diff
@@ -59,6 +59,7 @@ spec:
 
         - name: CI_SERVER_URL
           value: "http://gitlab"
+          # value: "https://gitlab.spack.io"
```
scottwittenburg: Should we have large runners use the internal interface as well?

opadron: Yes, and they already should be.  The medium runners were the only ones we had experimented with going through the public interface.

opadron:
@scottwittenburg @zackgalbreath ready for review.

PR #24, opened 2020-04-09
Use version that supports dynamic child pipelines

scottwittenburg:


scottwittenburg:
@opadron @zackgalbreath Is it this simple to upgrade to the version that will allow us to move to dynamic child pipelines and get rid of the need for pushing the generated yaml?

PR #28, opened 2020-05-06
Use Curator to delete old ElasticSearch indices

zackgalbreath:


kubernetes/spack/kibana-spack-io/curator/config-maps.yaml
```diff
@@ -0,0 +1,60 @@
...
+    actions:
+      1:
+        action: delete_indices
+        description: "Clean up ES by deleting old indices"
+        options:
+          timeout_override:
+          continue_if_exception: False
+          disable_action: False
+          ignore_empty_list: True
+        filters:
```
opadron, line 28: Let's apply this action only to the logstash indexes.  The new data from Gitlab shouldn't grow over time in the same way, and I think we should be OK with keeping it all.

```YAML
      # try adding this filter
      - filtertype: pattern
        kind: prefix
        value: logstash-
        exclude:
```

zackgalbreath: 👍

PR #29, opened 2020-05-19
Better elastic search integration

opadron:
- Configures Gitlab service to ensure the presence of the `gitlab-*` elasticsearch indexes.
 - Removes the debugging arguments in our medium runner.
 - Increases large runners to three (3) replicas.
 - Adds a new set of t3.micro runners, currently set to five (5) replicas.  These new runners include an experimental elasticsearch integration implementation.

The details of the ES implementation are in the new `gitlab-runner-scripts` ConfigMap.  Additional mounts can't be specified for pods created by the Kubernetes executor, so the implementation works around this limitation by generating pre-build and post-build scripts, which are passed along to runners verbatim when registering.  Unfortunately, this additional level of indirection can make the script logic very tough to follow.  The general process is as follows:

First the runners (through these generated scripts) instruct the job pods to create a python script, `/generate-payload`.  This script creates two payloads, one for the record representing the build being performed (`./payload.json`) and one for specifying type metadata for that record (`./index.json`).  These payloads are populated using the environment variables of the running build job.  The pods then run the python script, and submit the payload contents as new records to the specified elasticsearch cluster using curl.

The results of this process are continuously updating elasticsearch indexes: `spack-pre-build-YYYY.MM.DD` with pre-build records for all Spack jobs that started on `MM/DD/YYYY`, and `spack-post-build-YYYY.MM.DD` with post-build records (*also*) for all Spack jobs that started on `MM/DD/YYYY` (i.e.: the record, itself might have a timestamp that falls on a later day if the job runtime spans multiple days).  To be clear: every job that starts on a day will have a pre- and post-build record in the respective indexes for *that* day, even if the job's actual end time falls on a later day.

On Kibana, three new index patterns have been created, `spack-pre-build-*`, `spack-post-build-*`, and `spack-*-build-*`, against which queries may be performed to search through pre-build records, post-build records, or both, respectively.

Example queries:
 - All records for pre-ci jobs (execute against the `spack-*-build-*` index pattern):

```
job.name: generate-pipeline
```
[link](https://kibana.spack.io/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2d,to:now))&_a=(columns:!(_source),filters:!(),index:'5b7b9f20-9b87-11ea-97a2-05d755153fd0',interval:auto,query:(language:kuery,query:'job.name:%20generate-pipeline'),sort:!(!('@timestamp',desc))))

 - Post build records for ncurses builds (execute against the `spack-post-build-*` index pattern):

```
spack.spec_pkg_name: ncurses
```
[link](https://kibana.spack.io/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2d,to:now))&_a=(columns:!(_source),filters:!(),index:be4b8e40-9b90-11ea-97a2-05d755153fd0,interval:auto,query:(language:kuery,query:'spack.spec_pkg_name:%20ncurses'),sort:!(!('@timestamp',desc))))

 - Same as above, but limit only to failed builds.

```
(spack.spec_pkg_name: ncurses) AND (NOT result: 0)
```
[link](https://kibana.spack.io/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2d,to:now))&_a=(columns:!(_source),filters:!(),index:be4b8e40-9b90-11ea-97a2-05d755153fd0,interval:auto,query:(language:kuery,query:'(spack.spec_pkg_name:%20ncurses)%20AND%20(NOT%20result:%200)'),sort:!(!('@timestamp',desc))))

For now, the integration is limited to the new `t3.micro` runners. I plan to follow this PR up with another that extends it to the other runners.

#### Known Issues

 - Fields with empty string values need better handling, as the current implementation makes it impossible to distinguish them from non-empty string values in queries.

 - The new indexes do not include any log information, only "job-started" and "job-ended" records.  This work can be extended to collect other information (such as any generated job logs) and add them to elasticsearch.

 - The current implementation might not correctly submit post-build records in the event of a job failure.  Additional testing and development may be needed.

### DIFF
--- a/kubernetes/spack/gitlab-spack-io/runner/config-maps.yaml
+++ b/kubernetes/spack/gitlab-spack-io/runner/config-maps.yaml
@@ -1,0 +1,248 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gitlab-runner-scripts
+  namespace: spack
+  labels:
+    app: gitlab
+    svc: runner
+data:
+  generate-pre-build: |
+    cat << EOF
+    timestamp="\$( date '+@%s' )"
+    index="spack-pre-build-\$( date --date "\$timestamp" '+%Y.%m.%d' )"
+    datetime="\$( date --date "\$timestamp" '+%Y-%m-%dT%H:%M:%S.%N%z' )"
+
+    (
+      cat << __PYTHON_EOF
+    $( cat generate-payload )
+    __PYTHON_EOF
+    ) > /generate-payload
+
+    if   which python3 ; then py=python3
+    elif which python  ; then py=python
+    else                      py=python2
+    fi &> /dev/null
+
+    export datetime
+
+    rm -f ./index.json ./payload.json
+    "\$py" /generate-payload
+    payload_index="\$( cat ./index.json )"
+    payload="\$( cat ./payload.json )"
+
+    curl -XPUT \
+      --output /dev/null \
+      -u "${ELASTIC_SEARCH_USER}:${ELASTIC_SEARCH_PASS}" \
+      "${ELASTIC_SEARCH_URL}/\${index}" \
+       -H 'Content-Type: application/json' \
+       -d"\$payload_index" &> /dev/null
+
+    curl -XPOST \
+      --output /dev/null \
+      -u "${ELASTIC_SEARCH_USER}:${ELASTIC_SEARCH_PASS}" \
+      "${ELASTIC_SEARCH_URL}/\${index}/_doc" \
+       -H 'Content-Type: application/json' \
+       -d"\$payload" &> /dev/null
+    EOF
+
+  generate-post-build: |
+    cat << EOF
+    result="\$?"
+    export result
+
+    end_timestamp="\$( date '+@%s' )"
+    index="spack-post-build-\$( date --date "\$timestamp" '+%Y.%m.%d' )"
+    end_datetime="\$( date --date "\$end_timestamp" '+%Y-%m-%dT%H:%M:%S.%N%z' )"
+
+    (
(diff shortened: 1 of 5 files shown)


opadron:
@scottwittenburg I forgot to mention that, as a side product of this work, I also created this project on our Gitlab: https://gitlab.spack.io/opadron/spack-pipelines.  This is what I've been using as a test project that mapped jobs to `t3.micro` instances.  I also wanted to avoid disrupting any other active pipelines with my testing.

And as an aside to this aside: Just as we're mapping especially expensive build jobs to the larger instances, I'm thinking we can possibly find more cost savings by mapping especially cheap build jobs to these smaller ones.  I wanted to see if going smaller was feasible for some packages, so this exercise was partially motivated by that question.

scottwittenburg:
> @scottwittenburg I forgot to mention that, as a side product of this work, I also created this project on our Gitlab: https://gitlab.spack.io/opadron/spack-pipelines. This is what I've been using as a test project that mapped jobs to `t3.micro` instances. I also wanted to avoid disrupting any other active pipelines with my testing.

Ah yes, that was a good idea, thanks for the explanation.  Maybe I can do the same thing as I'm evaluating changes I plan to make to spack pipelines.

> And as an aside to this aside: Just as we're mapping especially expensive build jobs to the larger instances, I'm thinking we can possibly find more cost savings by mapping especially cheap build jobs to these smaller ones. I wanted to see if going smaller was feasible for some packages, so this exercise was partially motivated by that question.

Yes, good call.  I'm not sure yet if there's a way for us to know which are the cheap builds, but this sounds like something we should aim for.

PR #31, opened 2020-06-08
gh-gl-sync: add initial testing version

opadron:
Adds a new internal service (based on the work of @chuckatkins) that periodically syncs opened PRs on a github repo to branches on a Gitlab project.

TODO:

 - [x] Confirm correct operation on test repos.
 - [ ] Switch to main Spack repos.
 - [ ] (optional): Add status posting.
 - [ ] (optional): Add a webhook for continuous operation.

Optional items might get put off to their own PRs.


kubernetes/spack/internal-spack-io/gh-gl-sync/config-maps.yaml
```diff
@@ -0,0 +1,218 @@
...
+    then
+      CLOSED_REFSPECS=""
+      for PR in ${SYNCD_CLOSED_PR_BRANCHES}
+      do
+        echo "Adding respec for closed ${PR}"
+        CLOSED_REFSPECS="${CLOSED_REFSPECS} :github/${PR}"
+      done
+    fi
+
+    # Sync open PRs to OLCF
```
scottwittenburg, line 188: `OLCF` is probably a copy-paste remnant

chuckatkins: Not a copy paste, but a leftover from before I generalized it.  Still though, should just be `# Sync open GitHub PRs to GitLab`

kubernetes/spack/internal-spack-io/gh-gl-sync/cron-jobs.yaml
```diff
@@ -0,0 +1,34 @@
...
+            - name: scripts
+              configMap:
+                name: gh-gl-sync
+                defaultMode: 0700
+            - name: secrets
+              secret:
+                secretName: gh-gl-sync
+                defaultMode: 0700
+          nodeSelector:
+            "beta.kubernetes.io/instance-type": "t2.medium"
```
chuckatkins, line 34: This has almost no resource requirements and should be fine even on a `t2.nano`

kubernetes/spack/internal-spack-io/gh-gl-sync/cron-jobs.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: gh-gl-sync
+  namespace: spack
+spec:
+  schedule: "*/1 * * * *"
```
chuckatkins, line 8: This may take a bit of experimentation but every minute may be too frequent for spack given the sheer volume of PRs.

kubernetes/spack/internal-spack-io/gh-gl-sync/config-maps.yaml
```diff
@@ -0,0 +1,218 @@
...
+    except OSError:
+        sys.exit(1)
+
+    prs = json.loads(response.read())
+
+    for pr in prs:
+        print("%d %s" % (pr['number'], pr['head']['ref']))
+
+  github-prs-to-gitlab.sh: |
+    #!/usr/bin/env bash
```
chuckatkins, line 65: I don't actually know if this is bash specific or not.  It may be generic `sh` compliant, I haven't tried it though.

kubernetes/spack/internal-spack-io/gh-gl-sync/config-maps.yaml
```diff
@@ -0,0 +1,218 @@
...
+    TMP_WORKDIR=$(mktemp -d)
+    CLEANUP_WORKDIR=1
+    echo "Initializing local repo in ${TMP_WORKDIR}"
+    cd ${TMP_WORKDIR}
+    git init
+    git remote add github https://github.com/${GITHUB_REPO}.git
+    git remote add gitlab ${GITLAB_URL}
+
+    # Retrieve open PRs
+    OPEN_PR_BRANCHES="$(python3 ${BASEDIR}/github-list-prs.py -o ${GITHUB_REPO} --token ${GITHUB_TOKEN} | awk '{print "pr"$1"_"$2}' | sort -r)"
```
chuckatkins, line 129: IIRC the repo is a positional argument and needs to be specified last, i.e. `github-list-prs.py -t ${GITHUB_TOKEN} -o ${GITHUB_REPO}`.  The use of the token is also entirely optional.  It's helpful to get around API rate limits from a given IP but since this is being run in a cloud container I don't this that's really a concern is it?

kubernetes/spack/internal-spack-io/gh-gl-sync/cron-jobs.yaml
zackgalbreath: Good point, Chuck. I think Todd initially asked for 1 minute resolution so people don't get impatient waiting for their PRs to get tested. If you're feeling brave Omar, I'm okay with trying this as-is and scaling it back if it proves to be problematic. 

Medium term, I believe the plan is to migrate this functionality to a GitHub app. At that point the CronJob would go away and the polling interval would be irrelevant.

kubernetes/spack/internal-spack-io/gh-gl-sync/cron-jobs.yaml
zackgalbreath: Good point. Correct me if I'm wrong Omar, but I believe our "general" pool is t2.medium. So since we have these instances sitting around anyway, it probably doesn't hurt to run a lightweight cronjob there.

zackgalbreath:
I did some more manual experimentation with this. I can confirm that it is correctly mirroring GitHub PRs to GitLab branches, and that this triggers a GitLab CI job when the .gitlab-ci.yml file is present.

PR #33, opened 2020-06-18
Convert gh-gl-sync code to Python

zackgalbreath:


zackgalbreath:
@opadron Here's my WIP. The structure of this class may or may not change dramatically as I add tests, but hopefully it at least gives you an idea of the direction I'm headed.

zackgalbreath:
This is ready for you guys to take a look at @opadron @scottwittenburg

You'll probably notice right away that it's broken up into lots of little functions. This was so I could write tests for the majority of the code without having to actually talk to GitHub or GitLab.

We will still need to setup CI testing for this stuff as we continue to change it. GitLab CI is obviously a good fit, but then we would need a way to sync it from GitHub 😉

PR #36, opened 2020-07-15
update specs for kubernetes 1.15

opadron:
Also adds the following changes:

 - Upgrades cluster autoscaler from v1.12.3 to v1.14.7
 - Changes the instance type for the instance running Gitlab from `r5.xlarge` to `r5.4xlarge`.


opadron:
Edited title to reflect that we *actually* updated to v1.16.

PR #46, opened 2020-11-16
Cleanup pr mirrors

scottwittenburg:


scottwittenburg:
@zackgalbreath Now I just need to add those secrets to the cluster.  Otherwise, I've tested this change locally to make sure thay it actually does delete the mirrors when PRs are merged.

tgamblin (approved):

PR #48, opened 2020-11-18
mitigate docker ratelimit issues

opadron:
- Added a cluster-internal docker registry, accessible under the image tag prefix `k8s.internal/`.
 - Switched multiple pull policies from `Always` to `IfNotPresent`.
 - Switched multiple pod templates to pull from the internal registry instead of dockerhub.
 - Reduced the frequency of the spack auto-updater from once very five minutes to once every six hours.
 - Added convenience scripts for common tasks related to internal registry maintenance.


scottwittenburg (approved):
LGTM Thanks for getting this working!

PR #50, opened 2020-12-01
Temporarily mitigate docker apocalypse using ghcr

scottwittenburg:
Use images pushed to GitHub container registry personal account to temporarily work around recent dockerhub pull rate limits.

Supersedes #48


zackgalbreath (approved):

opadron (approved):
Changes applied.  We still need to run test jobs to make sure they're pulling the new images, but otherwise LGTM. :+1:

PR #52, opened 2020-12-08
Handle result pagination

scottwittenburg:
Changed all four github api requests to use PyGithub, as it provides a much simpler way to handle pagination (compared to doing it ourselves by reading the "Link" headers from the response).

It seems gitlab also paginates its api results, see [here](https://docs.gitlab.com/12.10/ee/api/README.html#pagination).  However, since we only query for pipelines associated with a single branch at a time, and since we further restrict the time range to the last few minutes, I didn't make any effort to handle pagination for the gitlab api requests.  It looks like the default is for gitlab to return 20 items per page, which seems like it might be fine for our purposes here.

### DIFF
--- a/gh-gl-sync/SpackCIBridge.py
+++ b/gh-gl-sync/SpackCIBridge.py
@@ -24,7 +24,7 @@ class SpackCIBridge(object):
         self.github_repo = ""
 
         self.py_github = Github(os.environ.get('GITHUB_TOKEN'))
-        self.py_gh_repo = self.py_github.get_repo('spack/spack')
+        self.py_gh_repo = self.py_github.get_repo('spack/spack', lazy=True)
 
     @atexit.register
     def cleanup():
@@ -243,8 +243,6 @@ class SpackCIBridge(object):
 
         post_data["target_url"] = pipeline["web_url"]
         post_data["context"] = "ci/gitlab-ci"
-
-        post_data = json.dumps(post_data).encode('utf-8')
         return post_data
 
     def dedupe_pipelines(self, api_response):
--- a/gh-gl-sync/SpackCIBridge.py
+++ b/gh-gl-sync/SpackCIBridge.py
@@ -6,6 +6,7 @@ import base64
 import boto3
 from datetime import datetime, timedelta, timezone
 import dateutil.parser
+from github import Github
 import json
 import os
 import re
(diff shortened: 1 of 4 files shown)


scottwittenburg:
Eh, using the `PyGithub` library module is making it harder to unit test the sync script, so I just removed one of the unit tests (I found that particular test to have limited value anyway).   But I guess I still broke something and will have to go figure out what it was and fix it.

zackgalbreath:
Thanks for working on this, Scott. It's looking pretty good to me so far. I'll take a stab at updating the relevant test cases today if you don't beat me to it.

zackgalbreath:
@scottwittenburg this is ready to merge & deploy if you're happy with my latest changes

scottwittenburg:
I ran into problems when trying to update the sync script to run in the existing directory we set up as a persistent volume claim.  It's going to take me longer than I wanted to debug it and make sure it's working correctly in that mode.

So in the meantime, what's in this branch is what's currently running on the cluster (every 6 minutes, even though it seems after the first push of all those missing branches, it's running much faster).

At least this way we'll have some PR testing going on until I can get the sync to work from an existing repo with saved state.

/cc @zackgalbreath @opadron

scottwittenburg:
So I was going to suggest we just merge this, wait for the container to get built, then apply the master version of the cronjob to the cluster.

But now that I'm thinking of it, why isn't this driving our dockerhub pull numbers over the top?  Especially since it runs so frequently?  I know we have the "if not present" pull policy on that cron job, but that doesn't help in other cases.

zackgalbreath:
> why isn't this driving our dockerhub pull numbers over the top?

Good question. The cron job runs on our "base pool" of t2.mediums, which aren't spun up & torn down by Cluster Autoscaler. My understanding of "IfNotPresent" is that we pull the image once and it stays locally cached on the node. So each time we bump the version we might have to pull the image N times, but after that it shouldn't count against our rate limit.

scottwittenburg:
Right, that makes sense, thanks.

So what do you think, should I merge it and then apply the change to the cluster?  The only change I expect it would result in, compared to what's running on the cluster now, is it will run every 2 minutes instead of every 6.

zackgalbreath:
Sounds like a good idea to me

PR #55, opened 2020-12-18
Mirror <merge_commit_sha> rather than "refs/pulls/<pr_number>/head"

scottwittenburg:
This is designed to fix the problem we discovered today where by the time the E4S pipeline runs, we might merge a much more recent `github/develop` into the PR branch than existed when the branch was committed/pushed to the spack github repo.  If we go with this, we will then drop [this part](https://gitlab.spack.io/spack/e4s/-/blob/master/.gitlab-ci.yml#L16) of our custom `before_script`.

GitHub generates a "test commit" by merging the head of the PR branch into the base (at some point).  If that commit comes back "None" in the api query for open pulls, it is likely because it cannot be merged with the base any longer, so we skip it.

Testing against a local standin for the target gitlab remote, I can look at the branches this is now pushing and see the kind of commit message GitHub is generating (which is what I expect we'll see in our spack pipelines on gitlab):

```
$ git log -n2 github/pr20463_slepc-set-arpack-dir-right
commit 82968278647d53704c2c6ce0b696a4d4ec7aaa8d (github/pr20463_slepc-set-arpack-dir-right)
Merge: 13e4d9ee3a 320b0a26cf
Author: eugeneswalker <38933153+eugeneswalker@users.noreply.github.com>
Date:   Fri Dec 18 04:48:27 2020 +0000

    Merge 320b0a26cf13d1d8988d4a642f179e7d4f1b493d into 13e4d9ee3a7eb324db6377e8f4cebb42285f7312

commit 320b0a26cf13d1d8988d4a642f179e7d4f1b493d
Author: eugeneswalker <eugenesunsetwalker@gmail.com>
Date:   Fri Dec 18 04:27:58 2020 +0000

    slepc: set --with-arpack-dir correctly
```

So the author will be correct, but we would no longer see the commit message on the latest PR commit as we do currently.

### DIFF
--- a/.github/workflows/spack_ci_bridge_docker.yml
+++ b/.github/workflows/spack_ci_bridge_docker.yml
@@ -28,7 +28,7 @@ jobs:
           context: ./gh-gl-sync
           file: ./gh-gl-sync/Dockerfile
           push: true
-          tags: zackgalbreath/spack-ci-bridge:0.0.7
+          tags: zackgalbreath/spack-ci-bridge:0.0.8
       -
         name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/gh-gl-sync/SpackCIBridge.py
+++ b/gh-gl-sync/SpackCIBridge.py
@@ -71,16 +71,24 @@ class SpackCIBridge(object):
         """ Return a list of strings in the format: "pr<PR#>_<headref>"
         for GitHub PRs with a given state: open, closed, or all.
         """
-        pr_strings = []
+        pr_dict = {}
         pulls = self.py_gh_repo.get_pulls(state=state)
         for pull in pulls:
+            if not pull.merge_commit_sha:
+                print("PR {0} ({1}) has no 'merge_commit_sha', skipping".format(pull.number, pull.head.ref))
+                continue
             pr_string = "pr{0}_{1}".format(pull.number, pull.head.ref)
-            pr_strings.append(pr_string)
-        pr_strings = sorted(pr_strings)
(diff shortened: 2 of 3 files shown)


zackgalbreath (approved):
This looks good to me. Thanks for working on it, Scott!

One question: do you think it's worth posting a failed status for merge conflicts, or are we okay letting GitHub point that out for us?

scottwittenburg:
> This looks good to me. Thanks for working on it, Scott!

No problem, thanks for jumping in and fixing the tests! 

> One question: do you think it's worth posting a failed status for merge conflicts, or are we okay letting GitHub point that out for us?

Yeah, that's a good idea.  Let me take a look at how to work that in.

opadron:
> So the author will be correct, but we would no longer see the commit message on the latest PR commit as we do currently.

How about amending the commit before pushing it?

scottwittenburg:
> How about amending the commit before pushing it?

The pull requests response from the github api seems to have the PR title in it, so that could be an option.  In order to get the message from the most recent commit, I think we would have to make a separate api request per PR?  Which will certainly make the sync run longer, I'm not sure how much longer.

Amending the commit to include the title will add to the runtime of the sync as well, and make us treat open prs and protected branches differently, but I can certainly take a stab at it if it makes sense.  I think I'd want to keep the merge commit message generated by github, and just add the title somehow, which I'm not sure yet how to do, but there's probably a way.

scottwittenburg:
After giving it a little more thought, I propose we leave amending the commit for a follow-on improvement we can make if it becomes a high priority in the future.

PR #57, opened 2021-01-05
Increase minio pvc size llnl

scottwittenburg:
Looking at the StorageClass associated with the persistent volume claims in the new cluster, it seems we may have dynamic volume provisioning as described [here](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/).  So this change may be enough to cause Kubernetes to create another volume of 100GB on behalf of gitlabs minio.

Not sure how to apply this change, given that it's in a helm chart and supposed to be managed by flux.

### DIFF
--- a/k8s/gitlab/temporary/release.yaml
+++ b/k8s/gitlab/temporary/release.yaml
@@ -106,6 +106,8 @@ spec:
           secretName: tls-gitlab-minio-temp
       nodeSelector:
         spack.io/node-pool: gitlab
+      persistence:
+        size: 100Gi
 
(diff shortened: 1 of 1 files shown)


opadron:
Can we redo this pull request against the `llnl` branch?  Once merged, I'll push it into production.

opadron (changes requested):
LGTM, other than the target branch.

scottwittenburg:
> LGTM, other than the target branch.

Thanks @opadron, I changed the target branch.

opadron (approved):

PR #60, opened 2021-03-04
Add a script to gather statistics on pipelines/jobs

scottwittenburg:


scottwittenburg:
Still a couple issues to sort out, but I ran this for the period last week between roughly Tuesday evening and Friday afternoon to produce the following output:

```
Summary

  Total pipelines: 1070
    failed 196 (18.32%)
    canceled 466 (43.55%)
    success 408 (38.13%)
  Total jobs: 40255
    failed 481 (1.19%)
    success 8376 (20.81%)
    canceled 30751 (76.39%)
    skipped 647 (1.61%)
  Failure categories:
    BROKEN_SPACK 464 (96.47%)
    TCP_CONNECTION_REFUSED 12 (2.49%)
    HELPER_CONTAINER_NOT_FOUND 1 (0.21%)
    POD_NOT_FOUND 3 (0.62%)
    UNKNOWN 1 (0.21%)
  The following jobs had unrecognized failures:
    https://gitlab.next.spack.io/spack/spack/-/jobs/95788

Finshed after 1741.716237 seconds
```

zackgalbreath:
Thanks for putting this together. I love seeing Real Data™ about the causes of our pipeline failures. Looks like `TCP connection refused` is the biggest culprit on the infrastructure side so far.

tgamblin:
That looks great to me!  I think this will be really helpful going forward, and will help to prioritize work on the infrastructure.

Questions (maybe for future PRs):
1. Any insights into the 3 error types here?  Is the failed TCP connection even visible on the server side?
2. Is it possible to get the total running time of the pipelines and/or jobs?  It would be good to know how much time we spend on failed and canceled pipelines vs. successful ones.
3. The total pipelines all have statuses -- so I'm assuming this does not include running pipelines -- just ones that have completed, right?

scottwittenburg:
> That looks great to me! I think this will be really helpful going forward, and will help to prioritize work on the infrastructure.

Yes, totally agree.

> Questions (maybe for future PRs):
> 
> 1. Any insights into the 3 error types here?  Is the failed TCP connection even visible on the server side?

I'm not sure yet.  Since we don't yet have all the log harvesting/gathering we had in the old cluster (not to mention the kibana / elastic search set up to comb through those logs), the process for searching the server side for information is fairly primitive.

> 2. Is it possible to get the total running time of the pipelines and/or jobs?  It would be good to know how much time we spend on failed and canceled pipelines vs. successful ones.

I think the more raw data I'm logging into that JSON object at the end may actually contain running time for jobs and pipelines, I'll take a look.

> 3. The total pipelines all have statuses -- so I'm assuming this does not include running pipelines -- just ones that have completed, right?

Well, `running` is a state I've seen on some jobs/pipelines I got back from the api.  I think in this case I ran it on a time range that didn't have any running pipelines, or we would have seen some of those in the summary.  Let me know if that didn't quite answer your question though

tgamblin:
> I think in this case I ran it on a time range that didn't have any running pipelines, or we would have seen some of those in the summary. Let me know if that didn't quite answer your question though

This is great -- it actually does more than I was asking.

zackgalbreath (approved):
Nice find on the "pipeline bridges" API endpoint

scottwittenburg:
> Nice find on the "pipeline bridges" API endpoint

Thanks for your review!

PR #62, opened 2021-03-09
Add sync to new infrastructure

scottwittenburg:


scottwittenburg:
@zackgalbreath Thanks for the tip on getting newlines out of the secrets, that did the trick.  So now what is in this branch is what is running currently on the new cluster.

scottwittenburg:
@opadron @zackgalbreath I just looked around for the extra bits like the scraping script itself and the Dockerfile, and it seems that was a PR (#60) we already merged into master.

So maybe the question now is: Can we merge `master` into `llnl` to get these changes in our current "staging" branch? 

The one problem that may need to be addressed is that the gh workflows that automatically build the images used in the cronjobs added in this PR require a merge to master to be triggered (likely why I made #60 against `master` instead of `llnl` in the first place).  Does it make sense to add that change to this PR?

opadron:
Can we change the workflow to trigger off `llnl`, at least until such time that it becomes the "new" `master`?

scottwittenburg:
> Can we change the workflow to trigger off `llnl`, at least until such time that it becomes the "new" `master`?

Looking back, I wasn't very clear, but that's what I meant when I said:

> Does it make sense to add that change to this PR?

zackgalbreath:
Sounds like a good idea to me too

opadron:
Oh, I understand, now.  Yes, I agree that this is the right thing to do.

scottwittenburg:
I can make that change in this PR, however, `llnl` is missing stuff from `master` that is required to build the images used by the cronjobs added here.

So what about attempting to merge `master` into `llnl`, and then I'll rebase this PR off the resulting `llnl` and change the workflow to trigger off `llnl`?  Or is there a better approach, or might that not work for some reason?

opadron:
That sounds good to me. #66

zackgalbreath:
I'll update the branch GitHub Actions is watching from `master` to `llnl` in a separate PR

PR #63, opened 2021-04-02
Upgrade GitLab to 13.8.7

zackgalbreath:


tgamblin:
@zackgalbreath looks like it was pretty easy -- is this it?

scottwittenburg:
> @zackgalbreath looks like it was pretty easy -- is this it?

Nope that didn't do it.  I forgot we had to use slightly customized gitlab images, so I'm building and pushing those, then I'll make another PR here and see if that's all we need.

opadron:
Hey, just a heads up, this got pushed directly into `production`, instead of to the `llnl` branch.

I understand that the current workflow is a bit of a pain, but please take care to make sure that new infrastructure PRs are made against the `llnl` branch (until I complete the migration).

opadron:
For the record, I ended up merging `production` into `llnl`, and then force-pushing `llnl` back to `production`.  (In case anyone going through the git history is wondering what the heck happened and lands on this PR!)

PR #70, opened 2021-05-07
Use Kubernetes service account for gitlab-api-scrape

zackgalbreath:
Migrate from IAM credentials to IAM role + kube serviceAccount.
Also change the S3 we write logs too, and version bump the relevant
Docker image.

### DIFF
--- a/scripts/gitlab_api/gitlab_api.py
+++ b/scripts/gitlab_api/gitlab_api.py
@@ -334,15 +334,12 @@ e.g. '2019-03-15T08:00:00Z').  If none is provided, the default is 24 hrs before
         print(read_back.read())
 
     if post_summary:
-        if "AWS_ACCESS_KEY_ID" in os.environ and "AWS_SECRET_ACCESS_KEY" in os.environ:
-            now = datetime.now()
-            now_dir = now.strftime('%Y/%m')
-            now_name = now.strftime('%Y_%m_%d_%H_%M')
-            object_name = 'pipeline-statistics/' + now_dir + '/daily_summary_{0}.txt'.format(now_name)
-            s3_client = boto3.client('s3')
-            try:
-                response = s3_client.upload_file('output.txt', 'spack-binaries-develop', object_name)
-            except ClientError as e:
-                print(e)
-        else:
-            print('\n *** WARNING: Missing credentials to write to S3 bucket *** \n')
+        now = datetime.now()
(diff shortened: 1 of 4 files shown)


scripts/gitlab_api/gitlab_api.py
```diff
@@ -334,15 +334,12 @@ def increment_failure_category(category):
...
-                print(e)
-        else:
-            print('\n *** WARNING: Missing credentials to write to S3 bucket *** \n')
+        now = datetime.now()
+        now_dir = now.strftime('%Y/%m')
+        now_name = now.strftime('%Y_%m_%d_%H_%M')
+        object_name = 'pipeline-statistics/' + now_dir + '/daily_summary_{0}.txt'.format(now_name)
+        s3_client = boto3.client('s3')
+        try:
+            response = s3_client.upload_file('output.txt', 'llnl-aws-39-logs', object_name)
```
scottwittenburg, line 343: I see that bucket was made last June, do you know who made it or what it's supposed to be for?  It's not in the same region as everything else we use, does that matter?

scottwittenburg (commented):
Thanks @zackgalbreath, looks pretty straightforward.  I just have the the one question about the bucket where we're putting the statistics.

scripts/gitlab_api/gitlab_api.py
zackgalbreath: Good question! No, I'm not sure what this bucket is about. I reused it because its name made it sound like an appropriate location for what we're trying to do here. I'm happy to consider a different location if you or Omar think that's a good idea.

zackgalbreath: I think you might be right. If I'm reading this correctly, it's free to write to S3 from an EC2 instance in the same region.


From the [S3 pricing docs](https://aws.amazon.com/s3/pricing/):

```
You pay for all bandwidth into and out of Amazon S3, except for the following:
Data transferred in from the internet.
Data transferred between S3 buckets. 
Data transferred from an Amazon S3 bucket to any AWS service(s) within the same AWS Region as the S3 bucket (including to a different account in the same AWS region).
```

scottwittenburg: I see you made a new bucket in us east virginia, looks good, thanks!

scottwittenburg (approved):
Awesome, thanks @zackgalbreath!

PR #72, opened 2021-05-07
Update production branch from llnl

zackgalbreath:


zackgalbreath:
Is this the preferred way to let flux know what we've been up to recently @opadron ?

zackgalbreath:
> To me this feels like something we would want to merge when we have everybody around to keep an eye on any potential aftermath. Maybe I'm just paranoid though.

totally agree

opadron:
> Is this the preferred way to let flux know what we've been up to recently @opadron ?

Yes, exactly!  Unfortunately, if we want to test-drive changes before merging, we still have to do so manually, but staging and canary deployments are something we can add at some later point in time.

opadron (approved):

PR #74, opened 2021-05-21
migrate CDash to the new cluster

zackgalbreath:


k8s/cdash/upgrade-db-schema.yaml
```diff
@@ -0,0 +1,84 @@
+---
```
zackgalbreath, line 1: @opadron any thoughts on how we might encapsulate this with deployments.yaml? All this copy/paste seems less than ideal.

opadron: Not really, our best bet would probably be whipping up our own helm chart, which might actually be worth doing at some point.  Having a turn-key CDash solution for kubernetes would be pretty nice.

opadron (approved):

PR #78, opened 2021-05-26
Runners add x86 64 tags

scottwittenburg:


opadron (approved):

scottwittenburg:
@opadron It looks like not all PRs to `llnl` are immediately followed up with another PR from `llnl` to `production`.  I assume sometimes we're just expecting the changes will go into `production` later on as part of a larger batch?  In general do you recommend just getting stuff into `llnl` and waiting for it to get picked up (for example) by you when you're bringing changes of your own to `production`?  Or should we aim to immediately do another PR to make our changes live on the cluster?

opadron:
That part of the process still needs to be improved upon.  Right now, it's all very manual.  I've been pushing updates to `production` on a mostly ad-hoc basis, but I'd love to automate this step.

For now, let's make a separate PR for pushing `llnl` -> `production` for each update.  They don't have to be 1:1 with PRs updating `llnl`, but ideally we'd have at most one opened at any given time.

scottwittenburg:
> For now, let's make a separate PR for pushing `llnl` -> `production` for each update. They don't have to be 1:1 with PRs updating `llnl`, but ideally we'd have at most one opened at any given time.

That sounds reasonable, thanks.  Since there aren't any others open for `llnl` -> `production` at the moment, I'll make one to bring these tags into `producrtion`.

PR #80, opened 2021-06-08
Update gitlab 13.8.8

scottwittenburg:
I'm proposing we may no longer need the custom images as those were our means to increase both the yaml size limit and the per-job needs limit, and I don't think we need them anymore.  

For one thing the lines of ruby we attempted to tweak to increase the needs limit no longer exist in that file, or anywhere.  Furthermore, they didn't exist in the `v13.8.7` images we created either, and so we have been running without the benefit of the increased needs limit since we updated to `v13.8.7`.

As for the yaml size limit, while that tweak still looks to be valid, our generated yaml files are so much smaller now that we no longer encode root spec yamls and write them into the pipeline yaml repeatedly, that I'm not sure we really need it anymore.  When we turned on the full E4S stack in spack PR testing, the generated yaml was around 550KB and the default yaml size limit is 1MB.

But this is a draft so we can discuss the proposal before merging.


zackgalbreath (approved):
Looks reasonable to me, I'd be willing to give this a shot. This kind of thing really makes me wish we had a canary deployment, though... I guess if we deploy this change and it breaks everything, it should be pretty easy to roll back? 🤷 🤞

opadron (approved):
LGTM.  Glad to see that we can simplify the deployment.

scottwittenburg:
Bad news.  This [PR](https://github.com/spack/spack/pull/23922) generated a pipeline yaml of only 408 KB and gitlab now says it is invalid because it's too large.   The tweak we were making looked to me to be changing the limit from 1MB to 1GB, but maybe not?  Or maybe there's some other factor?  I'm looking into it.

PR #83, opened 2021-06-09
Add HelmRelease for one new pool of graviton runners (c6g.4xlarge)

scottwittenburg:
In addition to the CloudFormation stack we created earlier today, this should give us one of the pools of graviton2 runners we need.  I `kubectl apply -f`'d  the release file so I could test it out, seems to be working.  See the pipeline here: 

https://gitlab.spack.io/scott/test-downstream/-/pipelines/43281

That's a private repo, but I added the reviewers as members, except Omar who doesn't seem to have an account on that instance.

### DIFF
--- a/k8s/runners/large-pub/release.yaml
+++ b/k8s/runners/large-pub/release.yaml
@@ -92,6 +92,7 @@ spec:
 
       nodeSelector:
         spack.io/node-pool: glr-large-pub  # pool for jobs
+        kubernetes.io/arch: amd64
 
     nodeSelector:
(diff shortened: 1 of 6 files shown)


scottwittenburg:
FYI @tgamblin

opadron (approved):

opadron:
NOTE: I think this got closed because of my mistake where I accidentally deleted the `llnl` branch the other day.  (This tells me that a better gitops process is more important.  We should prioritize this.)

PR #87, opened 2021-06-16
Reduce number of redundant spec builds after a PR is merged

scottwittenburg:
The goal of this PR is to reduce a lot of wasted compute time by
avoiding pushing branches where the base sha is the currently running
develop pipeline sha.  Once the develop pipeline has completed, those
branches will be pushed the next time the sync script runs.

### DIFF
--- a/gh-gl-sync/SpackCIBridge.py
+++ b/gh-gl-sync/SpackCIBridge.py
@@ -20,7 +20,7 @@ import urllib.request
 class SpackCIBridge(object):
 
     def __init__(self, gitlab_repo="", gitlab_host="", gitlab_project="", github_project="",
-                 disable_status_post=True, pr_mirror_bucket=None):
+                 disable_status_post=True, pr_mirror_bucket=None, main_branch=None):
         self.gitlab_repo = gitlab_repo
         self.github_project = github_project
         github_token = os.environ.get('GITHUB_TOKEN')
@@ -34,14 +34,20 @@ class SpackCIBridge(object):
 
         self.post_status = not disable_status_post
         self.pr_mirror_bucket = pr_mirror_bucket
+        self.main_branch = main_branch
 
         dt = datetime.now(timezone.utc) + timedelta(minutes=-4)
-        time_threshold = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
+        self.time_threshold_brief = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
+
+        # We use a  longer time threshold to find the currently running main branch pipeline.
+        dt = datetime.now(timezone.utc) + timedelta(minutes=-1440)
+        self.time_threshold_long = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
+
         self.pipeline_api_template = gitlab_host
         self.pipeline_api_template += "/api/v4/projects/"
         self.pipeline_api_template += urllib.parse.quote_plus(gitlab_project)
-        self.pipeline_api_template += "/pipelines?updated_after={0}".format(time_threshold)
-        self.pipeline_api_template += "&ref={0}"
+        self.pipeline_api_template += "/pipelines?updated_after={0}"
+        self.pipeline_api_template += "&ref={1}"
 
         self.commit_api_template = gitlab_host
         self.commit_api_template += "/api/v4/projects/"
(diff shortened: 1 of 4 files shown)


tgamblin (approved):

gh-gl-sync/SpackCIBridge.py
```diff
@@ -400,10 +467,19 @@ def sync(self):
             # Setup the local repo with two remotes.
             self.setup_git_repo()
 
+            # Find the currently running develop pipeline, if any, and get the sha
+            currently_running_develop_sha = None
+            develop_pipelines = self.get_pipelines_for_branch("develop", self.time_threshold_dev)
+            for dev_sha, dev_pipeline in develop_pipelines.items():
+                if dev_pipeline['status'] == "running":
+                    currently_running_develop_sha = dev_sha
```
zackgalbreath, line 475:
```suggestion
                    currently_running_develop_sha = dev_sha
                    break
```

gh-gl-sync/SpackCIBridge.py
```diff
@@ -415,9 +491,11 @@ def sync(self):
             # Find closed PRs that are currently synced.
             # These will be deleted from GitLab.
             closed_refspecs = self.get_prs_to_delete(open_prs["pr_strings"], synced_prs)
+            closed_refspecs = []
```
zackgalbreath, line 494: Does this mean we're no longer deleting old branches from GitLab?

gh-gl-sync/SpackCIBridge.py
```diff
@@ -36,12 +36,16 @@ def __init__(self, gitlab_repo="", gitlab_host="", gitlab_project="", github_pro
         self.pr_mirror_bucket = pr_mirror_bucket
 
         dt = datetime.now(timezone.utc) + timedelta(minutes=-4)
-        time_threshold = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
+        self.time_threshold_prs = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
+
+        dt = datetime.now(timezone.utc) + timedelta(minutes=-1440)
```
zackgalbreath, line 41:
```suggestion
        #We use a  longer time threshold to find the currently running develop pipeline.
        dt = datetime.now(timezone.utc) + timedelta(minutes=-1440)
```

gh-gl-sync/SpackCIBridge.py
```diff
@@ -400,10 +467,19 @@ def sync(self):
             # Setup the local repo with two remotes.
             self.setup_git_repo()
 
+            # Find the currently running develop pipeline, if any, and get the sha
+            currently_running_develop_sha = None
+            develop_pipelines = self.get_pipelines_for_branch("develop", self.time_threshold_dev)
```
zackgalbreath, line 472: Can we make this `self.main_branch_name` or so rather than hardcoding develop? I like the idea of keeping this sync script generally useful (not just for Spack).

zackgalbreath (changes requested):

gh-gl-sync/SpackCIBridge.py
scottwittenburg: There's no guarantee about the ordering of the returned pipelines, is there?  Is there any particular reason to pick the first one rather than the last?  Note that in our case at least, there's only ever going to be a single develop pipeline running anyway.

gh-gl-sync/SpackCIBridge.py
scottwittenburg: Thank you 😅 that was for debugging purposes and should not remain.

gh-gl-sync/SpackCIBridge.py
scottwittenburg: Comment makes sense, will fix.

gh-gl-sync/SpackCIBridge.py
scottwittenburg: Yes, makes sense.  How about another command line option?  Like maybe `--main-branch` which defaults to `develop`?

zackgalbreath: I like this idea

gh-gl-sync/SpackCIBridge.py
zackgalbreath: As far as I know, you're right: there's no guarantee what order GitLab's API will return its results in.

I guess if we wanted to be very thorough we would pull /all/ the running develop pipelines and check against all of them. What do you think, probably overkill?

scottwittenburg: To me it seems like overkill, considering our target use case involves pushing to a gitlab where "cancel redundant pending pipelines" is going to have to remain enabled forever.  Does the other project using this script not enable that feature?

zackgalbreath: As far as I'm aware Spack is the only project using this script that auto-cancels redundant pipelines. So I agree, probably overkill. Let's just grab the first (only) running dev pipeline & break out of the loop?

zackgalbreath (approved):

PR #92, opened 2021-06-23
Let develop pipelines run to completion

zackgalbreath:
Also, don't post canceled status to GitHub

### DIFF
--- a/gh-gl-sync/SpackCIBridge.py
+++ b/gh-gl-sync/SpackCIBridge.py
@@ -287,8 +287,10 @@ class SpackCIBridge(object):
             post_data["description"] = "Pipeline failed"
 
         elif pipeline["status"] == "canceled":
-            post_data["state"] = "failure"
-            post_data["description"] = "Pipeline was canceled"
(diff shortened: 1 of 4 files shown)


gh-gl-sync/SpackCIBridge.py
```diff
@@ -469,16 +478,15 @@ def sync(self):
...
             self.setup_git_repo()
 
-            currently_running_sha = None
             if self.main_branch:
                 # Find the currently running main branch pipeline, if any, and get the sha
                 main_branch_pipelines = self.get_pipelines_for_branch(self.main_branch, self.time_threshold_long)
                 for sha, pipeline in main_branch_pipelines.items():
                     if pipeline['status'] == "running":
-                        currently_running_sha = sha
+                        self.currently_running_sha = sha
```
scottwittenburg, line 486: Yeah, this is how I should have done it in the first place.  :+1:

scottwittenburg (approved):
This looks like it should do the two things it's supposed to do, thanks @zackgalbreath :+1:

PR #109, opened 2021-07-09
Add a pool of c6g.2xlarge GitLab runners

zackgalbreath:


k8s/runners/large-arm-pub/release.yaml
```diff
@@ -0,0 +1,101 @@
...
+        #
+        image: gitlab/gitlab-runner-helper:arm-latest
+      # cpuRequests: 50m
+      # cpuLimit: 50m
+      # memoryRequests: 144Mi
+      # memoryLimit: 144Mi
+
+      nodeSelector:
+        spack.io/node-pool: glr-large-pub  # pool for jobs
+        kubernetes.io/arch: arm64
```
scottwittenburg, line 98: Yesterday you tried removing this, right?  In favor of some other label placed on the autoscaling group?  Just making sure that didn't pan out and including this here is deliberate.

zackgalbreath: That's right. Changing this into a custom, manually-defined label/tag didn't help the situation.

opadron (approved):

PR #111, opened 2021-07-15
spackbot: deploy new version; Add GITLAB_TOKEN to secrets

tgamblin:
This adds one secret to the `spackbot` environment so that it can restart pipelines on gitlab.spack.io.

This also adds a `secrets-dummy.yaml` file so that I can remember how the heck to format private keys so that Kubernetes will actually give them to the bot correctly.  You can look in the file to see commands that will work.

### DIFF
--- a/k8s/spack/spackbot-spack-io/secrets-dummy.yaml
+++ b/k8s/spack/spackbot-spack-io/secrets-dummy.yaml
@@ -1,0 +1,22 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: spack-bot-credentials
+  namespace: spack
+  annotations:
+    fluxcd.io/ignore: "true"
+  labels:
+    app: spackbot-spack-io
+    svc: web
+data:
+  gitlab_token: |
+    # result of:
+    echo -n ${gitlab_token} | base64
+  github_webhook_secret: |
+    # result of:
+    echo -n ${gitlab_webhook_secret} | base64
+  github_private_key: |
+    # result of doing this to github private key download:
+    cat ~/Downloads/spack-bot.2021-07-14.private-key.pem  | perl -pe 's/\n/\\n/;' | perl -pe 's/\\n$/\n/;' | base64
+    # note that you can leave the newlines in the base64 encoded output -- k8s understands them.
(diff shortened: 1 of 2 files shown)


k8s/spack/spackbot-spack-io/secrets-dummy.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: spack-bot-credentials
+  namespace: spack
+  labels:
+    app: spackbot-spack-io
+    svc: web
+data:
+  gitlab_token: |
```
vsoch, line 11: Ah gotcha on how this works! Where were the secrets coming from before this file (for the webhook, etc). We’re they just set on the command line but not auto renewed or generated?

k8s/spack/spackbot-spack-io/secrets-dummy.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
opadron, line 1: Could you rename this to `secrets.yaml`?  That will match how the secrets are laid out in the other namespaces.

k8s/spack/spackbot-spack-io/secrets-dummy.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: spack-bot-credentials
+  namespace: spack
```
opadron, line 6: Make sure to include the following annotation.  This is needed so that flux doesn't confuse this for a real secret!

```suggestion
  namespace: spack
  annotations:
    fluxcd.io/ignore: "true"
```

opadron (changes requested):

k8s/spack/spackbot-spack-io/secrets-dummy.yaml
tgamblin: done

k8s/spack/spackbot-spack-io/secrets-dummy.yaml
tgamblin: I had a file like this that wasn't checked in (b/c secrets) and got lost in the move from `kubernetes` to `k8s`.  I deleted the old `kubernetes` directory where it lived. So I checked this one in and I figure why not use the magic formatting commands as the dummy secret placeholder values.

opadron (approved):

k8s/spack/spackbot-spack-io/secrets-dummy.yaml
tgamblin: we decided this should be called `secrets-dummy.yaml`, and that the real secrets files would be created from that.

PR #112, opened 2021-07-16
Adding first shot at drift server

vsoch:
@alecbcs is working on a drift analysis for spack (or BUILD SI) and part of that requires deploying this drift server:

https://github.com/buildsi/drift-server

We have a PR underway that is going to do an automated build for the container base:

https://github.com/buildsi/drift-server/pull/7

and our plan is to create an AWS database to connect to (via the `DRIFTSERVER_DATABASE_DSN` that is a secret). We will also define some username and password (also secrets) that have permission to write to it.

In terms of the kubernetes yaml recipes, I took my best effort first shot here! The container basically runs the app when you run it on port 8080, so it shouldn't need to look very different than the other spack webby things. I'm thinking before this gets added here so it's automated, we probably want to get the container deployed, create the database, and then test our deploying the files from our local machines. If that works, then would adding the recipes here automate it (but we would still need to deploy our secrets from our local machines?)
 
@opadron if you have maybe 30 minutes next week (and think it would be fun) I think this could be a fun way for @alecbcs and I to  meet up with you and walk through the steps of interacting with kubernetes to test this out. Hey, maybe we might even be useful someday! :laughing: 

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/k8s/spack/drift-server/certificates.yaml
+++ b/k8s/spack/drift-server/certificates.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: tls-drift-server
+  namespace: spack
+spec:
+  secretName: tls-drift-server
+  issuerRef:
+    name: letsencrypt
+    kind: ClusterIssuer
+  commonName: drift-server.spack.io
+  dnsNames:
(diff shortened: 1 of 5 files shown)


vsoch:
@opadron @alecbcs this is WORKING! wooo!!!! :D

PR #113, opened 2021-07-21
Add scripts to populate and query db with build statistics

scottwittenburg:
This PR adds two new scripts:

`generate_builds_stats.py`  Fetches and caches api endpoint results and job artifacts, and uses them to populate two tables in an `sqlite3` database: `pipelines` and `builds`.  The table schema look like this:

```
CREATE TABLE pipelines (
    pipeline_id INTEGER NOT NULL primary key,
    link TEXT,
    pr_number TEXT,
    branch TEXT);
```

```
CREATE TABLE builds (
    job_id INTEGER NOT NULL primary key,
    job_name TEXt,
    status TEXT,
    duration REAL,
    pkg_name TEXT,
    dag_hash TEXT,
    build_hash TEXT,
    full_hash TEXT,
    link TEXT,
    pipeline_id INTEGER NOT NULL,
    FOREIGN KEY (pipeline_id)
        REFERENCES pipelines (pipeline_id)
);
```

If you run the above script many times, only new things are fetched and added to the database, and all the data cached on the filesystem allow the subsequent runs to finish quickly.

`query_build_stats.py`  Just gives some example queries against the database.  I'm not very familiar with SQL, so the queries don't yet produce exactly what we want, but hopefully the information is there for someone who has stronger sql-fu.  If not, we can tweak what we're gathering and storing.

Eventually, we may want to consider standing up a service in the cluster backed by some persistent storage where we can keep the database growing, which will allow us to quickly formulate and issue queries against a long history of stored builds.  Right now, I'm just running the script locally and growing a database on my laptop.


scottwittenburg:
Just noticed, it looks like the project (or maybe just this directory) needs a `.gitignore` with an entry to avoid adding pycache files.

opadron:
This is awesome!

Not sure if this is what you were going for, but I think something like this will give you every spec that had built more than once in a given PR, along with the corresponding PR:

```sql
SELECT pipelines.pr_number AS pr,
       builds.full_hash AS hash,
       COUNT(*) AS count
FROM builds
     INNER JOIN pipelines
     ON builds.pipeline_id == pipelines.pipeline_id
WHERE pr IS NOT NULL
GROUP BY pr, hash
ORDER BY count DESC
```

Or maybe you want to work out how many times a build had to be retried in a PR (not counting the initial attempt):
```sql
SELECT pr,
       SUM(count) - COUNT(*) AS retries
FROM (
    SELECT pipelines.pr_number AS pr,
           COUNT(*) AS count
    FROM builds
         INNER JOIN pipelines
         ON builds.pipeline_id == pipelines.pipeline_id
    WHERE pr IS NOT NULL
    GROUP BY pr, builds.full_hash
)
ORDER BY retries DESC
```

Buyer beware: I have not tested any of these queries.  They almost certainly have bugs/typos! 😄

zackgalbreath (approved):

zackgalbreath (changes requested):
Looks like this file was added by accident:
scripts/gitlab_api/__pycache__/gitlab_api.cpython-39.pyc

PR #115, opened 2021-07-23
fix/update runners & fix prometheus release

opadron:
Last step in fixing the autoscaling issue.

Also updates all runners so that we have a complete offering of both x86 and arm runners.


zackgalbreath (approved):

opadron:
@zackgalbreath I'm thinking of pushing to `production` and then turning flux back on.  What do you think?

zackgalbreath:
I think that's a good idea. Feel free to reach out if things don't go smoothly.

PR #118, opened 2021-07-30
Bump 'brief threshold' from 15 to 30 minutes

zackgalbreath:
I continue to notice the occasional pipeline not getting its updated status posted back to GitHub.

### DIFF
--- a/.github/workflows/spack_ci_bridge_docker.yml
+++ b/.github/workflows/spack_ci_bridge_docker.yml
@@ -28,7 +28,7 @@ jobs:
           context: ./gh-gl-sync
           file: ./gh-gl-sync/Dockerfile
           push: true
-          tags: zackgalbreath/spack-ci-bridge:0.0.18
+          tags: zackgalbreath/spack-ci-bridge:0.0.19
       -
         name: Image digest
         run: echo ${{ steps.docker_build_sync.outputs.digest }}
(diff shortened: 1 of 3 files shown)


zackgalbreath:
This is anecdotal, but this failure-to-updated-GitHub-status mostly seems to happen to pipelines that were 'backlogged' due to being based on a version of develop whose pipeline hadn't finished running yet. Anyway, hopefully this threshold bump solves the problem.

I performed some manual timing results that indicated that this increase in our "brief threshold" shouldn't show the sync script down too much:

```
current main branch (15 mins): 2m23.735s
bump to 30 minutes: 2m33.659s
bump to 12 hours: 2m42.540s
```

scottwittenburg (approved):
Looks good to me.  Just curious if you have reason to think 30 min is enough to solve the problem, when even looking back 12 hrs doesn't add that much more time.

zackgalbreath:
> Looks good to me. Just curious if you have reason to think 30 min is enough to solve the problem, when even looking back 12 hrs doesn't add that much more time.

Good question. I guess I still feel some desire to keep the "brief" threshold as short as possible, just in case it does end up making the sync script slower, or GitHub gets mad at us for hammering the status API too much. So basically my plan is to bump this threshold up a little bit more (or maybe double it) each time we see this problem rear its ugly head.

PR #126, opened 2021-09-11
Depoly a temporary spackbot image built from e3405654bb (#125)

scottwittenburg:
Test the temporary container, in production of course!


scottwittenburg:
Perhaps that description is a bit too tongue-in-cheek?

PR #129, opened 2021-09-17
Increase brief threshold again

zackgalbreath:
It's still a mystery why this is happening, but we continue to see the
occasional GitHub PR not get its status updated after the corresponding
GitLab CI pipeline has finished running.


scottwittenburg (approved):
It seems like this has the potential to help, at least a little, and shouldn't cause the sync script to take significantly longer.

zackgalbreath:
I'll deploy this change to the cluster early next week

PR #130, opened 2021-09-17
URL encode branch names before hitting gitlab pipeline api

scottwittenburg:
I think this should fix an issue the sync script had preventing it from seeing pipelines on certain branches.


zackgalbreath:
I tested this out locally & it seemed to have worked as intended

zackgalbreath (approved):

PR #132, opened 2021-09-21
Only backlog PRs we would otherwise have pushed

scottwittenburg:
Instead of marking a PR backlogged anytime the github merge  commit includes a running develop, only backlog PRs if we would otherwise have needed to push its branch too.


scottwittenburg:
I know, I need to fix the test mockage etc...

zackgalbreath (approved):

PR #134, opened 2021-09-23
Set $CDASH_PRODUCTION_MODE = 1

zackgalbreath:
This should fix a bug we noticed where some API endpoints
attempt to get loaded over http instead of https.


scottwittenburg (approved):
Approving without really knowing what was wrong before or why this fixes it.

zackgalbreath:
> Approving without really knowing what was wrong before or why this fixes it.

That's the spirit!

PR #136, opened 2021-09-24
Gitops migration

opadron:
Adds my custom gitops controller.

Container image taken from my Github registry.  See https://github.com/opadron/images/tree/main/gitops for code and details.

TODO:

- [ ] Review/Merge this PR
- [ ] Update `production` branch
- [ ] Apply changes in `flux-chart.yaml`
  - This should get flux to switch from observing the `production` branch to the `flux` branch
- [ ] Follow up with a Gitlab staging environment

### DIFF
--- a/k8s-bootstrap/flux-chart.yaml
+++ b/k8s-bootstrap/flux-chart.yaml
@@ -22,8 +22,8 @@ nodeSelector:
 
 git:
   url: "ssh://git@github.com/spack/spack-infrastructure"
-  branch: "production"
-  path: "k8s"
+  branch: "flux"
+  path: "flux"
   user: "Spack Bot"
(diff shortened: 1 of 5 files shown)


scottwittenburg (approved):
Thanks @opadron, this looks good to me.  It does seem all the magic is happening in your images repo though, and not part of this review.  I did look through that to get a sense what's going on and verify cli args, etc.  With that level of review, it looks good, but I'm curious why you put that stuff over there instead of here?  Maybe you plan to use it in another project, in which case that could make sense.

opadron:
Thanks, @scottwittenburg!

You're right, the image and build process should be moved over to this repo.  I'll add that as a follow-up issue.

PR #140, opened 2021-10-04
add gitlab log cleaner

opadron:
Runs every day at 4 AM.  Goes through the gitlab minio instance deleting log files that:

 1. Are older than seven days, and
 2. Found under a prefix key that matches a very specific pattern.

### DIFF
--- a/k8s/gitlab/log-cleaner.yaml
+++ b/k8s/gitlab/log-cleaner.yaml
@@ -1,0 +1,71 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: log-cleaner-script
+  namespace: gitlab
+data:
+  script: |
+    #! /bin/bash
+    set +o history
+    mc alias set gitlab https://minio.gitlab.spack.io \
+        "$ACCESS_KEY" "$SECRET_KEY"
+    set -o history
+    echo
+
+    mc find gitlab/gitlab-artifacts --maxdepth 5 --path '*_*_*' -exec \
+        'mc find {} --maxdepth 3 --older-than 7d --name job.log' \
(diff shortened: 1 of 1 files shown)


zackgalbreath:
I'm curious, did you try running this manually? minio is still looking fuller than I would have expected

```
Filesystem                Size      Used Available Use% Mounted on
/dev/nvme2n1            984.2G    523.5G    460.7G  53% /export
```

k8s/gitlab/log-cleaner.yaml
```diff
@@ -0,0 +1,71 @@
...
+  script: |
+    #! /bin/bash
+    set +o history
+    mc alias set gitlab https://minio.gitlab.spack.io \
+        "$ACCESS_KEY" "$SECRET_KEY"
+    set -o history
+    echo
+
+    mc find gitlab/gitlab-artifacts --maxdepth 5 --path '*_*_*' -exec \
+        'mc find {} --maxdepth 3 --older-than 7d --name job.log' \
```
scottwittenburg, line 17: Seven days seems a little short, at least to me.  Since we have gitlab set to keep artifacts around for 30 days, maybe we could go with the same thing here?

zackgalbreath: good call, I agree that we should pick an expiration horizon and stick with it

opadron:
Thanks for the feedback.

@zackgalbreath yes, I did run it, but with a ` | head -n ... | ` stuck in the middle to limit the fallout in case I messed something up while testing interactively.  When we eventually push this to `production`, it'll go through and delete all the respective files.

@scottwittenburg you're totally right about the 7-day expiration.  I'll update it momentarily.

PR #142, opened 2021-10-07
Attempt to add spack monitor with new setup

vsoch:
We want to give another shot at deploying spack monitor (a Django application) on K8s!

This is the service at monitor.spack.io, but I am calling it builds.spack.io since that name is taken (and for the time being we need both to coexist). This one is different / updated because it has user authentication (OAuth2 via GitHub) to get a token.

I think probably a good approach would be to review here, and then when something looks relatively okay allow me to test? @opadron let me know what works for you, and c.c @tgamblin.

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/k8s/spack/builds-spack-io/ingress.yaml
+++ b/k8s/spack/builds-spack-io/ingress.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: builds-spack-io
+  namespace: spack
+spec:
+  tls:
+  - secretName: tls-spackbuilds
+  rules:
+  - host: builds.spack.io
+    http:
+      paths:
+      - backend:
+          serviceName: builds-spack-io
+          servicePort: 80
--- a/k8s/spack/builds-spack-io/certificates.yaml
+++ b/k8s/spack/builds-spack-io/certificates.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: tls-spackbuilds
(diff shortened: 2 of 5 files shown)


opadron (approved):
LGTM.  Give it a try and let me know how it goes! :+1:

vsoch:
okay updated with database and adding a default database port.

k8s/spack/builds-spack-io/ingress.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: builds-spack-io
+  namespace: spack
```
opadron, line 6: Here are the changes that I had to apply to your ingress in order to get around the `413` size limit issue.

```suggestion
metadata:
  name: builds-spack-io
  namespace: spack
  annotations:
    nginx.ingress.kubernetes.io/proxy-body-size: "10024M"
    nginx.ingress.kubernetes.io/client-body-buffer-size: "10024M"
    nginx.ingress.kubernetes.io/proxy-connect-timeout: "120"
    nginx.ingress.kubernetes.io/proxy-send-timeout: "120"
    nginx.ingress.kubernetes.io/proxy-read-timeout: "120"
```

vsoch: I love that I can control this directly from my configs! Let's give it a try. I'm building new containers with the migrations and will do another deploy after that.

vsoch:
and if you already applied ingress.yaml I realize I don't need to again :) But I'll redeploy with the correct migration files, and then do some builds to make sure everything is working as expected!

PR #148, opened 2021-10-21
Fix docker build

zackgalbreath:


.github/workflows/custom_docker_builds.yml
```diff
@@ -12,23 +12,21 @@ jobs:
...
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       -
         name: Build and push ci-bridge
         id: docker_build_sync
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
```
scottwittenburg, line 24: just curious, is this for security reasons?  Like you've vetted this sha as safe, and don't want them changing the action out from under you?

.github/workflows/custom_docker_builds.yml
```diff
@@ -12,23 +12,21 @@ jobs:
...
         with:
           context: ./images/gh-gl-sync
           file: ./images/gh-gl-sync/Dockerfile
           push: true
-          tags: spack/ci-bridge:0.0.21
+          tags: ghcr.io/spack/ci-bridge:0.0.21
```
scottwittenburg, line 29: Ugh, sorry I missed this in the initial review.

images/gh-gl-sync/requirements.txt
```diff
@@ -1,24 +1,33 @@
...
+pytest==6.2.5
+pytest-cov==3.0.0
+python-dateutil==2.8.2
+requests==2.26.0
+s3transfer==0.5.0
+six==1.16.0
+toml==0.10.2
+tomli==1.2.1
+urllib3==1.26.7
+wrapt==1.13.2
```
scottwittenburg, line 33: Also just curious how you came up with this list, which I assume is pinning all the versions for the same reason as using a hashed action.

.github/workflows/custom_docker_builds.yml
zackgalbreath: These specific hashes were copied from the example documented here:

https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action

scottwittenburg: Ok, good to know.

images/gh-gl-sync/requirements.txt
scottwittenburg: Since you ran it with those new deps, I still approve 😁

PR #152, opened 2021-11-04
Gitlab canary

opadron:
Adds canary deployments for Gitlab and one gitlab runner.  The runner is configured to run up to 3 jobs at a time on the `base` node pool.  The jobs can share instance resources with each other as well as any other pods running on the pool.  This should be fine for testing, and has the added benefit that jobs should start up and run quickly since it is unlikely that they would need to wait for cluster autoscaler to bring up a new instance.

While producing these patches, the value of having a tool to preview the results of applying a patch to some resources became clear, and so I also added the `gitops-patch.py` script.  This script can apply a patch and show you the result.  It uses the same logic that my `gitops` controller uses when applying patches.

As a final detail, I had to manually implement the processing and application of these patches since the only python library available doesn't work.  I've only implemented a subset of the standard, but realized that there was a part of the standard that was missing from the implementation: handling `~0` and `~1` escape sequences in paths.  This PR also adds code that handles these sequences in both the above script as well as the controller, so I needed to bump the version number on the image.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -201,7 +201,7 @@ jobs:
           context: ./images/gitops
           file: ./images/gitops/Dockerfile
           push: true
-          tags: ghcr.io/spack/gitops:0.0.1
+          tags: ghcr.io/spack/gitops:0.0.2
       -
         name: Image digest
         run: echo ${{ steps.docker_build_gitops.outputs.digest }}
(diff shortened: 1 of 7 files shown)


scottwittenburg (approved):
Looks good to me, can't wait to see whether it works!

opadron:
Thanks!

Here's my plan for tomorrow's maintenance window:

- [ ] Turn flux off
- [ ] Merge this PR
- [ ] Wait for github actions to build the new version of the gitops controller image.
- [ ] Manually update the gitops deployment.
- [ ] Merge the latest `main` changes to `production`.
- [ ] Confirm that the gitops controller generated all of the correct manifests and pushed to `flux` branch.
- [ ] Turn flux back on
- [ ] Cross my fingers and hope that nothing breaks!
- [ ] Fix the things that break...🙃

PR #154, opened 2021-11-05
fix gitops initialization logic/add gitlab namespace to patches

opadron:
Fixes two other issues identified during maintenance:

1. Fixes an issue where the gitops controller wasn't initializing properly on its first iteration.
2. Adds the new patch config maps to their correct namespaces.


opadron:
#152

PR #157, opened 2021-11-15
Add Metabase and monitoring namespace

kotfic:


kotfic:
@opadron PTAL - thanks!

k8s/metabase/postgres-service.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: metabase-postgres
+  namespace: monitoring
+  labels:
+    app: metabase-postgres
+spec:
+  type: NodePort
```
opadron, line 10: This should be a `ClusterIP` service.

opadron (changes requested):

opadron (approved):

PR #158, opened 2021-11-18
Document change control process

opadron:
This is the start of some text I'm putting together to document the change control process for this repo/the cluster.

Please have a look and provide feedback; both on the text and the process, itself.

What's missing that should be included?
What could be better explained or expanded on?
What do you think of the process?  Any changes you think we should make?


k8s/docs/change-control-process.md
```diff
@@ -0,0 +1,106 @@
...
+
+Although this change control process is centered around a single git repository,
+the contents are not "developed" in the same way that a code base would be.  We
+try to keep the process similar where possible for the sake of familiarity, but
+deviate where necessary.
+
+For example, there is no complete analog for local development and testing; only
+incomplete approaches with various tradeoffs.
+
+ - You could stand up your own local kubernetes cluster using something like
```
vsoch, line 17: Maybe kind https://github.com/kubernetes-sigs/kind would be another suggestion? I haven't tried either of these for years, but if I can remember I had more luck with the latter. And this statement hits the nail on the head - I haven't gone back to these approaches because I'm aware they cannot reproduce the actual setup that the cluster users.  But I wonder - is there a translation process, e.g., you can give a new user a set of templates to start with for the general kind of supported apps, and then when they are ready for production switch in or add the changes?

k8s/docs/change-control-process.md
```diff
@@ -0,0 +1,106 @@
...
+   [minikube](https://minikube.sigs.k8s.io/docs/) and iterate on that, but there
+   are a number of facilities that are hard or impossible to reproduce locally,
+   such as DNS and certificate management.  Also, many of the file contents in
+   this repo contain portions that are specific to the production cluster, and
+   would likely not work for a local development cluster without modification.
+
+ - You could deploy resources that you are iterating on directly to the
```
vsoch, line 24: I've had surprisingly an easy time with this! I think for new folks it would be helpful to have a guide akin to the manual steps  you walked @alecbcs and I though. It's important to have a frame of mind for how to figure these things out, or just strategies for doing that, and for knowing to do a step, check the outcome, and debug if something is wrong.

k8s/docs/change-control-process.md
```diff
@@ -0,0 +1,106 @@
...
+occur on the first and third Friday of every month, starting on midnight,
+eastern time, on that Friday and ending on midnight, eastern time, the next day.
+
+#### 3 - The flux Branch
+
+The `flux` branch is a branch that is automatically managed by middleware
+running on the cluster.  Its contents represent the exact resources that are
+provisioned on the cluster at any given time and are generated by combing the
+contents of the `main` and `production` branches for staged and production
+resources, respectively.
```
vsoch, line 64: Maybe add a note like "You would generally not open a pull request against or interact with this branch" (is that is the case). Also flux might get confused with https://github.com/flux-framework/flux-core

k8s/docs/change-control-process.md
```diff
@@ -0,0 +1,106 @@
...
+      is opened and merged.
+
+    - Cluster middleware triggers changes to the `flux` branch.
+
+    - Cluster middleware responds to changes in the `flux` branch by updating
+      cluster resources.
+
+    - The state of the cluster and services are reviewed for proper operation.
+
+ 3. If there are any issues that arise as a result of step 2, then remedial
```
vsoch, line 87: I would add in here that it's possible to have a local clone where you are interacting with the cluster, and the PR opened for a long period of time (e.g., I don't think I've ever had any of my PRs actually merged, and this is understandable because it's sort of "developing and tweaking as I go" work.  But it's good to set expectations about merging/not without branding not merging as "this isn't good enough" when that's not the case.

vsoch (commented):
This is great!

k8s/docs/change-control-process.md
opadron: You could start with more general templates, but the problem is that there is no one translation process that would work; what the generalized manifests would look like and how you would go from those to versions suitable for the cluster need to be evaluated on a case-by-case basis.

k8s/docs/change-control-process.md
opadron: That's a great point.

Speaking of which... we might want to start thinking about getting those PRs merged so that they can benefit from this process.  Iterating on your services would take a little more work, but it's something we should do once you feel like you've gotten those PRs to a good, stable state.

opadron:
@vsoch thanks for much for the feedback!  I've added some more to the text.

k8s/docs/change-control-process.md
opadron: Also, I have worked these and other tools and have gotten a lot of mileage with them.  They're definitely helpful!  I guess it's just really hard to faithfully reproduce some facilities locally, though there are efforts like [metallb](https://metallb.org/) that have been slowly chipping away at this deficit.

I think a guide on setting up your own local kubernetes environment might be a good thing for me to work on next!

k8s/docs/change-control-process.md
vsoch: What would you define as stable? I will likely continue to tweak the image tag if there are changes to the server - that's definitely died down from its original frequency, but could happen again after more discussion during a team meeting. The other changes that might be likely are upping the pvc sizes (like we did early this week, successfully!) I'm happy to leave them in a PR state if these kind of changes make the deployments not yet stable.

And for drift server ping @alecbcs!

opadron: So, I added some text that touches on this, but it really depends on you.  The question is "are your services at a point where you'd like their deployment state to be part of this change control process?"  The upside is that your deployment state is tied to the repo state, so there's never a question of what version of your application is currently deployed.  Also, updates to your services are rolled out on a regular schedule using a process that allows for easy roll back in case there are any issues with new changes.

The downside is that you no longer have the flexibility of just tweaking your application deployment live.  You would have to set up a separate deployment to have something to tweak.

vsoch: It sounds easier then to leave as is! Let's do that :)

mvandenburgh:
This is no longer relevant after switching to Flux v2.

PR #163, opened 2021-12-15
Gitlab runner revamp

opadron:
Revamps the gitlab runner configurations so that instead of running each job size on their own dedicated instance type, we run all jobs on `c5.12xlarge` instances, configure each job size to ask for a different amount of cpu and memory, and allow kubernetes to decide how to pack the jobs.

The proposed runner breakdown is summarized in the table below.  `Capacity` refers to the maximum number of jobs that can be ran simultaneously.  `Density` is the number of jobs that can fit on a single instance.

| Job Size (Tag) | Core Count | Memory (GiB) | Capacity | Density | 
|--|--|--|--|--|
| medium | 4 | 8 | 50 | 12
| large | 8 | 16 | 100 | 6
| xlarge | 16 | 32 | 5 | 3
| huge | 46 (48 - 2) | 94 (96 - 2) | 1 | 1

A single `c5.12xlarge` instance has 48 cores and 96 GiB of memory.  They are split 12 ways for medium jobs, 6 ways for large jobs, and 3 ways for xlarge jobs.  Huge jobs are given dedicated access to a single instance, minus two cores and two GiB of memory for overhead.

In addition, we used statistics from [stats.e4s.io](https://stats.e4s.io) to estimate the proportions of jobs ran that were made up of each job size and reconfigured the runner capacities to better match these proportions.  We found that a large fraction of all jobs ran where `large` jobs, about half as many jobs were `medium` and very few jobs were `xlarge` or `huge`.  We don't use `tiny` or `small` sizes, because our baseline resource requirements exceed what they can provide, and so this PR removes the `small` runners (the `tiny` runners had already been removed in a prior PR).

The proposed breakdown represents a total of approximately 23.5 `c5.12xlarge`s worth of capacity.  The underlying node pools and autoscaling groups have been reconfigured to accommodate this new capacity.  This PR also mirrors the configuration changes for the ARM runners and the `c6g.12xlarge` instances that run under the `huge` tag.  So, these changes actually represent about twice as much capacity; half of which is for x86 and the other half for arm.

I recommend that we merge these changes, push them to `production` during the next maintenance window (this Friday!), and dedicate the entire maintenance window to monitoring and troubleshooting any issues that might arise.  With these changes, I expect that we will considerably increase the number of simultaneous jobs that we are running, and the jobs will make more efficient use of the resources they are scheduled on.  However, considering that we've already been having issues with Gitlab availability during peak utilization, I expect those issues to get worse as we increase throughput.  We will very likely need to follow these changes up with additional changes that further scales out the Gitlab deployment in order to meet the increased peak demand.

Other misc. changes:
 - Now that all runners will be running on the same instance type, the `avx2` and `avx512` tags have been added to the medium runner.


zackgalbreath:
I'm having trouble understanding the difference between capacity & density, and why these numbers differ. Can I get an example?

scottwittenburg:
> I'm having trouble understanding the difference between capacity & density, and why these numbers differ. Can I get an example?

Let me take a stab at that as a test of whether I understand.  

`Capacity` is the number of simultaneous jobs we want to be able to handle for a particular tagged size, while `Density` is the number of simultaneous jobs of that tagged size @opadron is planning to pack onto a `c5.12xlarge`, which was in turn based on dividing up the cores and memory of a `c5.12xlarge` in such a way as to sort of preserve the cores/mem per tagged size we have currently.

And it seems the 23.5 instance estimate (for the number of `c5.12xlarge`) came from dividing `Capacity / Density` for each tagged size and summing them up.

zackgalbreath:
That makes more sense (and the math checks out). Thanks, Scott.

To put it another way, we will force "huge" jobs to run in serial (one at a time) even though we could run 23 in parallel with the proposed amount of compute capacity.

scottwittenburg:
> we will force "huge" jobs to run in serial (one at a time) even though we could run 23 in parallel with the proposed amount of compute capacity

Assuming `requestConcurrency` means what we both seem to be assuming it does, I agree that's an outcome of this.  We might tweak the `Capacity` of `huge` in that case.  O we can just make the change and then see if we can find some evidence that limiting those to one at a time is a significant bottleneck.

On a sort of related note, I'm sure curious how going from 20-28 instances each for a handful of node pools to ~25 `c5.12xlarge` will affect the cost.

opadron:
@scottwittenburg @zackgalbreath yes, you guys got the right idea.

@scottwittenburg the node pool is now 24 - 32 instances.  I'm keeping the two instances per AZ margin to handle failures.  We might also occasionally dip into that margin and grab an extra 25th instance during peak load, since overhead means that resources won't always divide as evenly as the above table would suggest.

I'm also very curious to see how the change will affect cost.  They are much more expensive instances, but the hope is we'll make more efficient use of their resources compared to what we're doing now.

scottwittenburg (approved):
Looks good as far as I can tell, thanks @opadron!  On the EC2 side, are you planning to start by only adding to the huge node pool, or are you thinking you'll reduce/remove the others as well?

opadron:
For now, I only increased the huge pool.  The other pools are still in place; we're just not using them anymore.  At some point, I'll remove them altogether unless we find another use for them.

scottwittenburg:
> For now, I only increased the huge pool.

Makes sense, especially in case we have to roll back the infrastructure change.  👍

zackgalbreath (approved):

PR #164, opened 2021-12-15
turn on async parsing for CDash

zackgalbreath:
This should allow us to more easily cope with 'bursty' workflows.


scottwittenburg (approved):
Yes, this makes sense, thanks @zackgalbreath!

PR #167, opened 2021-12-20
Add Slack notification to Prometheus Alert Manager

kotfic:
@opadron PTAL


kotfic:
Note that this also fixes the `dummyalert.sh` script to use client certs (these must be generated locally)

PR #169, opened 2021-12-22
Allow huge jobs to run on the glr-huge-pub pool

zackgalbreath:
I don't completely understand why, but the amount of memory we were
requesting (94Gi) was too much for cluster-autoscaler to spin up a new instance
for huge jobs:
```
Pod <snip> can't be scheduled on <huge-ASG>, predicate checking error:
Insufficient memory; predicateName=NodeResourcesFit;
reasons: Insufficient memory; debugInfo=
```
I also tried 90 GiB, but this resulted in the same error. Thus, the current
value of 88 GiB was more-or-less discovered by trial & error.

### DIFF
--- a/k8s/runners/huge-x86-pub/release.yaml
+++ b/k8s/runners/huge-x86-pub/release.yaml
@@ -76,7 +76,7 @@ spec:
 
       builds:
         cpuRequests: 46000m
-        memoryRequests: 94Gi
+        memoryRequests: 88Gi
 
       services: {}
       # cpuRequests: 50m


zackgalbreath:
I scaled down flux to manually deploy this change. Here's evidence of a huge job actually running now:

https://gitlab.spack.io/spack/spack/-/jobs/1393247

zackgalbreath:
@josephsnyder PTAL

josephsnyder:
It looks good to me.  I don't know where you found that log in the console yet, but the evidence of it running seems to be hard to beat.

opadron:
Noting instructions for getting a tighter bound.  I will handle these steps unless someone else gets around to it before I do:

1. Tweak the huge autoscaling group to have a minimum of 1 instance.  (We'll use this instance as our lab rat.)
2. Once the node comes up and the cluster sees it.  Add an `unschedulable` taint to keep jobs from running on it.
```Shell
kubectl cordon $NODE_NAME
```
3. Drain the node of any jobs that might be running on it (leaving daemonsets running).
```Shell
kubectl drain --ignore-daemonsets $NODE_NAME
```
4. Wait 5-10 minutes to give things a chance to settle down.
5. Get node resource stats.
```Shell
kubectl get nodes $NODE_NAME -o yaml | less
```
6. See `status.capacity` for total resources as they are reported to kubernetes.
7. See `status.allocatable` for available resources. (The difference will give you an idea of the baseline level of resource consumption).
8. Reduce the min # instances on the autoscaling group back to 0.
9. Ensure that cluster autoscaler eventually terminates the instance.

PR #179, opened 2022-01-13
Gitlab runner revamp revert (push to production)

opadron:
After #163, we're getting hit by a nasty Gitlab Runner bug where stale pods are being left around after they have finished building their packages.  As far as we can tell, these pods are left running indefinitely, occupying resources and preventing new work from running.  We don't know why this issue is affecting us, nor why it only started after we changed the way our runners were deployed.  This is a critical issue affecting the pipelines, and our immediate response is to revert the changes in the hopes that it goes away.

After this is merged, I'll update the underlying `huge` pools to only run up to 20 instances.


scottwittenburg (approved):
Thanks @opadron 🤞

PR #180, opened 2022-01-18
add temporary large runner (push to production)

opadron:
Temporarily adds a new `large` runner that will run side-by-side with the current runners.  This temporary runner will accept new jobs with the `large` tag and pack them into `c5.12xlarge` instances.  These instances are provisioned using a temporary node pool, `glr-huge-tmp-pub`.  The runner is currently set to run up to 10 jobs at a time.

The plan is to introduce this new runner and monitor jobs for evidence of the "stuck" pods issue we were observing last week.  Assuming the runner runs without issues, we will gradually increase the capacity until we observe the "stuck" pods issue, again.

### DIFF
--- a/k8s/runners/large-x86-tmp-pub/release.yaml
+++ b/k8s/runners/large-x86-tmp-pub/release.yaml
@@ -1,0 +1,98 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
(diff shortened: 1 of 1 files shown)


scottwittenburg (approved):
Thanks for this @opadron, and for answering my questions in chat 😁  

Looking forward to see how high we can go.  

And to getting `main` and `production` back in sync 😬

PR #182, opened 2022-01-21
Backport hotfixes

opadron:


.github/workflows/custom_docker_builds.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           context: ./images/gh-gl-sync
           file: ./images/gh-gl-sync/Dockerfile
           push: true
-          tags: ghcr.io/spack/ci-bridge:0.0.22
+          tags: ghcr.io/spack/ci-bridge:0.0.24
```
zackgalbreath, line 29: This diff doesn't make sense to me, as this change is already on `main`:
https://github.com/spack/spack-infrastructure/blob/main/.github/workflows/custom_docker_builds.yml#L29

Any idea what's going on here?

opadron: I'm not sure.  My guess is that there have been merges to `main` since `production` branched off, and the change appears independently in both branches.

PR #185, opened 2022-01-21
Update Production

opadron:
- New prometheus stack
- Slack notifications
- ingress metrics
- incremental runner revamp
- sync script improvements
- Gitlab DB node selector fix
- latest gitlab + runner in staging


zackgalbreath:
LGTM, but I defer to @kotfic to verify that this is the intended set of prometheus/metabase changes to pull in

opadron:
Recording for posterity: I'm running into errors restoring a Gitlab DB backup using Gitlab's documented process.  Apparently, they are expected and not an actual error, but I guess suppressing them and maintaining the confidence of their users is asking for too much.

https://gitlab.com/gitlab-org/gitlab/-/issues/266988#note_430408658

> When the db restore encounters errors, we fail the restore, except for two types of errors that we ignore. So if your restore succeeds, it should be good. We ignore does not exist and must be owner of errors, as these are expected error messages from pg-restore for the sql user that we have.

PR #188, opened 2022-02-03
increase nginx max replicas

opadron:
Increases max replicas by 4x, since the HPA has been oversubscribed by more than 3x.  This should leave a comfortable-enough margin for the replica count.


opadron:
@scottwittenburg should be minimal, as these pods run on the `base` pool.

PR #191, opened 2022-02-04
upgrade gitlab runners

opadron:


josephsnyder (approved):
I still don't quite get how to tell what gitlab-runner version is installed by the Helm chart but the Docker images are the latest for that category.

opadron:
You could use helm to see the defaults for the chart.  `helm show values` is the command you're looking for.

PR #193, opened 2022-02-05
Prometheus alertmanager hack

kotfic:
This makes sure all alert messages are picked up by the `AlertmanagerConfig` custom resource in the `monitoring` namespace.  Note that alerts originated in `source_namespace`

Note that It may be possible to do some post-alert label processing to rewrite source_namespace back to namespace. If so i would do that as a follow on PR.


kotfic:
This also includes the rule/alert for stuck pods in the pipeline namespace

opadron (approved):

PR #203, opened 2022-02-14
Send deprecated entries via config.

josephsnyder:
Eliminate the usage of several deprecated Helm variables in the
release.yaml file by setting the config.toml to contain those values
directly.

Deprecated options are marked here:
https://gitlab.com/gitlab-org/charts/gitlab-runner/-/blob/main/values.yaml
Commit SHA: 477f145bedcb68690d723826831d9c8c4f5b2d09

### DIFF
--- a/k8s/runners/huge-arm-pub/release.yaml
+++ b/k8s/runners/huge-arm-pub/release.yaml
@@ -31,26 +31,29 @@ spec:
     runners:
       config: |
         [[runners]]
+          requestConcurrency =  20
+          output_limit = 4096
           environment = ["FF_GITLAB_REGISTRY_HELPER_IMAGE=1"]
+          [runners.kubernetes]
+            helper_image = "gitlab/gitlab-runner-helper:arm-latest"
+            privileged = false
+            cpu_request = "46000m"
+            memory_request = "88Gi"
+            namespace = "pipeline"
+            pollTimeout = 600  # ten minutes
+            service_account = "runner"
+            [runners.kubernetes.node_selector]
+              "kubernetes.io/arch" = "arm64"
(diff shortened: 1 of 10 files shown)


josephsnyder:
I've marked this as a WIP with the thought that it'd be merged when all of the runner instances have been changed over to this new config section.  This release is applied to the cluster now (with a single `experimental` tag) and has examples of the running completing simple tasks here: https://gitlab.spack.io/spack/experimental/-/pipelines/99064/builds

opadron:
Hopping onto the runner pod, I don't see the `config.toml` reflects the settings.

josephsnyder:
> Hopping onto the runner pod, I don't see the `config.toml` reflects the settings.

The changes I make can be found in `config.template.toml`.  I'm pretty certain they are there!

josephsnyder:
@opadron, the concurrency is updated, both here and in the cluster.  When we get the go-ahead, I'll make these changes on the other runners.

opadron:
Looks good so far!  Go ahead and apply the changes to the other runners.

You might want to start with the staging runner, since it's a diff based on one of the production runners.  Comparing the two should give you an idea of what you'll need to change for the staging runner.

zackgalbreath:
The experimentation behind this PR created a lot of "disconnected" runners on gitlab.spack.io. I went through and manually deleted all of them. It would be good to verify that they automatically get cleaned up before we deploy this.

josephsnyder:
@opadron, the large-x86 instance we've been testing over de-registered correctly as the new one was properly registered.  I'm feeling confident in the creation/cleanup cycle.

I needed another overwrite as the "experimental" tag list snuck into the data.

PR #209, opened 2022-02-18
Increase concurrency for large jobs

opadron:


opadron:
@zackgalbreath note that I removed the `request_concurrency` setting while I was at it, since it doesn't make a difference in how wide we run.

zackgalbreath (approved):

PR #214, opened 2022-02-23
add cloud formation template for spot instances

opadron:


scripts/eks-node-group-with-spot.yaml
```diff
@@ -0,0 +1,314 @@
...
+          default: Instance Configuration
+        Parameters:
+          - NodeInstanceType
+          - NodeInstanceArch
+          - NodeGPUCount
+          - NodeInstanceProfile
+          - NodeImageId
+          - NodeVolumeSize
+
+    - Label:
```
josephsnyder, line 153: Just so we don't forget, this line is still misaligned!

josephsnyder (changes requested):

scripts/eks-node-group-with-spot.yaml
opadron: Thanks!

josephsnyder (approved):

PR #216, opened 2022-02-24
(SpackCIBridge) Stop posting duplicate statuses to commits

zackgalbreath:
Up until now, we've been unconditionally posting status to commits on GitHub. If we post too many statuses to a single commit then GitHub will not allow us to update it later. This causes a PR to be blocked until we manually intervene.

Other things this PR does:

* Print messages about our rate limit usage
* Cache calls to the ["get a commit" GitHub API endpoint](https://docs.github.com/en/rest/reference/commits#get-a-commit) in an effort to preserve our rate limit
* Suppress irrelevant errors emanating from `git log`
* Don't use GitLab CI to test draft PRs
* Reduce our cron job frequency from 3 minutes to 4 minutes in another attempt to preserve our rate limit

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -26,7 +26,7 @@ jobs:
           context: ./images/gh-gl-sync
           file: ./images/gh-gl-sync/Dockerfile
           push: true
-          tags: ghcr.io/spack/ci-bridge:0.0.24
+          tags: ghcr.io/spack/ci-bridge:0.0.25
       -
         name: Image digest
         run: echo ${{ steps.docker_build_sync.outputs.digest }}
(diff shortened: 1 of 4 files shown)


zackgalbreath:
Some examples of rate limit usage that I noticed. These shouldn't be taken as strictly "apples to apples" comparisons, but they all occurred during the same 6 hour window, so perhaps they're pretty close?

* The state of production prior to this PR: 322 API calls
* After caching calls to `get_commit()`: 250 API calls
* After introducing calls to `get_combined_statuses()`: 268 API calls

Again, these numbers shouldn't be taken as gospel, but they do provide evidence that this PR will reduce the number of times we have to hit GitHub's API while simultaneously fixing the problem of posting too many statuses to a single commit.

As this PR also reduces our polling frequency to every four minutes, we can afford an average of 5,000 / (60 / 4) = 333 API calls per run of this script.

scottwittenburg (approved):
We really needed a pass on this, thanks for taking the time @zackgalbreath!

scottwittenburg:
Hard to be 100% sure with this beast but the changes in `e3b3712`, "Better handling of draft PRs", look good to me.  My approval stands :standing_person:

PR #224, opened 2022-03-22
Add image and deployment for stuckpod mitigation

kotfic:
The substantive implementation is complete, tested and ready for review @opadron @zackgalbreath 

Still need too:
+ [x] add github job to build image 
+ [x] change image used in deployment
+ [x] briefly document

PR #226, opened 2022-03-23
Update Production

opadron:
Need to get #224 in prior to merging this.


kotfic (approved):
This looks good to me!

PR #227, opened 2022-03-25
Condense runners to two pools

josephsnyder:
One set of c5.4xlarege runners to handle all jobs.  Will still be called
glr-large-pub


k8s/runners/staging/release.yaml
```diff
@@ -67,4 +68,4 @@ data:
               "gitlab-ci/spack_spec_needs_rebuild" = "$SPACK_SPEC_NEEDS_REBUILD"
             [runners.kubernetes.node_selector]
               "kubernetes.io/arch" = "amd64"
-              "spack.io/node-pool" = "glr-xlarge-pub"
+              "spack.io/node-pool" = "glr-large-pub"
```
opadron, line 71: Let's have the testing and staging runners run on the same `glr-large-testing-pub` pool.  Be sure to set the pool's capacity to the sum of the two runners' `concurrency` settings + 8 instances of margin.

opadron (changes requested):
So far so good! :+1:

k8s/runners/staging/release.yaml
josephsnyder: Ah, I had this locally but didn't add it.  I think we have, 5 from testing + 3 from staging + 8 buffer = 16 which is the MaxSie right now.

opadron: Perfect!

opadron (approved):

opadron:
@josephsnyder as per the earlier conversation in chat, can we bump this up to 6 replicas instead of 4?

josephsnyder:
@opadron, done!

k8s/runners/staging/release.yaml
```diff
@@ -46,7 +46,8 @@ data:
           environment = ["FF_GITLAB_REGISTRY_HELPER_IMAGE=1"]
           [runners.kubernetes]
             privileged = false
-            cpu_request = "9000m"
+            cpu_request = "750m"
+            memory_request = "2Gi"
```
opadron, line 50: @kotfic could you add your changes to this runner?  This is the runner we use in staging.

opadron (approved):

PR #237, opened 2022-04-20
hotfix: add arm runners (production & testing)

opadron:
As requested from #aws-hack-week

Don't merge until @josephsnyder creates the new arm pools.


josephsnyder (approved):
The ARM stacks for both testing and production have been created and should be ready for runners.

PR #238, opened 2022-04-20
Look for stuck pipelines and run a new pipeline on the branch

scottwittenburg:
Query gitlab for running pipelines that were created more than
a day ago and try to run a new pipeline on the ref associated
with each stuck pipeline.  Also query for the pipelines child
pipelines, and try to cancel each one before finally canceling
the parent.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -231,3 +231,29 @@ jobs:
       -
         name: Image digest
         run: echo ${{ steps.docker_build_gitops.outputs.digest }}
+
+  unstick-pipelines:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Login to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      -
+        name: Build and push gitlab-clear-pipelines
+        id: docker_build_gitlab_clear_pipelines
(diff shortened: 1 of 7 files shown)


zackgalbreath (approved):
I presume you ran this manually as you were developing it. Just curious, roughly how many pipelines did this script unstick?

scottwittenburg:
Yes, I ran it manually.  But I didn't run it from the container yet, so in my testing I only let it do the actual cancel and re-run operations on a couple of the oldest stuck pipelines, leaving the rest for testing once the container is built (and before we merge main->production).

There were 38 running pipelines when I ran it this afternoon, 33 of them were older than one day, and the oldest of those was 37 days old.

opadron (approved):

scottwittenburg:
A couple more things I thought would be useful with this PR before merging and deployment: 

1. make "days old" a configurable threshold to control how far back we go just printing status before actually acting on pipelines
2. add a README.md briefly describing the issue this is working around 
3. It would be nice to know how often this has to clean up old pipelines.  I sort of recall discussing that there might eventually be a way for services like this to send metrics information somewhere that can host it for prometheus.  Is that hard to set up if we don't have it?

I don't know whether (3) is possible at this point, but (1) and (2) would be easy.

zackgalbreath:
> 3\. It would be nice to know how often this has to clean up old pipelines.  I sort of recall discussing that there might eventually be a way for services like this to send metrics information somewhere that can host it for prometheus.  Is that hard to set up if we don't have it?

I was thinking the same thing. Keeping track of how much cleaning this script really does would help prevent us from ignoring the problem going forward. @kotfic any advice on how to send custom metrics to prometheus?

scottwittenburg:
@opadrong @zackgalbreath WDYT?  Is it ok to add that animated GIF illustrating the problem this is working around?

PR #231, opened 2022-04-12
update ingress and prometheus releases so that they talk to each other

opadron:
Follows the instructions at

https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/#prometheus-and-grafana-installation-using-service-monitors

 1. Merge this PR
 2. Confirm that the changes worked
 3. Import Grafana dashboard as detailed near the bottom of the instructions linked above

PR #244, opened 2022-04-22
Ingress servicemonitor

kotfic:
This moves the unused service monitor from `ingress-nginx` namespace into `monitoring`  and sets up a basic off the shelf dashboard in Grafana


opadron:
Duplicate of #231?

kotfic:
Sorry somehow i guess i missed #231 completely.  I think i'd rather keep the service monitor in the monitoring namespace with everything else.  This also includes an ots dashboard which we can start using right away,  though i think we'll need a little more detailed dashboard soon.  Can we close #231 in favor of this?

opadron (approved):

PR #249, opened 2022-05-11
Prepare data migration

opadron:
Updates the postgresql subchart to run with 4 read replicas.  The hope is that these additional replicas will spread the IOPS load across multiple volumes, improving site reliability during peaks in demand.

In addition: a new `storageclass` based on `gp3` EBS volumes is planned.  The read replicas will be using this new storageclass, and the existing postgresql data storage will also be migrated to this new storageclass.

The details of getting this data migration right are very nuanced and addressing these details by hand is likely to be a very error-prone process, so this PR also includes a script that I wrote to automate this process.

When this PR is ready, and at a time of our choosing, we will need to take a planned outage and follow the checklist below to perform the necessary upgrades and migrations.

- [x] Suspend flux
- [ ] Merge this PR
- [ ] Merge `main` -> `production`
- [ ] Install CSI external snapshotter
    - needed for the CSI driver below
    - also might be useful if we ever decide to move to a kubernetes-native solution for managing snapshots
    - https://github.com/kubernetes-csi/external-snapshotter
- [ ] Install EBS CSI driver
    - This will be needed to support storage classes backed by EBS volume types other than `gp2`.
    - https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html
- [ ] Create/Install new storage class based on `gp3` EBS volumes
- [ ] Suspend gitlab and postgresql services.
- [ ] Create replacement PVC for postgresql
- [ ] Migrate postgresql data to the new PVC's PV using the `migrate-pvc.py` script.
- [ ] Create replacement PVC for staging postgresql
- [ ] Migrate staging postgresql data to the new PVC's PV using the `migrate-pvc.py` script.
- [ ] Resume all previously stopped services
- [ ] Confirm all services are operating properly
- [ ] Delete the old PV
- [ ] Merge all outstanding hotfix PRs to `production`
    - Several will likely be made as part of the above process
- [ ] Manually apply changes to the gitlab helm release
- [ ] Confirm new release deploys and operates properly
- [ ] Confirm new release includes additional postgresql read replicas
- [ ] Confirm all postgresql replicas run on the new `gp3`-based storage
- [ ] Confirm that the staging gitlab release applies the new needed patches.
- [ ] Resume flux

### DIFF
--- a/k8s/gitlab/release.yaml
+++ b/k8s/gitlab/release.yaml
@@ -223,5 +223,5 @@ spec:
         nodeSelector:
           spack.io/node-pool: gitlab
       persistence:
-        # storageClass: storageClassName
+        storageClass: gp3
         size: 100Gi
--- a/k8s/gitlab/release.yaml
+++ b/k8s/gitlab/release.yaml
@@ -172,8 +172,56 @@ spec:
           spack.io/node-pool: gitlab
 
     postgresql:
+      replication:
+        enabled: true
+        slaveReplicas: 4
+        synchronousCommit: "on"
+      shmVolume:
+        enabled: true
       master:
+        podLabels:
+          spack.io/environment: production
+          spack.io/release: gitlab
+          spack.io/app: postgresql
+        affinity:
+          podAntiAffinity:
+            requiredDuringSchedulingIgnoredDuringExecution:
+              - topologyKey: kubernetes.io/hostname
+                labelSelector:
+                  matchExpressions:
+                    - key: spack.io/environment
+                      operator: In
+                      values: [production]
+                    - key: spack.io/release
+                      operator: In
+                      values: [gitlab]
+                    - key: spack.io/app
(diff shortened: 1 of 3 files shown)


opadron:
I'm holding off on this PR because we will need to take a disruptive outage as part of the upgrade process, and I think we should have a discussion about when/how we would want to go about doing this.

opadron:
The main data migration completed successfully.  The transition to replicated postgresql database was successful with the following exception:

The slave replicas are not able to replicate the production DB.  Initial investigation suggests that there may be some issue specific to the bitnami postgresql image and/or storage volumes backed by anything other than `EBS gp2`.  The production DB is still operational and working fine, we just won't have proper replication until we figure out what's going wrong with that part.

See https://github.com/bitnami/bitnami-docker-postgresql/issues/180 for a description of the issue we are being effected by.

PR #257, opened 2022-05-23
Update Production

opadron:
- signing runners
 - protected runners
 - tried to setup distributed minio, but it didn't work (#256 needs to be merged before this PR)
 - notary images
 - other miscellaneous changes

@kotfic @scottwittenburg @zackgalbreath please have a look and make sure that everything in the change set is ready to go into production.


opadron:
Will undraft when #256 lands.

opadron:
@kotfic image didn't push.  I think we probably need to give you permission somewhere?

josephsnyder (approved):
My little change is there! I'll leave the others to comment on theirs

scottwittenburg (approved):
I don't see any issue with this, thanks.  Still probably good to let @kotfic and @zackgalbreath have a look before merging.

PR #263, opened 2022-05-26
Disambiguate tags and branches that have the same name

zackgalbreath:
This is necessary now that we are no longer prepending branch names with 'github/'


scottwittenburg (approved):
Thanks Zack :sweat_smile:

PR #267, opened 2022-05-27
add temporary stop-gap fix to spackbot (push to production)

opadron:


scottwittenburg (approved):
Easy to review when we've already seen it working!  Thanks @opadron :+1:

PR #273, opened 2022-06-16
install minio operator

opadron:
Installs the [minio operator](https://operator.min.io/), a component we plan to use for deploying a more robust and highly available object store.

This component follows the kubernetes operator pattern, where unlike helm charts or other template-based mechanisms, new controllers and CRDs are introduced allowing one to manage domain-specific resources (such as in this case, minio "tenants") as first-class kubernetes kinds.

With the operator installed, we can specify tenants like in the example below:

```YAML
---
apiVersion: minio.min.io/v2
kind: Tenant
metadata:
  name: my-minio-tenant
spec:
  credsSecret:
    name: my-minio-secrets
  pools:
    - servers: 1
      name: pool-0
      volumesPerServer: 4
      volumeClaimTemplate:
        spec:
          resources:
            requests:
              storage: 5Gi
```

I plan to follow up on this PR with new minio rollouts for staging and production based on this new operator.

### DIFF
--- a/k8s/kube-system/minio-operator/configmap.yaml
+++ b/k8s/kube-system/minio-operator/configmap.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: console-env
+  namespace: kube-system
+  annotations:
+    operator.min.io/authors: MinIO, Inc.
+    operator.min.io/license: AGPLv3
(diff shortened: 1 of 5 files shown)


opadron:
Note to reviewers: I'm leaving this PR/branch up in case the contents end up being useful.  But, if you do switch to just using an S3 bucket, you should probably disregard these changes.

PR #274, opened 2022-07-01
Support spackbot fixing style in worker process

scottwittenburg:
This PR adds a worker pod to the spackbot deployment which will handle
asynchronously the work of fixing isort style for PRs.  The worker pod
can be used for other asynchronous tasks in the future as well (e.g. PR
binary graduation), and we can increase the number of replicas if
necessary to handle the extra load.

This PR also adds a new, completely separate development version of spackbot
to run in the cluster (e.g. with its own name, https address, etc), which we can
use for testing changes to spackbot before going live with them.

This PR undoes the changes made in #267, now that we can fix the problem
in spackbot and deploy the change.


k8s/spack/spackbotdev-spack-io/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This directory contains what is basically a copy of spackbot to be used for development purposes.
+
+The development version of the app is registered as `spack-bot-test` under the [spack-test](https://github.com/organizations/spack-test) organization.  Also, it only has access to the [spack-test/spack](https://github.com/spack-test/spack) repo.
+
+## Workflow for testing with the development verion
+
+1. make local changes to `spack/spackbot` source code
+2. build, tag, and push new images for spackbot and spackbot-workers
+3. edit `k8s/spack/spackbotdev-spack-io/deployments` and update the `image` of the `spec[template][spec][containers]` elements to refer to the approprate test image tags
```
zackgalbreath, line 11: won't flux get in the way here?

scottwittenburg: Oh, hm.  I guess I didn't get hit by that this time because the development version of spackbot wasn't in main yet.  So then maybe the way to do it is to:

1. have spackbot gha workflow build and push production and development tags
2. here, we'd point production and development at those tags, respectively
3. during test cycle, don't change the images here, but just push new development images and do a rollout restart whenever you want to pick up and test a new image

What do you think?

zackgalbreath: That sounds good to me.

scottwittenburg: So I'm adding/pushing another tag on the spackbot side.  Those get pushed to `ghcr.io/spack/<image>:<tag>`, so it will be the case that anyone wanting to develop and test spackbot changes will need permission to push there.  Is that going to be a problem?  If so, would adding:

```
  annotations:
    fluxcd.io/ignore: "true"
```

... to the dev `deployments.yaml` of spackbot during testing allow us to change image names?  Or will flux not be able to ignore it if it exists in `main` without that annotation?

scottwittenburg: @opadron Do you have any thoughts on this?

opadron: So, the annotation does "what you probably expect" when flux sees it in the repo vs in the cluster.

In the repo, flux will simply not consider anything in that resource, so it won't have any effect.

In the cluster, flux will ignore *all* updates to the resource in question, even if the annotation is subsequently removed in the repo (it's still ignoring it in the cluster).

So, you're best bet would be to manually add the annotation to the live resource, and then manually remove it once you're ready for flux to take over, again.

scottwittenburg: Ok, thanks @opadron.  In that case, I think there's no need to push extra tags from the spackbot repo.  Since we'll have to disable flux on these two deployments in the cluster anyway (and re-enable afterwards), people can just push their development spackbot images wherever they want and point to them during development.

Does that sound reasonable to you?

zackgalbreath: this strategy sounds like it will work to me

tgamblin (approved):

PR #279, opened 2022-07-18
Don't run gh-gl-sync tests for flux automatic branches

zackgalbreath:


.github/workflows/spack_ci_bridge.yml
```diff
@@ -1,6 +1,10 @@
 name: spack_ci_bridge
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches-ignore:
+      - flux-sync
+      - flux-write-check-**
```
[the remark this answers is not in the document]
zackgalbreath: pretty sure it's a flux thing, I see "flux-sync" referenced here:

https://fluxcd.io/legacy/flux/faq/#why-does-my-ci-pipeline-keep-getting-triggered

.github/workflows/spack_ci_bridge.yml
```diff
@@ -1,6 +1,10 @@
 name: spack_ci_bridge
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches-ignore:
```
mvandenburgh, line 5: I think this should be
```suggestion
    tags-ignore:
```

I see a bunch of `flux-write-check-**` tags in https://github.com/spack/spack-infrastructure/tags, and https://fluxcd.io/legacy/flux/faq/#why-does-my-ci-pipeline-keep-getting-triggered mentions `flux-sync` is also a tag, not a branch.

mvandenburgh (changes requested):

.github/workflows/spack_ci_bridge.yml
```diff
@@ -1,6 +1,10 @@
 name: spack_ci_bridge
 
-on: [push, pull_request]
+on:
+  pull_request:
+    tags-ignore:
+      - flux-sync
+      - flux-write-check-**
```
mvandenburgh, line 7: Ah, it looks like `tags-ignore` is only supported for `push` events, not `pull_request` https://docs.github.com/en/github-ae@latest/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore, which is why there's no CI run for the latest commit. 

I think changing to

```suggestion
on:
  push:
    tags-ignore:
      - flux-sync
      - flux-write-check-**
    branches-ignore:
      - main
      # Maybe flux and production here as well?
```

would have the same effect, right?

zackgalbreath: Good catch. Or maybe just do `on: pull_request` and remove the CI on push altogether?

mvandenburgh (approved):

PR #280, opened 2022-07-20
Gitlab PostgreSQL RDS migration

mvandenburgh:
Gitlab's guide to using an external DB: https://docs.gitlab.com/charts/advanced/external-db/

First, we'll test out the RDS migration in staging. Note, because we likely do not want to maintain/pay for an additional postgres RDS instance for staging, an additional PR will be merged after this one that re-enables the in-cluster postgres DB for staging.

Steps:

- [x] Backup the gitlab staging database with the following command:
```sh
kubectl exec -it gitlab-staging-postgresql-master-0 -n gitlab -- bash -c "PGPASSWORD=<db_password> pg_dump -U gitlab -d gitlabhq_production" > dump.sql
```
- [x] Create an RDS database through the AWS console with the following settings:
  - Standard create
  - PostgreSQL (select latest version)
  - Master username: "gitlab"
  - Additional configuration: 
    - Initial database name: `gitlabhq_production`

- [x] Ensure the RDS DB has two users - `gitlab` and `postgres` by running the following SQL queries:
`CREATE ROLE postgres;`
`CREATE ROLE gitlab;`

- [x] Restore the backup to the RDS DB with the following command:
```sh
psql --host=<db url> --port=5432 --username=gitlab --password --dbname=gitlabhq_production -f dump.sql
```

- [ ] Merge this PR into `main`
	- Hopefully, this will trigger flux to make the changes in this PR and connect the gitlab staging deployment to the new database.

If all goes well in staging, proceed to the following nearly identical steps to apply the changes to production as well.

- [ ] Take down gitlab

- [ ] Backup the gitlab production database with the following command:
```sh
kubectl exec -it gitlab-postgresql-master-0 -n gitlab -- bash -c "PGPASSWORD=<db_password> pg_dump -U gitlab -d gitlabhq_production" > dump.sql
```
- [ ] Create an RDS database through the AWS console with the following settings:
  - Standard create
  - PostgreSQL (select latest version)
  - Master username: "gitlab"
  - Additional configuration: 
    - Initial database name: `gitlabhq_production`

- [ ] Ensure the RDS DB has two users - `gitlab` and `postgres` by running the following SQL queries:
`CREATE ROLE postgres;`
`CREATE ROLE gitlab;`

- [ ] Restore the backup to the RDS DB with the following command:
```sh
psql --host=<db url> --port=5432 --username=gitlab --password --dbname=gitlabhq_production -f dump.sql
```

- [ ] Create a PR merging `main` into `production` and merge it.

[...]

PR #283, opened 2022-08-02
Correctly specify hostname secret for RDS DB

mvandenburgh:
The gitlab `HelmRelease` doesn't support specifying the `host` of the postgres database as a secret (ref: https://docs.gitlab.com/charts/charts/globals.html#configure-postgresql-settings), so [`valuesFrom`](https://fluxcd.io/docs/components/helm/helmreleases/#values-overrides) must be used to do this. I believe this is why staging doesn't work after #280 was merged.

### DIFF
--- a/k8s/gitlab/release.yaml
+++ b/k8s/gitlab/release.yaml
@@ -11,6 +11,13 @@ spec:
     repository: https://charts.gitlab.io
     version: 5.6.3 # gitlab@14.6.3
 
+  valuesFrom:
+    - kind: Secret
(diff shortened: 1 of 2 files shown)


scottwittenburg (approved):
Interesting to know gitlab doesn't support specifying host this way.

mvandenburgh:
Looks like this didn't work. After this change, it appears gitlab doesn't even recognize the `host` parameter, and is falling back to the default dns name for the in-cluster db 😦

PR #284, opened 2022-08-02
Use a build matrix for building/pushing docker images

mvandenburgh:
I noticed the `custom_docker_builds` workflow has a lot of repeated commands. A [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) can be used here to shorten the workflow file and eliminate the repetition.


scottwittenburg (approved):
Thanks, this is nice.

PR #285, opened 2022-08-02
Fix missing array index in staging patch file

mvandenburgh:


k8s/gitlab/staging/release.yaml
```diff
@@ -71,7 +71,7 @@ data:
     - op: remove
       path: /spec/values/gitlab/gitlab-shell/service
     - op: replace
-      path: /spec/valuesFrom/name
+      path: /spec/valuesFrom/0/name
```
scottwittenburg, line 74: I'm happy to just approve this and see what happens, but is there a way to test the patch syntax locally to make sure it's correct?

scottwittenburg (approved):

k8s/gitlab/staging/release.yaml
mvandenburgh: yes, you can run the json patching script manually with
`python3 scripts/gitops-patch.py k8s/gitlab/release.yaml k8s/gitlab/staging/release.yaml`

mvandenburgh: Maybe it would be worth having a CI check that runs that script on every staging patch

scottwittenburg: Yeah, that might make sense.

PR #286, opened 2022-08-03
Fix format of HelmRelease secret values

mvandenburgh:
Another attempt following #283. https://fluxcd.io/legacy/helm-operator/helmrelease-guide/values/#secrets

### DIFF
--- a/k8s/gitlab/release.yaml
+++ b/k8s/gitlab/release.yaml
@@ -12,11 +12,11 @@ spec:
     version: 5.6.3 # gitlab@14.6.3
 
   valuesFrom:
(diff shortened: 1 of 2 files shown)


mvandenburgh:
@scottwittenburg #283 didn't work correctly, it seems the flux helm controller isn't picking up the secret correctly. Apparently the syntax for specifying secret values is slightly different for older versions of the flux helm controller, so I tried that here. I'm not sure what version we're running the cluster, if this PR doesn't work I'll have to take a deeper look at it

scottwittenburg (approved):

PR #287, opened 2022-08-04
Specify RDS DB hostname directly

mvandenburgh:
The database instance isn't publicly accessible, so this should be safe to do.


scottwittenburg (approved):
Since the database host is inside a vpc and not publicly accessible, I don't see why the name can't be specified here.

PR #288, opened 2022-08-04
Fix fluent-bit parsing

mvandenburgh:
Enabling the `Replace_Dots` setting in fluent-bit as suggested here https://github.com/fluent/fluent-bit/issues/4386 is apparently needed to avoid some log parsing issues. After deploying that change to the cluster and deleting the existing OpenSearch index, everything appears to be working now.
The `Trace_Error` setting enables more verbose logging for the fluent-bit pods themselves, and is what helped me debug this particular issue, so I think it's worth keeping that on as well.

### DIFF
--- a/k8s/logging/fluent-bit.yaml
+++ b/k8s/logging/fluent-bit.yaml
@@ -149,6 +149,8 @@ stringData:
         AWS_Auth        On
         AWS_Region      us-east-1
         Retry_Limit     6
+        Replace_Dots    On
+        Trace_Error     On
 
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


scottwittenburg (approved):
Is the image tag update related to the problem, or are you just updating to be safe?  Otherwise, this looks good to me, and now I can see at least one `spackbot-workers` log entry in there, so it seems it could be working.

mvandenburgh:
> Is the image tag update related to the problem, or are you just updating to be safe?

Just to be safe. I noticed we were using a really old image, probably because I copied most of that yaml from an AWS tutorial. I separated it out into its own commit to make this clearer.

scottwittenburg (approved):
Thanks @mvandenburgh :+1:

PR #294, opened 2022-08-10
Add script + CI workflow for validating staging JSON patches

mvandenburgh:
Following up on https://github.com/spack/spack-infrastructure/pull/285#discussion_r936096127

This should make it easier to catch mistakes in the staging JSON patches.

### DIFF
--- a/scripts/validate_staging_patches.py
+++ b/scripts/validate_staging_patches.py
@@ -3,42 +3,76 @@
 from pathlib import Path
 import subprocess
 import sys
+import yaml
+
+
+def get_yaml_path(name, namespace):
+    """Returns the path to a k8s YAML file given its name and namespace."""
+    for file in Path(__file__).parents[1].rglob("k8s/**/*.y*ml"):
+        if "staging" in str(file):
+            continue
+        with open(file) as fd:
+            for doc in yaml.safe_load_all(fd):
+                try:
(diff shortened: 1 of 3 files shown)


scripts/validate_staging_patches.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for patch_file in patch_files:
+
+        # Go to root of namespace directory (i.e. 'k8s/gitlab' for gitlab namespace yamls)
+        current_dir = patch_file.parent
+        while current_dir.parent.name != "k8s":
+            current_dir = current_dir.parent
+
+        # Run the gitops-patch.py script on each yaml in the namespace. If one of them succeeds,
+        # assume it is valid.
```
AlmightyYakob, line 20: Just for my own understanding, why is it that the success of one file constitutes the success of the entire directory?

mvandenburgh: The original idea was that if the staging yaml applied successfully to a single "regular" yaml, then it's valid. Which may not necessarily be true, but it gives a non-zero amount of confidence that all staging patches are valid. 
However, I've rewritten the script to explicitly apply each staging patch to their intended yaml, so this should now work 100% of the time.

mvandenburgh:
We are moving away from the custom gitops controller, so this is no longer needed.

PR #296, opened 2022-08-12
Start specifying resource requests

scottwittenburg:
Most of the things we deployed in the cluster do not specify the
amount of memory or cpu they request, although this information is
critical to allow kubernetes to make good pod scheduling decisions.  

This PR applies cpu and memory requests to some of the largest 
resource consumers that did not previously specify resource requests.

Using prometheus queries crafted by @kotfic, we were able to get
some measurements of what kind of resources are getting consumed
by the cluster services/apps now, and use that information to provide
these initial resource requests.  

We will need to follow up a few more PRs doing the same until we
have set requests for everything running in the cluster.

I'm not sure about the `k8s-bootstrap` changes, as to whether they'll
get picked up by flux automatically.  So we will have to keep an eye
on pods in the `flux` namespace to see if these requests show up in
the live objects, and figure out some action to take if not.

Links (also thanks to @kotfic):

https://sysdig.com/blog/kubernetes-resource-limits
https://itnext.io/k8s-monitor-pod-cpu-and-memory-usage-with-prometheus-28eec6d84729

### DIFF
--- a/k8s/aws-node-termination-handler/release.yaml
+++ b/k8s/aws-node-termination-handler/release.yaml
@@ -19,3 +19,8 @@ spec:
     emitKubernetesEvents: true
     enablePrometheusServer: true
     prometheusServerPort: 9092
+
+    resources:
+      requests:
+        cpu: 10m
+        memory: 200M
--- a/k8s/cdash/deployments.yaml
+++ b/k8s/cdash/deployments.yaml
@@ -30,6 +30,10 @@ spec:
             exec /bin/bash /docker-entrypoint.sh
         image: kitware/cdash:spack_deploy_v4
         imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            cpu: 1600m
+            memory: 1G
         ports:
         - containerPort: 80
           name: web
--- a/k8s/cert-manager/release.yaml
+++ b/k8s/cert-manager/release.yaml
@@ -8,13 +8,18 @@ spec:
   releaseName: cert-manager
   chart:
     name: cert-manager
-    repository: https://charts.jetstack.io                        
+    repository: https://charts.jetstack.io
     version: v1.0.2  # cert-manager@1.0.2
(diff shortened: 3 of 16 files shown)


scottwittenburg:
Another note: we'll have to take care when applying these changes, to verify that we have enough nodes available in the `base` and `beefy` node pools.  After this change takes effect, kubernetes will likely not be able to cram so many pods onto each node.

scottwittenburg:
@tgamblin If this doesn't solve all the problems spackbot is having (and there some evidence it might do that), it should at least allow us to see logs from old spackbot pods, which will help us understand what the actual problem is.

scottwittenburg:
> Is the idea for dev spackbot to catch bugs earlier with a smaller limits?

I pretty much just followed the usage stats from prometheus, and the dev spackbot just had less resource consumption, maybe because it hasn't been doing as much.

> Do you have the queries you did to list the list resource consuming pods (like in the article)?

I have my own queries I did to find the top offenders, but @kotfic put together the queries that produced the numbers I used (that produced over a hundred pods/containers, hence the focus on the top 10).  Here are the ones I adapted from the article:

Top 10 cpu usage without cpu requests:

```
topk(10,sum by (namespace,pod,container)(rate(container_cpu_usage_seconds_total{container!="",namespace!="pipeline"}[5m])) unless sum by (namespace,pod,container)(kube_pod_container_resource_requests{resource="cpu"}))
```

Top 10 memory usage without memory request:

```
topk(10,sum by (namespace,pod,container)(container_memory_usage_bytes{container!="",namespace!="pipeline"}) unless sum by (namespace,pod,container)(kube_pod_container_resource_requests{resource="memory"}))
```

scottwittenburg:
Removed the changes touching files in the k8s-bootstrap directory, so those can be handled separately.  See #297.

zackgalbreath (approved):

PR #297, opened 2022-08-15
Resource requests for bootstrapping

scottwittenburg:
Resource requests for things that need to be bootstrapped.


mvandenburgh:
I think this can be closed, as we no longer use these bootstrap resources (removed in #374)

PR #306, opened 2022-08-16
Use different queue names for develop vs production

scottwittenburg:
Spackbot now supports specifying the queue name as an environment
variable.  Supply different task queue names to develop vs prod
spack deployments to prevent workers from picking up jobs aimed
at the other deployment.


scottwittenburg:
Depends on [this](https://github.com/spack/spackbot/pull/65) spackbot change.

tgamblin (approved):

PR #311, opened 2022-08-24
Install prometheus service for gitlab CI pipeline metrics

mvandenburgh:


mvandenburgh:
We decided against this approach for pipeline metrics.

Issue #316, opened 2022-09-06
GitLab sometimes neglects to start a pipeline when a branch is pushed


zackgalbreath:
We have noticed a situation where our sync script pushes a branch to GitLab, but GitLab never starts a pipeline for it. This problem confusingly manifests itself as "pending - waiting for the style check to complete" on the relevant GitHub PR.

alecbcs:
Hey @zackgalbreath was this fixed by #317? Or are you still working on a more long term fix?

zackgalbreath:
> Hey @zackgalbreath was this fixed by #317? Or are you still working on a more long term fix?

#317 mitigates this issue, but it doesn't fully resolve it. GitLab is still periodically failing to start pipelines for us, but now we have a cronjob to detect these skipped pipelines & start them automatically every 30 minutes.

We like to keep track of mitigated issues like this so we can keep an eye out for upstream fixes & other opportunities to properly resolve them.

PR #317, opened 2022-09-06
Mitigation for skipped pipelines (issue #316)

zackgalbreath:


zackgalbreath:
This approach seems to work, but it's pretty slow (~9 - 10 minutes per run).

@kotfic has [some prior art](https://gitlab.spack.io/chris/pipeline-workbench/-/tree/main/watch) that directly queries the GitLab database instead of repeatedly querying the API.

@mvandenburgh do you think that approach is still sound after our move to AWS RDS? I'm not sure how to connect to that database to test / debug a script like this.

mvandenburgh:
> @mvandenburgh do you think that approach is still sound after our move to AWS RDS? I'm not sure how to connect to that database to test / debug a script like this.

Yeah, anything like that that worked before with the in-cluster DB should still work. The RDS database lives in the same VPC as the old one, so from the perspective of the components of the cluster that directly interact with the DB, nothing has really changed, it's still just a postgres database.

mvandenburgh (dismissed):
Sounds good to me. Since this solution is proven to work, I agree we should merge as-is and focus on higher priority things (as long as peformance is sufficient)

tgamblin:
Can this description be updated with rationale?

PR #329, opened 2022-10-11
Add ARM runners for GitLab staging deployment

mvandenburgh:


scottwittenburg (approved):
Glancing through, it looks good to me.  Thankfully it's just staging and we can iterate if something's off.

k8s/runners/staging/large-arm-pub/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+      value: 300  # five minutes
+    - op: replace
+      path: /spec/values/concurrent
+      value: 3
+    - op: replace
+      path: /spec/values/checkInterval
+      value: 1
+    - op: replace
+      path: /spec/values/runners/tags
+      value: "arm,aarch64,graviton,graviton2,small,medium,large,huge,{ENV},public,aws,spack"
```
mvandenburgh, line 40: FYI, to hopefully make this easier to review -
This file is identical to the staging patch for the x86 runner `release.yaml` with two differences -
1) All references to `runner-x86-*`, etc are replaced by `runner-arm-*` (obviously)
2) This line highlighted by this comment, i.e. the patch for `values.runners.tags`, was changed to the equivalent value from `k8s/runners/large-arm-pub/release.yaml`, but with `{ENV}` added in.

PR #326, opened 2022-10-06
Add a copy of aws-auth ConfigMap to version control

zackgalbreath:
Whenever modifying this ConfigMap in the cluster, please try to remember to also update this file by running:

`kubectl -n kube-system get cm aws-auth -o jsonpath='{.data.mapRoles}' > aws-auth.yaml`

PR #334, opened 2022-10-17
Update production

scottwittenburg:


mvandenburgh (approved):

scripts/aws-auth.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - system:nodes
+  - system:node-proxier
+  rolearn: arn:aws:iam::588562868276:role/R004-EKS-SELF-MANAGED-NG-IAM-PROF-NodeInstanceRole-SKR87YY9U7QU
+  username: system:node:{{EC2PrivateDNSName}}
+- groups:
+  - system:bootstrappers
+  - system:nodes
+  - system:node-proxier
+  rolearn: arn:aws:iam::588562868276:role/Spack-AmazonEKSFargatePodExecutionRole
+  username: system:node:{{SessionName}}
```
AlmightyYakob, line 16: What does this file do? Will this be managed by flux and reset any changes to the `aws-auth` configmap in the cluster? If so, there's a role that karpenter needs that's missing

```
- groups:
  - system:bootstrappers
  - system:nodes
  rolearn: arn:aws:iam::588562868276:role/KarpenterInstanceNodeRole
  username: system:node:{{EC2PrivateDNSName}}
```

scottwittenburg: @zackgalbreath can probably answer better, but IIRC, this file is only added so we know how to replace the values in the cluster if it ever disappears.  But maybe that means the missing role should be added anyway?  So this thing contains all the required bits.

zackgalbreath: That's right. See here for more context: https://github.com/spack/spack-infrastructure/pull/326#issuecomment-1270519312

AlmightyYakob: Gotcha. In that case, I think we should add it:
```suggestion
  username: system:node:{{SessionName}}
- groups:
  - system:bootstrappers
  - system:nodes
  rolearn: arn:aws:iam::588562868276:role/KarpenterInstanceNodeRole
  username: system:node:{{EC2PrivateDNSName}}
```

mvandenburgh: Will that file be picked up and applied by flux? I don't know off the top of my head if flux only track changes in the `k8s/` directory or the entire repo. If the latter, I guess we would need a `fluxcd.io/ignore: "true"` here.

AlmightyYakob: Is there any reason flux would need to ignore this file? Or are you just concerned about us needing to modify the configmap in place and flux resetting our changes?

mvandenburgh: > Is there any reason flux would need to ignore this file?

I believe @zackgalbreath's intent of committing that file was to have it solely for reference, not for flux to manage. Whether or not we want flux to manage it, and if that introduces any problems, etc i think would be a seperate discussion, I was just pointing out that this might not behave how we expect

AlmightyYakob: I see. In that case, I agree. It would probably be good to add the flux ignore to this file, just to be safe.

zackgalbreath (approved):

PR #336, opened 2022-10-19
Stop blocking PRs due to a currently running "main branch" pipeline

zackgalbreath:
Remove the 'waiting for base develop commit pipeline to succeed' functionality from our sync script. It doesn't jive well with '@spackbot run pipeline' and it will soon be unnecessary once we roll out PR binary graduation.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -27,7 +27,7 @@ jobs:
           - python-aws-bash
         include:
           - docker-image: gh-gl-sync
-            image-tags: ghcr.io/spack/ci-bridge:0.0.29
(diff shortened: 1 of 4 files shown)


scottwittenburg (approved):
This looks good to me, thanks @zackgalbreath! 

To be really effective at avoiding building duplicate hashes, this would have needed to avoid pushing to your PR branch if its `develop` parent was in the list including the currently running develop pipeline, *as well as any commits between there and the tip of `develop`*.  So as it was implemented, this wasn't even really a half-measure, but something less in most cases.  And the trouble it caused (allowing people to use spackbot to run pipelines on old code) wasn't worth it. 

Once we have PR binary graduation, that's where we'll really be able to avoid those duplicate builds anyway.

PR #338, opened 2022-10-21
Karpenter: Use consolidation instead of ttlSecondsAfterEmpty

AlmightyYakob:
Since we are now allowing almost all instance types to be provisioned with Karpenter, it's possible that a large node is provisioned for some temporary workload, and then is left running at a very low utilization, since pods are being scheduled onto it within the 10 minute TTL.

This PR changes our de-provisioning method to [Consolidation](https://karpenter.sh/v0.18.1/tasks/deprovisioning/#consolidation), which will optimize for cost by deprovisioning larger than necessary nodes.

### DIFF
--- a/k8s/karpenter/provisioners.yaml
+++ b/k8s/karpenter/provisioners.yaml
@@ -5,10 +5,12 @@ kind: Provisioner
 metadata:
   name: default
 spec:
-  # Terminate nodes after 10 minutes of idle time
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
Going to wait to merge this until the morning

zackgalbreath (approved):

PR #340, opened 2022-10-26
Explicitly list the EC2 instance families that we will use in CI

zackgalbreath:


AlmightyYakob:
It seems like that list of families must be almost all of them. Would it be easier to just exclude the ones that aren't allowed?

zackgalbreath:
> It seems like that list of families must be almost all of them. Would it be easier to just exclude the ones that aren't allowed?

Good point, that does make for a shorter list.

k8s/karpenter/provisioners.yaml
```diff
@@ -17,11 +17,40 @@ spec:
       memory: 20Ti
 
   requirements:
-    # Only spin up arm64 nodes
+    # Only spin up arm64 and amd64 nodes
     - key: "kubernetes.io/arch"
       operator: In
       values: ["arm64", "amd64"]
 
+    # Don't provision nodes from certain instance families.
```
AlmightyYakob, line 25: Sorry to be a stickler, but just so we have a record of why we're doing this
```suggestion
    # Don't provision nodes from certain instance families.
    # These are the instance families that don't support x86_v3
```

AlmightyYakob (dismissed):
Left one suggestion but otherwise looks good

AlmightyYakob (approved):

zackgalbreath:
For posterity's sake, here's some `archspec` data courtesy of #339 

[x86_64_instance_uarch.csv](https://github.com/spack/spack-infrastructure/files/9873642/x86_64_instance_uarch.csv)

PR #343, opened 2022-11-01
Separate provisioners per CPU architecture

zackgalbreath:
Also new testing runners to use the new provisioners. The current (old) provisioners and runners remain untouched by this change.

### DIFF
--- a/k8s/karpenter/provisioners/runners/graviton/3/provisioners.yaml
+++ b/k8s/karpenter/provisioners/runners/graviton/3/provisioners.yaml
@@ -1,0 +1,63 @@
+---
+# Provisioner for graviton3 gitlab runners
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: glr-graviton3
+spec:
+  providerRef:
+    name: default
+
+  # Consolidation will de-provision larger than necessary nodes
+  consolidation:
+    enabled: true
+
+  # Resource limits for this provisioner only
+  limits:
+    resources:
+      cpu: 5k
+      memory: 20Ti
+
+  requirements:
+      # Only spin up arm64 nodes
+    - key: "kubernetes.io/arch"
+      operator: In
+      values: ["arm64"]
+
+    # Only provision nodes with graviton3 processors.
+    # Data gleaned from https://aws.amazon.com/ec2/graviton/
+    - key: "karpenter.k8s.aws/instance-family"
+      operator: In
+      values:
+        - "c7g"
+
+    # Instance Size
+    - key: "karpenter.k8s.aws/instance-size"
+      operator: In
+      values:
+        - "medium"
+        - "large"
+        - "xlarge"
+        - "2xlarge"
+        - "3xlarge"
+        - "4xlarge"
+        - "6xlarge"
+        - "8xlarge"
+
+    # Availability Zones
+    - key: "topology.kubernetes.io/zone"
+      operator: In
+      values:
+        - "us-east-1a"
+        - "us-east-1b"
+        - "us-east-1c"
+        - "us-east-1d"
+
+    # Try spot, fall over to on-demand
+    - key: "karpenter.sh/capacity-type"
+      operator: In
+      values: ["spot", "on-demand"]
+
+  # Only provision nodes for pods specifying the glr-graviton3 node pool
+  labels:
+    spack.io/node-pool: glr-graviton3
--- a/k8s/karpenter/provisioners/runners/x86_64/v2/provisioners.yaml
+++ b/k8s/karpenter/provisioners/runners/x86_64/v2/provisioners.yaml
@@ -1,0 +1,56 @@
+---
+# Provisioner for x86_64_v2 gitlab runners
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: glr-x86-64-v2
+spec:
+  providerRef:
+    name: default
+
+  # Consolidation will de-provision larger than necessary nodes
+  consolidation:
+    enabled: true
+
+  # Resource limits for this provisioner only
+  limits:
+    resources:
+      cpu: 5k
+      memory: 20Ti
+
+  requirements:
+    # Only spin up amd64 nodes
+    - key: "kubernetes.io/arch"
+      operator: In
+      values: ["amd64"]
+
+    # Instance Size
+    - key: "karpenter.k8s.aws/instance-size"
+      operator: In
+      values:
+        - "medium"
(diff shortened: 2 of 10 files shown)


k8s/karpenter/provisioners/runners/x86_64/v3/provisioners.yaml
```diff
@@ -0,0 +1,114 @@
...
+  consolidation:
+    enabled: true
+
+  # Resource limits for this provisioner only
+  limits:
+    resources:
+      cpu: 5k
+      memory: 20Ti
+
+  requirements:
```
AlmightyYakob, line 21:
```suggestion
  requirements:
    # Only spin up amd64 nodes
    - key: "kubernetes.io/arch"
      operator: In
      values: ["amd64"]

```

k8s/karpenter/provisioners/runners/x86_64/v2/provisioners.yaml
```diff
@@ -0,0 +1,55 @@
...
+  consolidation:
+    enabled: true
+
+  # Resource limits for this provisioner only
+  limits:
+    resources:
+      cpu: 5k
+      memory: 20Ti
+
+    # Only spin up amd64 nodes
```
AlmightyYakob, line 21: I think the `requirements` section got dropped here

```suggestion

    requirements:
    # Only spin up amd64 nodes
```

k8s/karpenter/provisioners/runners/graviton/3/provisioners.yaml
```diff
@@ -0,0 +1,58 @@
...
+  consolidation:
+    enabled: true
+
+  # Resource limits for this provisioner only
+  limits:
+    resources:
+      cpu: 5k
+      memory: 20Ti
+
+  requirements:
```
AlmightyYakob, line 21:
```suggestion
  requirements:
      # Only spin up arm64 nodes
    - key: "kubernetes.io/arch"
      operator: In
      values: ["arm64"]

```

k8s/runners/testing/graviton/2/release.yaml
```diff
@@ -0,0 +1,86 @@
...
+              "metrics/spack_spec_needs_rebuild" = "$SPACK_SPEC_NEEDS_REBUILD"
+            [runners.kubernetes.node_selector]
+              "spack.io/node-pool" = "glr-graviton2"
+
+      # default image
+      image: "busybox:1.32.0"
+      imagePullPolicy: "if-not-present"
+      locked: false
+
+      tags: "testing2,arm,aarch64,graviton,graviton2,small,medium,large,huge"
```
AlmightyYakob, line 77: I'm confused about some of the tags shared by almost all of these runners. Namely:

* `testing2`
* `small`
* `medium`
* `large`
* `huge`

As is, a job could just be tagged with `huge`, and it would potentially run on any of these? Is that desired? If so, feel free to ignore this comment, but I just wanted to make sure of that behavior.

AlmightyYakob (changes requested):
Looks good overall, just a few fixes / suggestions. I also opted to specify the `kubernetes.io/arch` label in places where it may not be _strictly_ necessary, since I think the allowed instance types are all the desired architecture anyway. My rational for this is just to protect against accidentally including in invalid instance type, as well as just a sanity check.

AlmightyYakob:
> @AlmightyYakob I plan to locally apply these changes and kick off some test jobs to see if the new provisioners work the way they should. PTAL when you get a chance & let me know if you see any problems with this plan.

I think that sounds good, but one thing to note is that there is an existing testing provisioner configured already, defined [here](https://github.com/spack/spack-infrastructure/blob/cdceab2f135d5f23fda968987dd94413396af336/k8s/karpenter/provisioners.yaml#L78-L121).

I think you'd just want to delete that before applying this configuration by running
```
 kubectl delete provisioner testing
```

k8s/runners/testing/graviton/2/release.yaml
zackgalbreath: Yes, your understanding is correct. A job can only be served by a runner that matches all of its tags. But the inverse is not true. Extra tags on a runner do not prevent it from picking a job it otherwise would be able to run.

In the not-too-distant future, I think we should drop all of the small/medium/large/etc flags from our runners and the jobs that get generated by Spack, but that's a task for another PR.

zackgalbreath:
Here's some evidence that this actually works as intended:

* graviton2: https://gitlab.spack.io/zackgalbreath/test-project/-/jobs/4128681
* graviton3: https://gitlab.spack.io/zackgalbreath/test-project/-/jobs/4128684
* x86_64_v2: https://gitlab.spack.io/zackgalbreath/test-project/-/jobs/4128686
* x86_64_v3: https://gitlab.spack.io/zackgalbreath/test-project/-/jobs/4128738
* x86_64_v4: https://gitlab.spack.io/zackgalbreath/test-project/-/jobs/4128739

You'll notice that the `x86_64_v2` job was picked up by an `x86_64_v4` runner. This is because GitLab CI currently doesn't currently provide a notion of "runner priority".

We could remove the "backwards compatible" tags from our more modern runners, but cost-wise I'm not convinced this is necessary. Here's the current price different between comparable graviton2 and graviton3 instances in us-east-1 (N. Virginia):

* c6g.medium: $0.034 per hour
* c7g.medium: $0.0363 per hour

AlmightyYakob:
This looks good to me. Just a couple of questions:

* Is the plan to merge this PR with just testing, or are we going to update the main provisioners here as well? I see that the testing runners use non testing node-pool names
* Currently, each provisioner is able to provision `5k` CPU and `20Ti`. The total combined resources for all of these provisioners is _a lot_ . I'm not sure what values we want for each, but just wanted to point it out.

zackgalbreath:
Good questions.

> Is the plan to merge this PR with just testing, or are we going to update the main provisioners here as well? I see that the testing runners use non testing node-pool names

I see how this can be confusing. My motivation was to set up the new uarch provisioners for production (not testing) runners. But then I wanted to make sure they work before we cut over for real. Let's say that this experiment has been successful. I'll close this PR and open a new one with a new set of production runners that use these new provisioners.

> * Currently, each provisioner is able to provision `5k` CPU and `20Ti`. The total combined resources for all of these provisioners is _a lot_ . I'm not sure what values we want for each, but just wanted to point it out.

Noted. I definitely don't want to introduce an artificial bottleneck in Karpenter when compared to the number of jobs our GitLab CI runners can serve. Nonetheless, I'll revisit these numbers and come up with something more reasonable.

zackgalbreath:
Closing in favor of #345

PR #348, opened 2022-11-16
Add provisioner for base node pool

AlmightyYakob:
This adds the karpenter provisioner required for the `base` node pool.

It doesn't seem there's any code here for the autoscaling group associated with the base node pool, so I guess that will just need to be manually deprovisioned?

### DIFF
--- a/k8s/cluster-autoscaler/release.yaml
+++ b/k8s/cluster-autoscaler/release.yaml
@@ -1,3 +1,5 @@
+# TODO: If gitlab node-pool is ever migrated to karpenter, delete this file.
+# This is still necessary, even with the base node-pool gone, since CAS needs to run somewhere
 ---
 apiVersion: helm.fluxcd.io/v1
 kind: HelmRelease
--- a/k8s/karpenter/provisioners/base/provisioner.yaml
+++ b/k8s/karpenter/provisioners/base/provisioner.yaml
@@ -1,0 +1,31 @@
+---
+# Provisioner for base pods
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: base
+spec:
+  providerRef:
+    name: default
+
+  # Consolidation will de-provision larger than necessary nodes
+  consolidation:
+    enabled: true
+
+  limits:
+    resources:
+      cpu: 128 # Limit to 64 t3.small nodes (2 vCPUs each)
+
+  requirements:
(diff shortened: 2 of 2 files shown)


AlmightyYakob:
@zackgalbreath a few questions:

* I currently only see `t3.small` used in the base node group. Do you think it's fine to just keep this, or should we expand it?
* I restricted to only use on-demand, as I don't think we'll want to deal with the logistics of spot instances in the base node pool. However, if we are okay with that, we could likely save a decent amount of cost.
* I didn't enforce any limits yet on this, but from looking at the ASG in EC2, it seems like 64 is the max. Should I apply a resource limit of 64*`t3.small`?

zackgalbreath:
> * I currently only see `t3.small` used in the base node group. Do you think it's fine to just keep this, or should we expand it?
> * I restricted to only use on-demand, as I don't think we'll want to deal with the logistics of spot instances in the base node pool. However, if we are okay with that, we could likely save a decent amount of cost.
> * I didn't enforce any limits yet on this, but from looking at the ASG in EC2, it seems like 64 is the max. Should I apply a resource limit of 64*`t3.small`?

In order to reduce the number of variables, let's keep it the same as what we're currently doing with Cluster Autoscaler for the time being: up to 64 t3.small on-demand instances.

AlmightyYakob:
I think the migration plan here should be:

1. Merge this PR
2. Apply the provisioner
3. gradually scale down the max number of nodes in the base autoscaling group, making sure karpenter is picking up the pods and provisioning correctly
4. Delete the autoscaling group

AlmightyYakob:
> I think the migration plan here should be:
> 
> 1. Merge this PR
> 2. Apply the provisioner
> 3. Gradually scale down the max number of nodes in the base autoscaling group, making sure karpenter is picking up the pods and provisioning correctly
> 4. Delete the autoscaling group

@zackgalbreath How does this sound to you? If it sounds good, can I get an approval? (already addressed your above comment).

zackgalbreath:
> > I think the migration plan here should be:
> > 
> > 1. Merge this PR
> > 2. Apply the provisioner
> > 3. Gradually scale down the max number of nodes in the base autoscaling group, making sure karpenter is picking up the pods and provisioning correctly
> > 4. Delete the autoscaling group
> 
> @zackgalbreath How does this sound to you? If it sounds good, can I get an approval? (already addressed your above comment).

Sounds good to me!

zackgalbreath (approved):

PR #347, opened 2022-11-10
Add documentation for our "deferred pipelines" strategy

zackgalbreath:

PR #349, opened 2022-11-17
Defer GitLab pipelines based on main branch history

zackgalbreath:
Don't push PR branches to GitLab for pipeline testing if they are based on a commit of the main branch that is newer than what has already been tested by GitLab.

### DIFF
--- a/images/gh-gl-sync/SpackCIBridge.py
+++ b/images/gh-gl-sync/SpackCIBridge.py
@@ -35,15 +35,15 @@ class SpackCIBridge(object):
         self.sync_draft_prs = sync_draft_prs
         self.main_branch = main_branch
         self.currently_running_sha = None
-        self.currently_running_url = None
+        self.latest_tested_main_commit = None
 
         self.prereq_checks = prereq_checks
 
         dt = datetime.now(timezone.utc) + timedelta(minutes=-60)
         self.time_threshold_brief = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
 
-        # We use a longer time threshold (36 hours) to find the currently running main branch pipeline.
-        dt = datetime.now(timezone.utc) + timedelta(minutes=-2160)
+        # We use a longer time threshold (72 hours) to find the currently running main branch pipeline.
+        dt = datetime.now(timezone.utc) + timedelta(minutes=-4320)
         self.time_threshold_long = urllib.parse.quote_plus(dt.isoformat(timespec="seconds"))
 
         self.pipeline_api_template = gitlab_host
@@ -114,11 +114,6 @@ class SpackCIBridge(object):
         pulls = self.py_gh_repo.get_pulls(state="open")
         print("Rate limit after get_pulls(): {}".format(self.py_github.rate_limiting[0]))
         for pull in pulls:
-            if not pull.merge_commit_sha:
-                print("PR {0} ({1}) has no 'merge_commit_sha', skipping".format(pull.number, pull.head.ref))
-                self.unmergeable_shas.append(pull.head.sha)
-                continue
-
             backlogged = False
             push = True
             if pull.draft and not self.sync_draft_prs:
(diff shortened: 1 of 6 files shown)


zackgalbreath:
Replaces #347

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -153,10 +148,45 @@ def list_github_prs(self):
...
+                    if subprocess.run(
+                            ["git", "merge-base", "--is-ancestor", merge_base_sha, self.latest_tested_main_commit]
+                            ).returncode == 0:
+                        print(f"{pr_string}'s merge base IS an ancestor of latest_tested_main "
+                              "{merge_base_sha} vs. {self.latest_tested_main_commit}")
+                        try:
+                            subprocess.run(["git", "checkout", pr_string], check=True)
+                            subprocess.run(
+                                ["git", "merge", "-m", "Merge develop", self.latest_tested_main_commit],
+                                check=True)
```
scottwittenburg, line 177: I wonder it we need to add `--no-ff` here, to make sure we always get a real merge commit with two parents?  If it's at all possible we end up with a fast-forward merge, I don't think the check I've added [here](https://github.com/spack/spackbot/pull/70/files#diff-32369261a6ca2d1e63555fead2186d95cc1825c7fcf9eee26b0b00b47a95b3a4R84) will work as intended (i.e. the `parent_ids` will only contain a single commit sha, the one from `develop`).

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -252,26 +282,23 @@ def get_open_refspecs(self, open_prs):
         """Return lists of refspecs for fetch and push given a list of open PRs."""
```
scottwittenburg, line 282: Is this now just, e.g. "Return list of refspecs to push, given a list of open PRs"?

IIUC we no longer need to fetch refspecs for PRs because we had to do it already for each PR [here](https://github.com/spack/spack-infrastructure/pull/349/files#diff-822abeb2c9470f1fa0a4dd001219035b31166c1cf3a784f8248c38e93c3cdf2aR156)?

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -536,14 +567,21 @@ def sync(self):
             self.setup_git_repo()
 
             if self.main_branch:
-                # Find the currently running main branch pipeline, if any, and get the sha
+                # Find the currently running main branch pipeline, if any, and get the sha.
+                # Also get the latest commit on the main branch that has a completed pi
```
scottwittenburg, line 571: ...peline

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -293,10 +320,10 @@ def fetch_github_branches(self, fetch_refspecs):
         fetch_args = ["git", "fetch", "-q", "github"] + fetch_refspecs
         subprocess.run(fetch_args, check=True)
 
-    def build_local_branches(self, open_prs, protected_branches):
-        """Create local branches for a list of open PRs and protected branches."""
-        print("Building local branches for open PRs and protected branches")
-        for branch in open_prs["pr_strings"] + protected_branches:
+    def build_local_branches(self, protected_branches):
+        """Create local branches for a list of rotected branches."""
+        print("Building local branches for rotected branches")
```
scottwittenburg, line 325: missing the p in two instances above

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -536,14 +567,21 @@ def sync(self):
...
-                # Find the currently running main branch pipeline, if any, and get the sha
+                # Find the currently running main branch pipeline, if any, and get the sha.
+                # Also get the latest commit on the main branch that has a completed pi
                 main_branch_pipelines = self.get_pipelines_for_branch(self.main_branch, self.time_threshold_long)
                 for sha, pipeline in main_branch_pipelines.items():
+                    if self.latest_tested_main_commit is None and \
+                            (pipeline['status'] == "success" or pipeline['status'] == "failed"):
+                        self.latest_tested_main_commit = sha
+
                     if pipeline['status'] == "running":
```
scottwittenburg, line 578: Initially I didn't understand why we wouldn't need here a similar guard to the one above, where we make sure we don't set `self.currently_running_sha` if it's already set.  I guess maybe it's because this whole thing hinges on there only ever being a single running main branch pipeline at a time?

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -559,16 +597,20 @@ def sync(self):
             tags = self.list_github_tags()
 
             # Get refspecs for open PRs and protected branches.
-            open_refspecs, fetch_refspecs = self.get_open_refspecs(open_prs)
+            open_refspecs = self.get_open_refspecs(open_prs)
+            fetch_refspecs = []
             self.update_refspecs_for_protected_branches(protected_branches, open_refspecs, fetch_refspecs)
```
scottwittenburg, line 602: At first I didn't get why we create an empty `fetch_refspecs` here, and pass it into the method below.  Why leave support for `fetch_refspecs` in `update_refspecs_for_protected_branches` but remove it elsewhere?  But now I see that method adds refspecs to the list, associated with protected branches, so we fetch them below in `fetch_github_branches`.  :+1:

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -536,14 +567,21 @@ def sync(self):
             self.setup_git_repo()
 
             if self.main_branch:
-                # Find the currently running main branch pipeline, if any, and get the sha
+                # Find the currently running main branch pipeline, if any, and get the sha.
+                # Also get the latest commit on the main branch that has a completed pi
                 main_branch_pipelines = self.get_pipelines_for_branch(self.main_branch, self.time_threshold_long)
                 for sha, pipeline in main_branch_pipelines.items():
+                    if self.latest_tested_main_commit is None and \
```
scottwittenburg, line 574: Looks like 36 hrs is the long time threshold.  If that's what we wanted when looking for the currently running pipeline, is it still enough, now that we're looking for the latest completed one?

scottwittenburg: I don't know one way or another yet, just thinking out loud.

images/gh-gl-sync/SpackCIBridge.py
zackgalbreath: doesn't hurt to be safe, I'll add this check

images/gh-gl-sync/SpackCIBridge.py
zackgalbreath: 36 hours is *probably* safe to get the most recent completed develop pipeline, but I don't see any harm in doubling that either, so I'll do so now.

images/gh-gl-sync/SpackCIBridge.py
zackgalbreath: Thanks your noticing this. Yes, your understanding is correct for why we no longer need to populate `fetch_refspecs` for open PRs.

images/gh-gl-sync/SpackCIBridge.py
zackgalbreath: Yeah, I'm not in love with how these functions/variables have evolved, but it seems like you managed to correctly figure out the current state of things.

images/gh-gl-sync/SpackCIBridge.py
zackgalbreath: I think you're right that we want to do `--no-ff` here. Let's experiment with it to make sure there aren't any unintended side-effects.

PR #350, opened 2022-11-28
Fix memory limits for runner provisioners

AlmightyYakob:
It seems the math for each runner provisioner's memory limits was low by a factor of 2. This PR essentially doubles the memory limits for each runner provisioner (except testing, which was correct).

This has has an actual impact on the `x86-64_v4` provisioner, as that's been hitting memory limits due to this, but the other provisioners haven't gotten at all close to their limits. Nonetheless, I felt it was correct to bump them to what was intended. I wouldn't be opposed to decreasing them, but that wasn't the intent of this PR.

### DIFF
--- a/k8s/karpenter/provisioners/runners/graviton/2/provisioners.yaml
+++ b/k8s/karpenter/provisioners/runners/graviton/2/provisioners.yaml
@@ -16,7 +16,7 @@ spec:
   limits:
     resources:
       cpu: 3840 # 16 vCPUs * 6 replicas * 20 concurrent jobs * 2 runners (public & protected)
-      memory: 7680Gi # 64 Gi * 6 replicas * 20 concurrent jobs * 2 runners (public & protected)
+      memory: 15Ti # 64 Gi * 6 replicas * 20 concurrent jobs * 2 runners (public & protected)
 
   requirements:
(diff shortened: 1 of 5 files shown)


zackgalbreath (approved):

AlmightyYakob:
> Another thing we could (should?) do is to play around with [`check_interval`](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#how-check_interval-works) to "slow down" the x86_64_v4 / graviton3 runners & give the other runners a greater probability of picking up jobs they can handle.

Interesting. What makes you think the `x86-64_v4` / `graviton3` runners are getting higher priority than other runners?

zackgalbreath:
> Interesting. What makes you think the `x86-64_v4` / `graviton3` runners are getting higher priority than other runners?

The fact that it was the `v4` provisioner that maxed out its memory limits. I've also anecdotally seen the `graviton3` provisioner fail jobs because of spot capacity issues (it consists of a single EC2 instance family).

I could be wrong, but it seems like "spreading the work out" better among our available runners would help us avoid bottlenecks like these. Unfortunately GitLab has no concept of runner priority, so fiddling with `check_interval` is the best tool we have AFAIK.

PR #355, opened 2022-12-01
Karpenter provisioner fixes

zackgalbreath:
* Allow Karpenter to spin up bigger instances
  * c7g.8xlarge is apparently not big enough to build ParaView
* Update node pool for signing runner
  * glr-large-pub doesn't exist anymore


AlmightyYakob (approved):
Those are some large instance sizes :sweat_smile:

PR #356, opened 2022-12-06
Add memory requests to gitlab helper containers

mvandenburgh:
Adds a memory request of 512MB for `gitlab-runner-helper` init containers. 512MB is just a guess on my part, based on [what the gitlab runner helper does](https://stackoverflow.com/questions/67620001/what-is-gitlab-gitlab-runner-helper-docker-image-used-for), and may not be ideal. However, setting it can't hurt and is better than what we have now (nothing), and will hopefully help mitigate the OOMKilled issue until we can get an idea of the actual memory requirements of these containers (which will in turn help us determine suitable memory _limits_ to put on this container)

### DIFF
--- a/k8s/runners/protected/graviton/2/release.yaml
+++ b/k8s/runners/protected/graviton/2/release.yaml
@@ -35,6 +35,7 @@ spec:
           [runners.kubernetes]
             helper_image = "gitlab/gitlab-runner-helper:arm-latest"
             privileged = false
+            helper_memory_request = "512m"
 
             cpu_request = "750m"
             cpu_request_overwrite_max_allowed = "16"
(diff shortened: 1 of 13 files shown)


danlamanna:
FWIW I noticed these error messages in the logs of a runner pod:
```
W0123 15:50:07.923547       7 warnings.go:70] spec.initContainers[0].resources.requests[memory]: fractional byte value "512m" is invalid, must be an integer                                   │
│ W0123 15:50:07.923627       7 warnings.go:70] spec.containers[1].resources.requests[memory]: fractional byte value "512m" is invalid, must be an integer  
```
 I think `512m` is invalidly formatted, though I haven't gotten a chance to look at it.

zackgalbreath:
#754

PR #364, opened 2022-12-19
Add pipline label to all runner provisioners

AlmightyYakob:
This adds the `spack.io/pipeline: true` label to all pipeline provisioners. This was something that was missed during the karpenter switchover.


kotfic (approved):
LGTM

PR #375, opened 2023-01-27
Update karpenter's AWSNodeTemplate

zackgalbreath:
This changes were discovered by trial & error while attempting to get karpenter to provision nodes with attached EBS volumes

### DIFF
--- a/terraform/karpenter.tf
+++ b/terraform/karpenter.tf
@@ -81,11 +81,11 @@ resource "kubectl_manifest" "karpenter_node_template" {
       name: default
     spec:
       subnetSelector:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: "true"
       securityGroupSelector:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: "spack-production"
       tags:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: spack-production
       blockDeviceMappings:
         - deviceName: /dev/xvda
           ebs:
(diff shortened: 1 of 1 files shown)


zackgalbreath:
In case it helps, here's the section of the `node-templates.yaml` file that I manually edited & applied to the cluster while attempting to get this functionality to work again:

```yaml
  securityGroupSelector:
    karpenter.sh/discovery: 'spack-production'
  subnetSelector:
    karpenter.sh/discovery: 'true'
  tags:
    karpenter.sh/discovery: spack-production
```

terraform/karpenter.tf
```diff
@@ -81,11 +81,11 @@ resource "kubectl_manifest" "karpenter_node_template" {
       name: default
     spec:
       subnetSelector:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: "true"
       securityGroupSelector:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: "spack-production"
```
mvandenburgh, line 86:
```suggestion
        karpenter.sh/discovery: ${module.eks.cluster_name}
```
These are equivalent, while also avoiding hardcoding the cluster name in multiple places.

terraform/karpenter.tf
```diff
@@ -81,11 +81,11 @@ resource "kubectl_manifest" "karpenter_node_template" {
...
       tags:
-        karpenter.sh/discovery: ${module.eks.cluster_name}
+        karpenter.sh/discovery: spack-production
```
mvandenburgh, line 88:
```suggestion
        karpenter.sh/discovery: ${module.eks.cluster_name}
```

See above comment

mvandenburgh (changes requested):
I think the only needed change here is setting the subnet selector one to `true`, the other changes don't have any effect.

mvandenburgh (approved):

PR #376, opened 2023-01-27
Add Karpenter provisioners for gitlab and beefy node pools

mvandenburgh:
I already applied these during the migration so that we could avoid having to bring in cluster-autoscaler for the new cluster. 

I basically just looked at the Auto-Scaling groups in AWS to figure out what kind of instance types cluster-autoscaler was using for these node pools; I'm not sure if this is an ideal set up, or if we want to add more instance types for these provisioners, etc. but I'll leave that to @AlmightyYakob.


AlmightyYakob (approved):
I think this is good for now. Incorporating lots of instance types is mainly a concern if we care about utilizing spot instances, which we probably don't want in this case for the sake of stability.

PR #381, opened 2023-02-02
Make flux ignore metabase pvc yaml

mvandenburgh:
This PVC is automatically created by metabase. Since it's bound to a PV, you can't do a `kubectl apply` on it.  Flux didn't create this resource and doesn't realize it already exists, so it's trying to do a `kubectl apply` anyways and is failing (at least, that's my theory). I've already applied this YAML to the cluster, and it fixed the error.

### DIFF
--- a/k8s/metabase/postgres-pvc.yaml
+++ b/k8s/metabase/postgres-pvc.yaml
@@ -4,6 +4,8 @@ kind: PersistentVolumeClaim
 metadata:
   name: metabase-postgresql
   namespace: monitoring
+  annotations:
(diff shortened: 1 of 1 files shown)


mvandenburgh:
BTW, `kustomize.toolkit.fluxcd.io/reconcile: disabled` is roughly equivalent to the old `fluxcd.io/ignore: "true"` from Flux 1. 

https://fluxcd.io/flux/components/kustomize/kustomization/#reconciliation

mvandenburgh:
This is actually needed when Flux is bootstrapping metabase for the first time. The problem was, i had manually ran `kubectl apply` when I needed to let flux do it.

PR #385, opened 2023-02-08
Update the ID for UO's intermediate signing public key

zackgalbreath:
This should hopefully fix our failing signing jobs


mvandenburgh (approved):
Should we also bump the version of the `notary` image? I'm not sure where to check if the notary job pull policy is set to `Always` or `IfNotPresent`.

zackgalbreath:
> Should we also bump the version of the `notary` image? I'm not sure where to check if the notary job pull policy is set to `Always` or `IfNotPresent`.

It looks like we don't have to; I just checked and it indeed is `imagePullPolicy: "always"` (k8s/runners/signing/release.yaml)

PR #386, opened 2023-02-08
Reusable `spack` terraform module + staging deployment

mvandenburgh:


terraform/staging/main.tf
```diff
@@ -0,0 +1,112 @@
...
+  exec {
+    api_version = "client.authentication.k8s.io/v1beta1"
+    args        = ["eks", "get-token", "--cluster-name", module.staging_cluster.cluster_name, "--region", "us-west-1"]
+    command     = "aws"
+  }
+}
+
+
+provider "github" {
+  owner = "mvandenburgh" # TODO: change this
```
AlmightyYakob, line 74: Should this be changed as a part of this PR?

terraform/production/main.tf
```diff
@@ -91,8 +75,36 @@ provider "github" {
...
+  private_subnets = [
+    "10.0.64.0/19",
+    "10.0.96.0/19",
+    "10.0.128.0/19",
+    "10.0.160.0/19",
+  ]
+  database_subnets = [
+    "10.0.192.0/19",
+    "10.0.224.0/19",
+  ]
```
AlmightyYakob, line 101: I'm wondering if we can provide defaults for some of these variables in the module? Mainly thinking of the following

- `single_nat_gateway`
- `vpc_cidr`
- `public_subnets`
- `private_subnets`
- `database_subnets`

Don't feel too strongly about it, just a thought.

terraform/staging/main.tf
```diff
@@ -0,0 +1,112 @@
...
+    "10.0.128.0/19",
+    "10.0.160.0/19",
+  ]
+  database_subnets = [
+    "10.0.192.0/19",
+    "10.0.224.0/19",
+  ]
+
+  flux_repo_name   = "spack-infrastructure"
+  flux_repo_owner  = "mvandenburgh"
```
AlmightyYakob, line 105: Similar to my other comment, shouldn't this point to the `spack/spack-infrastructure` repo?

AlmightyYakob (commented):
Great work! Looks good overall, just a few questions/comments.

terraform/staging/main.tf
mvandenburgh: Flux is still pointed at https://github.com/mvandenburgh/spack-infrastructure/tree/flux2, since we need a github personal access token from spackbot from Todd in order to configure it with `spack/spack-infrastructure`. I didn't want to block this PR on that so I just left it as-is.

terraform/production/main.tf
mvandenburgh: I went back and forth on this, and ultimately decided that since we're only using this module in two places currently, the benefits of having explicit parameters all in one place outweighed the ease of reusability that having defaults would provide. I don't feel strongly either, though, especially if there's an argument for including defaults that I didn't consider.

terraform/staging/main.tf
mvandenburgh: (see https://github.com/spack/spack-infrastructure/pull/386#discussion_r1103167404)

AlmightyYakob (approved):

PR #387, opened 2023-02-08
Make notary job export public (reputational) key

scottwittenburg:


scottwittenburg:
If we can have this, then right after it runs, the signing jobs can push the public key to the mirror.  That is accomplished in this spack [PR](https://github.com/spack/spack/pull/35409).

scottwittenburg:
@kotfic @zackgalbreath This (along with the spack PR mentioned in the comment above) is what I have come up with so far for making sure the public portion of the reputational key gets installed in the mirror. 

How does this feel to you?  Any other ideas?

kotfic (approved):
LGTM

PR #394, opened 2023-02-21
Update SMTP username

mvandenburgh:
It seems that the password for the SMTP credentials used by GitLab and Metabase that's currently stored as a secret in the cluster is not correct. So, I regenerated the iam access key, which also resulted in the user name getting changed; this PR updates the GitLab helm release with this new user name.


scottwittenburg (approved):
Approving since it seems to be working, and I see this PR didn't introduce the inclusion of the access key id in the yaml.

PR #396, opened 2023-02-22
Fix syntax error

mvandenburgh:
Flux failed to apply the changes from #390 due to a minor syntax error - `spec.containers` is a `list`, not a `map`.


AlmightyYakob (approved):
my bad, good catch

PR #397, opened 2023-02-22
Replace Role with ClusterRole for secrets backup script

AlmightyYakob:
Since the secrets backup cron job attempts to list secrets across _all_ namespaces, the service account it uses requires a binding to a cluster role.

### DIFF
--- a/k8s/secrets-backup/service-accounts.yaml
+++ b/k8s/secrets-backup/service-accounts.yaml
@@ -8,26 +8,25 @@ metadata:
     eks.amazonaws.com/role-arn: arn:aws:iam::588562868276:role/SecretsBackupRole-production
 
 ---
-kind: Role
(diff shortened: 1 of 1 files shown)


mvandenburgh (commented):
Can the `Role` and `RoleBinding` be removed? I believe the difference is that `ClusterRole` allows access across all namespaces, while `Role` is scoped to a single namespace.

AlmightyYakob:
> Can the `Role` and `RoleBinding` be removed? I believe the difference is that `ClusterRole` allows access across all namespaces, while `Role` is scoped to a single namespace.

I was going to leave it there to be safe, but you're right, there's really no reason to keep it there. I also just realized that we'll also need to add permissions to access config maps, so I'll do that as well.

mvandenburgh (approved):
LGTM

PR #403, opened 2023-03-02
Improvements to docker publish workflow

mvandenburgh:
Fixes some minor inconveniences/issues that I happened to notice in our docker build workflow.

- Removes the publishing of new images on push to the `production` branch. We don't use `production` for change control anymore so it is no longer "special" and shouldn't trigger new image builds.
- Removes some redundant entries in the `matrix` object - all of the removed items are duplicated in the `include` object.
- Limits the docker publish workflow to only run when an image is actually changed. We're currently wasting cycles rebuilding and uploading every docker image, even if none of them change.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - images/**
+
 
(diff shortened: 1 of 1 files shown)


AlmightyYakob (approved):
This all makes sense to me. Good work!

PR #404, opened 2023-03-03
Configure staging cluster using Kustomize

mvandenburgh:
I haven't applied the terraform changes to production yet; since we're making a change to the Flux config, it will require extra care so that we don't disrupt anything. I have applied the terraform changes to staging though, and the Kustomizations work as intended.

Some extra care needs to be taken when merging this, since we're touching the Flux configuration. Steps that need to be taken to apply this:
- [x] _Before merging this PR_, disable Flux on the production cluster
- [x] Merge this PR
- [ ] Apply the terraform changes to production
- [ ] Restart Flux pods in production
	- Last time I changed something in Flux (in #402), I needed to restart the various Flux `*-controller` pods in order for the changes to take effect.
- [ ] Re-enable Flux on the production cluster


scripts/generate_kustomizations.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def _gen_kustomization_manifest(resources: list[str]):
+    return {
+        "apiVersion": "kustomize.config.k8s.io/v1beta1",
+        "kind": "Kustomization",
+        "resources": resources,
+    }
+
+
```
AlmightyYakob, line 16: I think a comment here explaining what this script is doing would be good, just for posterity.

AlmightyYakob (commented):
This all makes sense and looks good to me, but I'm somewhat hesitant to say I approve, mainly because there's just a lot of changes here, and even ignoring the moved files, something could easily slip through the cracks. Do you see any reasonable way to test this?

terraform/staging/main.tf
```diff
@@ -118,9 +118,9 @@ module "staging_cluster" {
   ]
 
   flux_repo_name   = "spack-infrastructure"
-  flux_repo_owner  = "mvandenburgh"
-  flux_branch      = "staging"
-  flux_target_path = "k8s/"
+  flux_repo_owner  = "spack"
+  flux_branch      = "kustomize"
```
AlmightyYakob, line 122: Should this be `main`?

mvandenburgh: Yeah sorry that was a temporary thing to test this and I forgot to pop that commit off. Fixed now.

k8s/staging/kube-system/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../production/kube-system/metrics-server.yaml
```
AlmightyYakob, line 4: What does this accomplish? Is it just using the same configuration as prod?

If a matching `kustomization.yaml` file is not provided for staging, does it get run on staging?

mvandenburgh: Yup, it's just using the same configuration as prod. If a YAML is not listed in a `kustomization.yaml`, it doesn't get run on staging.

AlmightyYakob: The only concern I have is reproducibility between prod and staging, since we're running a subset of the services in staging. However, if you don't see any issue there, then it's fine by me.

AlmightyYakob (dismissed):

AlmightyYakob (approved):

PR #409, opened 2023-03-17
Add more regexes to reduce 'other' category

scottwittenburg:


scottwittenburg:
I'd like to update the deployment associated with this, but something seems out of sync.  For one thing, I don't see `upload-gitlab-failure-logs` appearing anywhere in `k8s`, so how is this running at all?  Also, it seems like the `docker-image` and `include` lists in the `custom_docker_builds.yml` got out of sync, as those lists are parallel until somewhere near the bottom.

mvandenburgh:
> I'd like to update the deployment associated with this, but something seems out of sync. For one thing, I don't see `upload-gitlab-failure-logs` appearing anywhere in `k8s`, so how is this running at all? 

There isn't a `Deployment` associated with the `upload-gitlab-failure-logs`; instead, it runs as a `Job`.

We have a FastAPI server running as [a `Deployment`](https://github.com/spack/spack-infrastructure/blob/main/k8s/custom/gitlab-error-processor/deployments.yaml). GitLab is configured to send a webhook to this server whenever a job finishes. Upon receiving a web hook from GitLab, the server kicks off a `upload-gitlab-failure-logs` job (see https://github.com/spack/spack-infrastructure/blob/main/images/gitlab-error-processor/app.py#L50), which uses the job template [here](https://github.com/spack/spack-infrastructure/blob/main/images/gitlab-error-processor/job-template.yaml).

> Also, it seems like the `docker-image` and `include` lists in the `custom_docker_builds.yml` got out of sync, as those lists are parallel until somewhere near the bottom.

This will be fixed by #403. Looks like it's just waiting to be merged, I'll do that now

mvandenburgh (approved):

PR #414, opened 2023-03-20
Manage elasticache engine/version with Terraform

mvandenburgh:
Will allow us to bump the elasticache/redis version more easily. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_cluster#engine_version


mvandenburgh:
Sorry, I should have explained the rationale for this better.

> Will this result in an upgrade on merge? What version are we currently using?

No, we're currently on version 7. According to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_cluster#engine_version, terraform defaults to the latest version if not specified. So once version 8 is released, running `terraform apply` will suddenly result in an elasticache update. This change encodes the version so that we can explicitly bump it when needed.

AlmightyYakob (approved):

PR #415, opened 2023-03-21
Binary cache pruning tools and image for spack

kwryankrattiger:
Some tools used for pruning. Once these scripts are proven they could probably be integrated into a cron job type of deployment, but for now they are here as a record of the strategy for determining which hashes to prune.

* Makefile
  - original script developed by @haampie
* common.functions
  - some tools used by the pruning tools
* prunable-files.functions 
  - Makefile translated to bash with additional checks for more robust fetching and error checking when building lists of hashes
* s3-prune.sh 
  - script for batch removing a list of files from s3
* pruner
  - script that builds the list of hashes to prune and removes them


kwryankrattiger:
https://github.com/haampie/spack-prune-specs

images/ci-prune-buildcache/ci_buildcache_prune.py
```diff
@@ -0,0 +1,377 @@
...
+    stack = pipeline_job.name.replace("-generate", "")
+
+    job = project.jobs.get(jid, lazy=True)
+    try:
+        artifact_path = "jobs_scratch_dir/concrete_environment/spack.lock"
+        artifact = job.artifact(artifact_path)
+        lock = json.loads(artifact)
+        return stack, [h for h in lock["concrete_specs"].keys()]
+    except gitlab.exceptions.GitlabHttpError:
+        return []
```
scottwittenburg, line 109: Given that you return empty list in this particular exception case, how do you know if/when you're hitting it and that you might not be getting all the data?

images/ci-prune-buildcache/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM fedora:40
```
mvandenburgh, line 1: Is there a reason to not just use the standard `python` base image we usually use for python-based cron jobs?

k8s/production/custom/prune-buildcache/cron-jobs.yaml
```diff
@@ -0,0 +1,40 @@
...
+    spec:
+      activeDeadlineSeconds: 1200 # terminate any running job after 20 minutes
+      backoffLimit: 0
+      template:
+        spec:
+          serviceAccountName: prune-buildcache
+          restartPolicy: Never
+          containers:
+          - name: pruner
+            image: ghcr.io/spack/ci-prune-buildcache:0.0.1
```
mvandenburgh, line 20: You'll also need to add an entry to the job matrix in this workflow https://github.com/spack/spack-infrastructure/blob/main/.github/workflows/custom_docker_builds.yml to build this image.

images/ci-prune-buildcache/ci_buildcache_prune.py
```diff
@@ -0,0 +1,490 @@
...
+import multiprocessing.pool as pool
+import os
+from io import BytesIO
+from urllib.parse import urlparse
+
+import buildcache
+import helper
+from pruner import DirectPruner, IndexPruner, OrphanPruner
+
+# Script Config
```
mvandenburgh, line 17: We should configure sentry here as well, so we get notified of any errors.

k8s/production/custom/prune-buildcache/service-accounts.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prune-buildcache
+  namespace: custom
+  annotations:
+    eks.amazonaws.com/role-arn: arn:aws:iam::588562868276:role/DeleteAndPutObjectsFromBucketSpackBinaries
```
mvandenburgh, line 8: I think this service account would be better off as a terraform `kubectl_manifest`, as we could then just reference this IAM role with TF instead of duplicating the ARN here.

mvandenburgh (changes requested):
k8s stuff looks good for the most part, just a few minor missing things/questions

images/ci-prune-buildcache/ci_buildcache_prune.py
mvandenburgh: Fixed in https://github.com/spack/spack-infrastructure/pull/415/commits/74a0696e3dc472041945f5a9162b54969ed7b004

k8s/production/custom/prune-buildcache/service-accounts.yaml
mvandenburgh: Fixed in https://github.com/spack/spack-infrastructure/pull/415/commits/f5ebc95d9270d0f8e5f330654c4840c51eaa3565

k8s/production/custom/prune-buildcache/cron-jobs.yaml
mvandenburgh: Fixed in https://github.com/spack/spack-infrastructure/pull/415/commits/c9aee27233c62764e79b7f738884bed5a64742d4

mvandenburgh:
@kwryankrattiger it looks like the `DeleteAndPutObjectsFromBucketSpackBinaries` role you referred to in your [service account definition](https://github.com/spack/spack-infrastructure/pull/415/commits/53c774ba16cc575ce6e94018812ac3c9f15c30e8#diff-788a48b7875118fce696c4e8cd5656d389af8b5c1b6d8dd6d7af13daff3309ecR8) doesn't exist in AWS. I added it to Terraform, so it will be automatically created and assigned to that service account. 

Also, adding the docker image to the build workflow caused CI to do a test build, and it looks like something is causing it to fail - https://github.com/spack/spack-infrastructure/actions/runs/8421844121/job/23059698970?pr=415

mvandenburgh (dismissed):
The k8s stuff looks good to me :+1: I'll leave it to @scottwittenburg to review pruning script itself separately.

images/ci-prune-buildcache/buildcache.py
```diff
@@ -0,0 +1,183 @@
...
+            processes = stride
+
+        # Only spawn as many processes as needed
+        processes = min(stride, processes)
+
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(self.url.netloc)
+
+        def delete_keys_f(i: int):
+            time.sleep(1)
```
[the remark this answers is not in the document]
danlamanna: Are they timing out or are you getting a [429 error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)?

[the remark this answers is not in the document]
kwryankrattiger: This was the error from boto3

```
Error -- Skipping pruning of e4s-cray-sles
An error occurred (SlowDown) when calling the DeleteObjects operation (reached max retries: 4): Please reduce your request rate.
```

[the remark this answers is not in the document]
kwryankrattiger: Thngs I tried

* Batch size from 1000 -> 100
* Parallel level from 16 -> 1
* sleep(1)

Only the sleep fixed it reliably.

images/ci-prune-buildcache/ci_buildcache_prune.py
```diff
@@ -0,0 +1,497 @@
...
+PRUNE_REF = os.environ.get("PRUNE_REF", "develop")
+SINCE_DAYS = int(os.environ.get("PRUNE_SINCE_DAYS", 14))
+
+# Gitlab API (for fetching artifacts)
+GITLAB_URL = os.environ.get("GITLAB_URL", "https://gitlab.spack.io")
+GITLAB_PROJECT = os.environ.get("GITLAB_PROJECT", "spack/spack")
+gl = gitlab.Gitlab(GITLAB_URL)
+
+# S3
+BUILDCACHE_URL = os.environ.get("BUILDCACHE_URL", f"s3://spack-binaries/{PRUNE_REF}").rstrip("/")
```
danlamanna, line 34: Can this fail if it isn't in the environment? Otherwise I'd rather staging not default to trying to mess with the production bucket.

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
danlamanna, line 1:
```suggestion
#!/bin/bash
set -euo pipefail
```

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+if [ -z $1 ]; then
+  now=$(date --iso-8601)
+else
+  now=$1
+fi
+
+snapshot_dir="$PWD/snapshot-${now}"
+
+# Use the $SPACKROOT/etc/ dir for configs
+SPACK_DISABLE_LOCAL_CONFIG=1
```
danlamanna, line 14: Is this leftover from debugging?

images/ci-prune-buildcache/helper.py
```diff
@@ -0,0 +1,45 @@
+from datetime import datetime
+import json
+import re
+
+
+hash_re = re.compile(r"-([a-zA-Z0-9]{32,32})\.")
```
danlamanna, line 6:
```suggestion
hash_re = re.compile(r"-([a-zA-Z0-9]{32})\.")
```

images/ci-prune-buildcache/ci_buildcache_prune.py
```diff
@@ -0,0 +1,497 @@
...
+
+        key_path = os.path.join(self.cache_dir, key)
+        if os.path.exists(key_path):
+            raise Exception(f"Duplicate cache key: {key_path}")
+
+        # This is happening in parallel, try it but don't fail
+        if not os.path.isdir(os.path.dirname(key_path)):
+            try:
+                os.makedirs(os.path.dirname(key_path))
+            except:
```
danlamanna, line 58:
```suggestion
            except Exception:
```

images/ci-prune-buildcache/buildcache.py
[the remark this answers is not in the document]
danlamanna: I think we're probably better off reconfiguring `max_attempts` https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#defining-a-retry-configuration-in-a-config-object-for-your-boto3-client.

Might even be worth giving [adaptive retry mode](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#adaptive-retry-mode) a shot.

images/ci-prune-buildcache/entrypoint.sh
kwryankrattiger: nope, this is intentional. Unless you are running spack with a specific config root or on a local system this should probably be set. It prevents accidentally injecting persistent spack configs for detecting spack configs used by another system installation of spack.

danlamanna: Oh, then does it need to be exported?

kwryankrattiger: it probably should be to make it clear. But I think it was fine since spack runs in the current shell, it worked for me locally.

images/ci-prune-buildcache/ci_buildcache_prune.py
kwryankrattiger: I left it open since I don't care if/how it fails here. afaict the only ways this would fail is if a) I can't write to the fs which means the entire operation is borked anyway or b) the directory already exists in which case we are fine to continue.

images/ci-prune-buildcache/buildcache.py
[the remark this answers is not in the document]
kwryankrattiger: I will try these today. Does 10 retries seam reasonable? I can't tell from the adaptive description at what point the retry is waiting forever, but there was an example for standard that used 10 so maybe it is fine.

images/ci-prune-buildcache/entrypoint.sh
danlamanna: My lack of bash-fu knowledge might be showing, but aren't only exported variables passed to the environment of child processes? I don't see how spack could ever see this value.

images/ci-prune-buildcache/ci_buildcache_prune.py
kwryankrattiger: This is setup to make the local running of the script less of a hassle. Credentials are required to actually perform the  destructive operation so even if this does run in staging incorrectly configured it shouldn't have credentials to modify production.

images/ci-prune-buildcache/buildcache.py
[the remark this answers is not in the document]
kwryankrattiger: that seems to have fixed my issue locally. Pushed changes.

images/ci-prune-buildcache/Dockerfile
mvandenburgh: Just standardization. If there are benefits to using Fedora that the `python` image doesn't have, I don't have any objections

images/ci-prune-buildcache/ci_buildcache_prune.py
kwryankrattiger: There are cases when there really is no `spack.lock` file. I originally had this throw an error, but in all of the cases where I could not fetch the lock it simply did not exist.

The assumption here is the Gitlab endpoint is reachable.

mvandenburgh (dismissed):

images/ci-prune-buildcache/ci_buildcache_prune.py
scottwittenburg: Ok, I guess that's fine.  But the call site for this method expects to unpack both a stack and list from the return value, so if you end up in here, isn't that going to result in a breaking stop in execution?  I think it will raise `ValueError: too many values to unpack (expected 2)`.

mvandenburgh (approved):

images/ci-prune-buildcache/ci_buildcache_prune.py
kwryankrattiger: ah think you are right! good catch, I will fix that!

PR #417, opened 2023-03-22
Escape parentheses in taxonomy regexes

mvandenburgh:
I noticed that python's `re.compile` method doesn't escape parentheses, so they're being treated as special characters instead of literals currently.


AlmightyYakob:
Was this leading to those errors not being properly picked up?

PR #418, opened 2023-03-22
Add classification for failed jobs with no trace

mvandenburgh:
Another kind of error being labeled as `other` are jobs with no trace, like this one https://gitlab.spack.io/spack/spack/-/jobs/6258279.

### DIFF
--- a/images/upload-gitlab-failure-logs/taxonomy.yaml
+++ b/images/upload-gitlab-failure-logs/taxonomy.yaml
@@ -142,6 +142,9 @@ taxonomy:
       grep_for:
         - 'Error: sha256 checksum failed for .+'
 
+    no_trace: null
+      # See upload_gitlab_failure_logs.py for how this is used
+
   deconflict_order:
     # API Scrape erorrs
     - 'job_log_missing'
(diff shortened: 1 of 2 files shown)


AlmightyYakob (approved):
LGTM, but I'm wondering if it makes sense to classify this as "No trace". I'm not sure what cause leads to no trace being recorded, but the result is that the error is "unknown". Having an "Unknown" category for errors we as of yet can't identify might be useful

mvandenburgh:
> LGTM, but I'm wondering if it makes sense to classify this as "No trace". I'm not sure what cause leads to no trace being recorded, but the result is that the error is "unknown". Having an "Unknown" category for errors we as of yet can't identify might be useful

Isn't "other" already serving that purpose? And yeah I agree that "no trace" isn't useful at face value for determining the true cause of the error, but I think it would be good to eliminate these types of errors from the "other" pool, at least for now

AlmightyYakob:
> Isn't "other" already serving that purpose? And yeah I agree that "no trace" isn't useful at face value for determining the true cause of the error, but I think it would be good to eliminate these types of errors from the "other" pool, at least for now

Other seems to mean "unclassified", as in, the cause could be known but currently isn't. However, in this case, there's no trace available, so there's no way (from just the job logs) for us to know what happened.

I won't die on this hill, I was just pointing out that "no trace" seems to indicate that we know that the cause of the error is, when in fact all we know is that we have no information about the underlying error.

PR #420, opened 2023-03-22
Do not evict runner pods

AlmightyYakob:
This will prevent runner pods from being evicted via consolidation, among other things. [Here](https://karpenter.sh/preview/concepts/deprovisioning/#disabling-deprovisioning) are the relevant docs.

### DIFF
--- a/k8s/runners/protected/graviton/2/release.yaml
+++ b/k8s/runners/protected/graviton/2/release.yaml
@@ -19,7 +19,7 @@ spec:
   chart:
     spec:
       chart: gitlab-runner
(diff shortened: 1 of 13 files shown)


zackgalbreath (approved):

mvandenburgh (approved):

AlmightyYakob:
Update: I watched for a new runner pod to show up, and it did, with the `karpenter.sh/do-not-evict` annotation successfully applied. If that is functioning correctly, we should see the amount of errors associated with "pod not found" drop in the coming days.

PR #424, opened 2023-03-22
[Failed Jobs Dashboard] Order by Doc Count instead of Term Value in Pie Chart

AlmightyYakob:
I was previously ordering by "Term Value", which I thought would be the document count, but now I have no idea what it means, since it was giving the wrong results.


zackgalbreath (approved):

mvandenburgh (approved):
I did the same thing with the bar chart at first :)

AlmightyYakob:
> I did the same thing with the bar chart at first :)

At least I'm not alone :smiling_face_with_tear: 

Do you know what "Term Value" even means?

mvandenburgh:
Nope. I thought it meant alphabetical order, but clearly it doesn't...

PR #426, opened 2023-03-22
Add gitlab `build_failure_reason` to dashboard

mvandenburgh:
Screenshot from 2023-03-22 15-10-49


mvandenburgh:
This was easy to put together, since we already do a similar thing for the `error_taxonomy` field, so I just went ahead and did it. @zackgalbreath @AlmightyYakob do you think this is worth adding?

AlmightyYakob (approved):
I'm okay with it, but I also don't really know what `build_failure_reason` means.

PR #430, opened 2023-03-24
Add "do-not-evict" annotation to GitLab runner manager pods

zackgalbreath:


zackgalbreath:
I made sure this was the right place to add this setting my manually editing & applying the x86_64 testing YAML file. I verified that I saw the annotation show up as expected before flux overwrote my changes.

AlmightyYakob (commented):
Should we also include the test runners?

zackgalbreath:
> Should we also include the test runners?

I dunno, what do you think? I left them out on purpose so they don't cost us more money than necessary, but maybe it's worth the surcharge for a more faithful reproducer of our production runners.

AlmightyYakob:
> > Should we also include the test runners?
> 
> I dunno, what do you think? I left them out on purpose so they don't cost us more money than necessary, but maybe it's worth the surcharge for a more faithful reproducer of our production runners.

I doubt cost will really be an issue, afaik they're barely ever used anyway. I was mainly concerned with reproducibility.

AlmightyYakob (approved):

PR #431, opened 2023-03-27
Use emptiness deprovisioning method for runner nodes

AlmightyYakob:
Consolidation doesn't really make sense for runner nodes, since all runner pods are marked as `do-not-evict` and as such, consolidation will at best just not work, and at worst cause higher costs / scheduling issues.


zackgalbreath (approved):
LGTM. One question. This PR leaves the `do-not-evict` annotations in place on the GitLab CI pods. Is this still a good idea, or in light of these changes is it now unnecessary cruft that we should remove?

PR #432, opened 2023-03-27
Add deprovisioning for beefy and gitlab provisioners

AlmightyYakob:
I noticed that our `beefy` and `gitlab` node pools are currently not configured with any deprovisioning, meaning nodes that are provisioned for these node pools will stay there indefinitely, whether they're being used or not. This may not be a huge problem, but imo every provisioner should be configured with a deprovisioner of some sort, as otherwise large bursts of activity can leave empty nodes that rack up costs.

I defaulted to consolidation with these, but I'm not really sure if that fits the use case for these node pools. @zackgalbreath what do you think?

### DIFF
--- a/k8s/karpenter/provisioners/beefy/provisioner.yaml
+++ b/k8s/karpenter/provisioners/beefy/provisioner.yaml
@@ -8,6 +8,10 @@ spec:
   providerRef:
     name: default
 
+  # Consolidation will de-provision larger than necessary nodes
+  consolidation:
+    enabled: true
+
   requirements:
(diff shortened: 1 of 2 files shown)


zackgalbreath:
Not 100% sure, but consolidation seems safe to me in these cases. The two services that use these provisioners are gitlab and prometheus.

mvandenburgh (approved):
Makes sense to me. In practice I don't think the gitlab and Prometheus nodes will ever be consolidated, but i agree it's good to have

PR #434, opened 2023-03-28
Add custom domain for opensearch

mvandenburgh:
Adds Terraform code for provisioning a custom domain for opensearch (`opensearch.spack.io`). I already applied it to AWS, and everything appears to be working.

As part of this PR, I took a quick first pass at encoding OpenSearch into our terraform code. For now I have just encoded the OpenSearch cluster itself, and not any of the related resources like Cognito or Cloudwatch. I think in the future we'll want to have a separate Terraform module for configuring opensearch since there's so many moving parts, but until someone has the bandwidth to do that, I think this is sufficient.

### DIFF
--- a/terraform/modules/spack/opensearch.tf
+++ b/terraform/modules/spack/opensearch.tf
@@ -31,6 +31,9 @@ resource "aws_opensearch_domain" "spack" {
   }
 
   domain_endpoint_options {
+    custom_endpoint_enabled = true
+    custom_endpoint = "opensearch.spack.io"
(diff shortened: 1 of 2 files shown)


mvandenburgh:
> but I guess here open search is not in k8s so we have to do it differently? Asking mostly for my personal understanding.

Yup exactly. OpenSearch is completely separate from k8s so it has to be done this way instead of relying on the k8s cert-manager.

AlmightyYakob (approved):
LGTM but I'm not very familiar with the opensearch configuration

PR #435, opened 2023-03-28
Add data links that point to opensearch

AlmightyYakob:
This will make it easier to view exactly what errors are occuring per category, from the grafana dashboard.


scottwittenburg (approved):
Seems like this is working since you showed it to me live.  This will make it a lot easier to find instances of each type of error.  Thanks!

PR #438, opened 2023-03-29
Wait an hour before considering a pipeline as skipped

zackgalbreath:
Update our skipped pipelines mitigation to wait an hour before using the API to trigger a pipeline for a branch that otherwise does not have one for its HEAD commit.

This resolves a problem we noticed recently where GitLab eventually reacted to the original push event. That resulted in two pipelines for a given commit.


scottwittenburg (approved):
It's unfortunate we have to do these gymnastics, but I don't see a way around it given the behavior we have observed from gitlab.

The logic looks good to me, thanks @zackgalbreath!

PR #440, opened 2023-03-29
Prefer classifying spack errors over system ones

scottwittenburg:
In cases like [this](https://gitlab.spack.io/spack/spack/-/jobs/6355081) one, where there is a spack error (in this case a timeout acquiring a write lock), followed by an infrastructure failure, we prefer to classify the error with the spack error.  This PR moves all the spack-related errors to a higher precedence than the infrastructure-related errors.


scottwittenburg:
@kwryankrattiger can you rebase on this when it's merged?

zackgalbreath (approved):

alalazo:
Not sure how this taxonomy is implemented, but can't we attach in some way all the matching tags to a record?

kwryankrattiger:
> Not sure how this taxonomy is implemented, but can't we attach in some way all the matching tags to a record?

It is implemented here, it uses a precedence for the errors on only classify one. The way it is uploaded to opensearch, to my understanding, is a single scalar field.
https://github.com/spack/spack-infrastructure/blob/main/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py#L77

tgamblin:
> Not sure how this taxonomy is implemented, but can't we attach in some way all the matching tags to a record?

I think for purposes of triage, we want one category per error.  I'm ok with preferring Spack errors here -- I think the infrastructure errors are going to be more prevalent when they're serious, and choosing how best to break a tie isn't going to affect the triage decision much.

alalazo:
> I think for purposes of triage, we want one category per error. 

@tgamblin Sure, I would keep triaging the same, but have a separate field with all the matches - see the comments in https://github.com/spack/spack-infrastructure/pull/442 In the worst case it's a field we won't look at, but I think it might be useful to have all matches. 

Say we have a lot of Spack fetch errors. One thing I would try, to prioritize the ones to look at first, is select the ones that are also not matching a network error, etc.

PR #442, opened 2023-03-30
Refactor the taxonomy script for testing classification

kwryankrattiger:
I re-worked this a bit so I can use this script as an imported module for testing.

```python
import upload_gitlab_failure_logs as taxonomy

with read("my-test-trace.txt", "r") as trace:
    print(taxonomy.classify(trace))
```

### DIFF
--- a/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
+++ b/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
@@ -4,36 +4,64 @@ import os
 import re
 from pathlib import Path
 
-import gitlab
 import yaml
-from opensearch_dsl import Date, Document, connections
+
+GITLAB_TOKEN = os.environ.get("GITLAB_TOKEN")
+OPENSEARCH_ENDPOINT = os.environ.get("OPENSEARCH_ENDPOINT")
+OPENSEARCH_USERNAME = os.environ.get("OPENSEARCH_USERNAME")
+OPENSEARCH_PASSWORD = os.environ.get("OPENSEARCH_PASSWORD")
 
 
-class JobPayload(Document):
-    timestamp = Date()
-
-    class Index:
-        name = "gitlab-job-failures-*"
-
-    def save(self, **kwargs):
-        # assign now if no timestamp given
-        if not self.timestamp:
-            self.timestamp = datetime.now()
-
-        # override the index to go to the proper timeslot
-        kwargs["index"] = self.timestamp.strftime("gitlab-job-failures-%Y%m%d")
-        return super().save(**kwargs)
+def load_taxonomy():
+    with open(Path(__file__).parent / "taxonomy.yaml") as f:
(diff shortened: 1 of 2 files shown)


mvandenburgh (commented):
These changes are good with me, but they will conflict with @AlmightyYakob's PR #433.

kwryankrattiger:
np, it is a pretty easy thing to fix-up. I also added the script I used to do the testing.

Usage is something like

```bash
python validate_taxonomy.py $(ls trace*)
```

images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
```diff
@@ -4,36 +4,64 @@
...
+                job_error_class = error_class
+                break
+    else:
+        job_error_class = "other"
 
-GITLAB_TOKEN = os.environ["GITLAB_TOKEN"]
-OPENSEARCH_ENDPOINT = os.environ["OPENSEARCH_ENDPOINT"]
-OPENSEARCH_USERNAME = os.environ["OPENSEARCH_USERNAME"]
-OPENSEARCH_PASSWORD = os.environ["OPENSEARCH_PASSWORD"]
+    return job_error_class
```
alalazo, line 44: Following https://github.com/spack/spack-infrastructure/pull/440#issuecomment-1490634469, why can't we:
```suggestion
    return matching_patterns, job_error_class
```
?

images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
```diff
@@ -63,25 +91,9 @@ def main():
...
-    # with the highest priority in the "deconflict order".
-    # Otherwise, assign it a taxonomy of "other".
-    if len(matching_patterns):
-        for error_class in taxonomy["deconflict_order"]:
-            if error_class in matching_patterns:
-                job_error_class = error_class
-                break
-    else:
-        job_error_class = "other"
+    job_error_class = classify(job_trace)
```
alalazo, line 94: And here do something like:
```suggestion
    matches, job_error_class = classify(job_trace)
    job_input_data["matches"] = sorted(matches)
```
?

kwryankrattiger: I think one good reason is not filtering to a single match can lead to mislabeled classifications. The ordering below is meant to identify the "best' error match.

For example, `file_not_found` is only an error if there are no other errors since it is possible for uploading artifacts to not find a file, but not be a relevant error.

mvandenburgh (approved):
Sorry, thought I approved this already. This will need to be rebased to resolve conflicts but the changes LGTM

mvandenburgh:
We've since refactored the taxonomy classifier script into the webhook handler, so this is no longer relevant. You can achieve this functionality currently by importing the function like so:
```python
from analytics.job_processor.dimensions import _assign_error_taxonomy
```

PR #445, opened 2023-03-30
Add runner pod affinity

AlmightyYakob:
This PR is aimed at placing pods on nodes that already have pods running ("pod packing"), in order to prevent low node utilization.

### DIFF
--- a/k8s/runners/protected/graviton/2/release.yaml
+++ b/k8s/runners/protected/graviton/2/release.yaml
@@ -61,6 +61,20 @@ spec:
             namespace = "pipeline"
             poll_timeout = 600  # ten minutes
             service_account = "runner"
+
+            # Place pod close to other pipeline pods if possible ("pack" the pods tightly)
+            # Docs: https://docs.gitlab.com/runner/executors/kubernetes.html#define-nodes-where-pods-are-scheduled
+            [runners.kubernetes.affinity]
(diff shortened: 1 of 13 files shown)


AlmightyYakob:
@alecbcs

zackgalbreath (approved):
the documentation referenced in this change
https://docs.gitlab.com/runner/executors/kubernetes.html#define-nodes-where-pods-are-scheduled

refers to `topology_key = "failure-domain.beta.kubernetes.io/zone"`

but @AlmightyYakob correctly used `topology.kubernetes.io/zone` instead as this is the replacement for the deprecated `failure-domain.beta.kubernetes.io/zone`

https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesioregion

AlmightyYakob:
Waiting to merge this until we fix our error reporting, so that if this results in an increase in errors we can see it.

alecbcs:
> Waiting to merge this until we fix our error reporting, so that if this results in an increase in errors we can see it.

Is there an existing PR for the error reporting improvements we can link to as a dependency for this PR?

AlmightyYakob:
> Is there an existing PR for the error reporting improvements we can link to as a dependency for this PR?

No, this is a recent problem we've encountered and doesn't have an associated issue/PR.

PR #446, opened 2023-03-30
Fix runner node selection

AlmightyYakob:
Currently we're being slightly inefficient with our pod scheduling, since we don't allow any sharing between "node pools". For example, if an `x86-64-v2` job is kicked off, it's able to run on any node provisioned by the `v2`, `v3` or `v4` provisioners, since it's compatible with all of those. However right now, it'll only be placed on nodes that are provisioned by the `v2` provisioner, which may lead to some wasted utilization if it could have fit on a `v3` or `v4` provisioned node.

### DIFF
--- a/k8s/production/runners/protected/x86_64/v3/release.yaml
+++ b/k8s/production/runners/protected/x86_64/v3/release.yaml
@@ -63,9 +63,21 @@ spec:
             poll_timeout = 600  # ten minutes
             service_account = "runner"
 
-            # Place pod close to other pipeline pods if possible ("pack" the pods tightly)
-            # Docs: https://docs.gitlab.com/runner/executors/kubernetes.html#define-nodes-where-pods-are-scheduled
             [runners.kubernetes.affinity]
+              # Schedule this pod on any node with x86_64 >= v3
+              [runners.kubernetes.affinity.node_affinity]
+              [runners.kubernetes.affinity.node_affinity.required_during_scheduling_ignored_during_execution]
+                [[runners.kubernetes.affinity.node_affinity.required_during_scheduling_ignored_during_execution.node_selector_terms]]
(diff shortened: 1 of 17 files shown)


AlmightyYakob:
I believe this is ready to go. I've tested this in staging by creating pods for each type of microarch, that just sleep forever, ensuring they stay on the node they're scheduled on. Then, I played around with spinning up various types of nodes. Indeed, it seems that pods won't schedule on a lower version microarch (e.g. requires v4 when only a v2 node is available), and that when multiple nodes _are_ available, it will weight towards its minimum requirement (i.e. if there are v2, v3, and v4 nodes available, a v2 pod will schedule on the v2 node if it has capacity).

Furthermore, I also observed that in a situation where there is only a v4 node active, and a v2 pod is created, it will schedule on the v4 node, instead of spinning up an entirely new v2 node.

k8s/production/runners/protected/x86_64/v3/release.yaml
```diff
@@ -63,12 +63,39 @@ spec:
             poll_timeout = 600  # ten minutes
             service_account = "runner"
 
-            # Place pod close to other pipeline pods if possible ("pack" the pods tightly)
-            # Docs: https://docs.gitlab.com/runner/executors/kubernetes.html#define-nodes-where-pods-are-scheduled
             [runners.kubernetes.affinity]
+              [runners.kubernetes.affinity.node_affinity]
+
+              # Schedule this pod on any node with x86_64 >= v3
```
mvandenburgh, line 69: Why does the signing runner need x86_64 >= v3? Is it incompatible with v2?

AlmightyYakob: It was scheduled in the x86-64-v3 pool prior to this, so I was just ensuring that this change doesn't break compatibility.

mvandenburgh (approved):
:+1:
It's unfortunate that we don't have a good way to fully test these sorts of changes in staging yet, but as long as your tests were successful, this all makes sense to me.

PR #452, opened 2023-04-05
Add pie chart for highest build time by package

AlmightyYakob:
image (2)

I'll no doubt be making more changes to this, but just want to encode what I have to start.

### DIFF
--- a/k8s/prometheus/custom/pipeline-user-impact-dashboard.yaml
+++ b/k8s/prometheus/custom/pipeline-user-impact-dashboard.yaml
@@ -36,10 +36,93 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
-      "id": 31,
       "links": [],
       "liveNow": false,
       "panels": [
+        {
+          "datasource": {
+            "type": "postgres",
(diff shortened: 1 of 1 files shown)


tgamblin:
This looks great!  Stuff we'll want to know:
1. Total build time (the graph probably needs an "other" wedge, or it needs to say what percent of total build time the top packages are)
2. total number of jobs (we already have this elsewhere)
3. how many jobs comprise each wedge
4. min/max/average/median build time in each wedge
5. what percent of each wedge was retries vs. non-retried builds

AlmightyYakob:
> This looks great! Stuff we'll also want to know:
> 
> 1. Total build time (the graph probably needs an "other" wedge, or it needs to say what percent of total build time the top packages are)
> 2. total number of jobs (we already have this elsewhere)
> 3. how many jobs comprise each wedge
> 4. min/max/average/median build time in each wedge
> 5. what percent of each wedge was retries vs. non-retried builds

Since a lot of this would depend on other changes to our data analysis infrastructure (e.g. setting up a replica gitlab db), I think we should merge this as is, and I'll create a new issue for your comment.

mvandenburgh (approved):

PR #453, opened 2023-04-07
Add classification for missing sha256sum command

AlmightyYakob:
Recently we've been seeing the following errors:

```
sha256sum: command not found
```

which are being incorrectly classified as `file_not_found`, since this errors seems to directly cause that to occur afterwards.

This will require running reclassifying the last few days of errors if we'd like them to be accurate.


AlmightyYakob:
@scottwittenburg to your knowledge have I structured this correctly? Specifically, I'd like the `sha256sum_not_found` classification to take precedence over `file_not_found`.

scottwittenburg (approved):
Yes, it looks like this will do what you want, thanks!

PR #456, opened 2023-04-10
Provide role arn in provider eks get-token args

AlmightyYakob:
The `kubectl`, `kubernetes` and `helm` providers would all raise `Unauthorized` when trying to run terraform apply/plan, even though `kubectl` functioned from the CLI without issue. This is because the role arn is provided in the local kube config file, but it was missing here.

### DIFF
--- a/terraform/production/.terraform.lock.hcl
+++ b/terraform/production/.terraform.lock.hcl
@@ -142,6 +142,26 @@ provider "registry.terraform.io/hashicorp/random" {
   ]
 }
 
+provider "registry.terraform.io/hashicorp/time" {
+  version     = "0.9.1"
+  constraints = ">= 0.7.2"
+  hashes = [
+    "h1:NUv/YtEytDQncBQ2mTxnUZEy/rmDlPYmE9h2iokR0vk=",
+    "zh:00a1476ecf18c735cc08e27bfa835c33f8ac8fa6fa746b01cd3bcbad8ca84f7f",
+    "zh:3007f8fc4a4f8614c43e8ef1d4b0c773a5de1dcac50e701d8abc9fdc8fcb6bf5",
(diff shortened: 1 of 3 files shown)


terraform/production/.terraform.lock.hcl
```diff
@@ -142,6 +142,26 @@ provider "registry.terraform.io/hashicorp/random" {
...
+    "zh:a6d9206e95d5681229429b406bc7a9ba4b2d9b67470bda7df88fa161508ace57",
+    "zh:aa299ec058f23ebe68976c7581017de50da6204883950de228ed9246f309e7f1",
+    "zh:b129f00f45fba1991db0aa954a6ba48d90f64a738629119bfb8e9a844b66e80b",
+    "zh:ef6cecf5f50cda971c1b215847938ced4cb4a30a18095509c068643b14030b00",
+    "zh:f1f46a4f6c65886d2dd27b66d92632232adc64f92145bf8403fe64d5ffa5caea",
+    "zh:f79d6155cda7d559c60d74883a24879a01c4d5f6fd7e8d1e3250f3cd215fb904",
+    "zh:fd59fa73074805c3575f08cd627eef7acda14ab6dac2c135a66e7a38d262201c",
+  ]
+}
+
```
mvandenburgh, line 164: Where did this package come from? I don't see any other dependency changes that would trigger this lockfile update.

AlmightyYakob: I'm not sure but I'm guessing it's utilized by a provider that we use, and was able to be upgraded. Other examples of providers we don't directly use but are included in the lock file are `random` and `cloudinit`.

AlmightyYakob: I should also mention the reason this seemingly unrelated change is being included here: terraform yells at me otherwise.

mvandenburgh (approved):

PR #457, opened 2023-04-10
CI Runners for parallel cluster

kwryankrattiger:
Adding some runners managers that dispatch to  AWS pcluster AMIs using the AWS autoscaler.

After reading the docs a few more times, I think it is sufficient to use the GitLab runner docker containers we have with the `docker+machine` executor. My understanding is this executor runs the runner manger in docker, and then then pipeline is launched on the "machine" using the AMI directly. This should mean that we _don't_ have to build a custom AMI with Gitlab components installed. This PR should be enough to start testing.

https://docs.gitlab.com/runner/configuration/runner_autoscale_aws/


AlmightyYakob:
Also, I've got no idea how auth is supposed to work in this case. Since we're not running a job _on_ the cluster _from_ the cluster, and are instead directly attempting to spin up nodes and run jobs on EC2 images, we'll need to set up some IAM roles that allow this to be done. Karpenter is setup with roles to allow it to spin up nodes, but we're not using that here, so we'll need something similar.

k8s/runners/protected/pcluster/release.yml
```diff
@@ -0,0 +1,89 @@
...
+            MaxBuilds = 100
+            MachineDriver = "amazonec2"
+            MachineName = "gitlab-ci-pcluster-protected-%s"
+            MachineOptions = [
+              "amazonec2-ami=ami-011244318c573c915", # aws-parallelcluster-3.5.1-amzn2-hvm-x86_64-202303171148 2023-03-17T11-52-52.966Z
+              "amazonec2-region=us-east-1",
+              "amazonec2-use-private-address=true",
+              "amazonec2-tags=runner-manager-name,gitlab-aws-autoscaler,gitlab,true,gitlab-runner-autoscale,true",
+              "amazonec2-instance-type=t3.medium",
+              "amazonec2-iam-instance-profile=BEEF",
```
kwryankrattiger, line 63: Need to figure out what this needs to be

kwryankrattiger: There is also a field for key-pairs. I think both of these can be expressed as environment variables, is it possible to  populate those via secrets here?

@AlmightyYakob

kwryankrattiger:
Setup the runners and provisioner for staging in a fixup commit, this commit should be reverted before merging.

terraform/modules/spack/karpenter.tf
```diff
@@ -104,3 +104,39 @@ resource "kubectl_manifest" "karpenter_node_template" {
...
+        # We use the "deployment_name" variable here instead of the full cluster name
+        # because the full cluster name isn't available at the time that we bootstrap
+        # the VPC resources (including subnets). However, "deployment_name" is also
+        # a unique-per-cluster value, so it should work just as well.
+        karpenter.sh/discovery: ${var.deployment_name}
+      securityGroupSelector:
+        karpenter.sh/discovery: ${module.eks.cluster_name}
+      tags:
+        karpenter.sh/discovery: ${module.eks.cluster_name}
+        spack.io/pcluster: true
```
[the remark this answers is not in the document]
kwryankrattiger: ah nice, trying it now

k8s/production/karpenter/provisioners/runners/pcluster/graviton/3/provisioners.yaml
```diff
@@ -0,0 +1,71 @@
...
+        - "10xlarge"
+        - "12xlarge"
+        - "16xlarge"
+
+    # Availability Zones
+    - key: "topology.kubernetes.io/zone"
+      operator: In
+      values:
+      - "us-west-1b"
+      - "us-west-1c"
```
mvandenburgh, line 56:
```suggestion
      values:
      - "us-west-2a"
      - "us-west-2b"
      - "us-west-2c"
      - "us-west-2d"
```

k8s/production/runners/protected/graviton/3/release.yaml
```diff
@@ -48,7 +48,7 @@ spec:
           [runners.kubernetes]
             helper_image = "gitlab/gitlab-runner-helper:arm-latest"
             privileged = false
-            helper_memory_request = "512m"
+            helper_memory_request = 512
```
mvandenburgh, line 51: Why this change? Does the "512m" string not work?

mvandenburgh (commented):
I left one last question, but besides that I think this is good to go.

terraform/modules/spack/karpenter.tf
```diff
@@ -104,3 +104,46 @@ resource "kubectl_manifest" "karpenter_node_template" {
     helm_release.karpenter
   ]
 }
+
+locals {
+  pcluster_ami_id = {
+    "prod": "ami-00b2c701206072ffc"
+    "staging": "ami-0e2c34d361e37afa2"
```
mvandenburgh, line 111: Note, these are custom "EKS-optimized" AMIs based on the AWS parallel cluster AMI. The packer code I used to build them is on this branch - https://github.com/mvandenburgh/amazon-eks-ami/tree/pcluster-ami

mvandenburgh: If these pcluster runners are going to be a longer term thing, we may want to run this packer build in CI. The AMI images are dependent on kubernetes version, so left un-automated this is another thing we'll have to manually update when we get to the kubernetes upgrade.

kwryankrattiger:
I am okay with this as is, but I can't merge it.

k8s/production/runners/protected/graviton/3/release.yaml
mvandenburgh: The rest of the runners in `k8s/production/runners/*` have this as a string, so I think before merging we should either back out this change, or update the rest of the runner memory requests to also be an int. Maybe the former is better, as the gitlab docs usually have these types of fields set as strings? (see `helper_cpu_limit` in this example - https://gitlab.com/gitlab-org/gitlab-runner/blob/ffdb32cb9cd1e325b5018cea2ddd6084b6ec741e/docs/executors/kubernetes.md#define-keywords-in-the-config-toml)

AlmightyYakob (approved):

k8s/production/runners/protected/pcluster/graviton/3/release.yaml
```diff
@@ -0,0 +1,139 @@
...
+              name = "spack-intermediate-ci-signing-key"
+              mount_path = "/mnt/key/"
+              read_only = true
+            [[runners.kubernetes.volumes.host_path]]
+              name = "software"
+              mount_path = "/opt"
+              read_only = true
+            [[runners.kubernetes.volumes.host_path]]
+              name = "keys"
+              mount_path = "/mnt/key"
```
scottwittenburg, line 111: I wonder if this is overwriting the `spack-intermediate-ci-signing-key` stuff mounted in the same location above?  See this job, where there don't seem to be any keys there:

https://gitlab.spack.io/spack/spack/-/jobs/7122837#L404

kwryankrattiger: Delete this, use the mounted secret > the host path. Host path is the AMI location, which doesn't have secrets.

PR #459, opened 2023-04-11
Create one NAT gateway per AZ

mvandenburgh:
We discovered recently that one thing causing high AWS costs is the fact that we don't have a NAT gateway in every availability zone. The terraform vpc module we use exposes a parameter that allows us to configure that, so I've enabled that here.


mvandenburgh:
I haven't applied this in production yet, but I did apply it in staging; it took about a minute and it didn't appear to cause any downtime for gitlab.

mvandenburgh:
Superseded by #478

PR #462, opened 2023-04-12
Manage secrets with sealed secrets

AlmightyYakob:
Currently, secrets are managed in an ad-hoc way, where the secret is created in the cluster, and then added here in a completely commented file named `secrets-dummy`, purely for reference. [Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) provide a way for us to securely commit our secrets into our codebase, without the need for any workarounds.


AlmightyYakob:
Waiting to merge this until https://github.com/spack/spack-infrastructure/pull/404 is merged. I'll then rebase this off of that and deal with conflicts, as well as configure staging, since staging isn't currently covered by this PR.

k8s/production/cdash/sealed-secrets.yaml
```diff
@@ -0,0 +1,20 @@
...
+    root-password: AgDTy47AEtzteGZqRl+OivJUJ/BFLHxZYLVaJ6H/LwPhEyGWhJVZz1VhEnSE3IFkK9avwE3VUvgteyzfBfe8/8X0sHIv3M6ZAQRqr9WyYqnQla64Olco5gvJptJRSgkt3CE+WRWxq/qMoyzuxacYaJsUyOIsq6V8dCMhGTzQEheJQUGfWhXIzxpkOMqcn72qmpAo1ZXWdAYrZfwOymvk0kOOXt/1DAdbPXg4euENgCJIuJahvUUkgOGzhJqYKyaBzKWbj6LBfL8Cw2hdw+ByljwFUegujoUZeeRjArgV6t4jd05z7O9nm5KMhv2cPhW1Dk6AcKkcmETcByYOWf/k6N42hGvxSybYCrsfEHQSTbphkWlEMjNk3Dw1/RJw1h2zK3tPsEMKJNudXvncZ374wa7CLLSqB0mWf+2dn1oM5Kc7cIUzUev4Ea1OfBxSJbdLtTDAkgMjMGL6O5Fyov2cGQ2Y0y8SOluhIkOvcMAjIIlEG59UX+KzhFD/+ob2J0LPxfeY+tgHe3se277r1qjCNAPOK4yMacgfR32jjrtOw2i55iEGCGQzbAmhdVHq+mhM6LFpv0qKaFLEiFEXn0vuiun0p0XMdfSAWrmzJCDlbwNO1Y9eKqhOXfWilvqiOBLmRpOr3sKsOYdWa+CFhnzI5OziEnB6Bje6nNmoJKuhwtSrpuZhsOzI2SF4yD32xjFybfwurdENNOF0fNVDSYXjbPmS
+  template:
+    metadata:
+      annotations:
+        kustomize.toolkit.fluxcd.io/reconcile: disabled
```
mvandenburgh, line 19: Why disable flux? Don't we want changes to secrets to be synchronized with the github repo?

AlmightyYakob: A couple of things:

1. This is under `spec.template`, which means it'll be applied to the created/updated `Secret` resource, not the `SealedSecret` resource. So long as the `SealedSecret` resource is tracked, the secrets will get updated
2. Given the point above, I still wasn't sure if we should keep these annotations on the (non-sealed) secrets. All of the secrets currently have this annotation, so was just trying to keep things consistent. I don't know if this annotation even has any effect if the resource isn't checked into source, maybe you can provide some insight there.

mvandenburgh (approved):
Looks good to me, although i haven't tested it.

mvandenburgh:
Should we disable flux in production, merge this, and re-enable it when we're sure it works ok in staging? I'm not sure the best way to test all of this tbh, unless you've already done so.

AlmightyYakob:
> Should we disable flux in production, merge this, and re-enable it when we're sure it works ok in staging? I'm not sure the best way to test all of this tbh, unless you've already done so.

Just to clarify, I have already tested this to the best of my ability by doing the following:

1. Applied one or more of these files (under a different name) to the cluster, checking the contents of the unsealed secret and verifying that it's correct
2. Verified that the private and public key pair match by temporarily storing the private key locally and using the kubeseal [recovery mode](https://github.com/bitnami-labs/sealed-secrets#can-i-decrypt-my-secrets-offline-with-a-backup-key) to decrypt the secret files locally, verifying that their contents are what they should be
3. Applied the `"sealedsecrets.bitnami.com/managed" = "true"` annotation to all existing secrets that have corresponding `sealed-secrets.yaml` files in this PR
4. Verified that an existing secret which has the above annotation can be correctly "taken control of" by the sealed secrets manager (i.e. that the sealed secret would be unsealed and the regular secret data updated without issue, even though all sealed data in this PR is the same as the existing secrets).


I'm still happy to do a test in staging first, but I don't think it'll really be testing what we care about, since the staging cluster uses a different key pair, and doesn't contain most of the secrets we're encoding here. Essentially what I'm saying is, I think the most meaningful test is verifying that we've sealed all of the existing secret data correctly, which I've done to the best of my knowledge.

mvandenburgh:
> I'm still happy to do a test in staging first, but I don't think it'll really be testing what we care about, since the staging cluster uses a different key pair, and doesn't contain most of the secrets we're encoding here. Essentially what I'm saying is, I think the most meaningful test is verifying that we've sealed all of the existing secret data correctly, which I've done to the best of my knowledge.

Agreed, sounds like you've tested this reasonably so I think it's good to merge :+1:

mvandenburgh (approved):

PR #464, opened 2023-04-13
Add a "service" tag to x86_64_v2 runners

scottwittenburg:
Having this tag on these runners, which should include some of the cheaper and more available instances we provision, will allow us to target these instances for the likes of the noop, cleanup, and reindex jobs.

### DIFF
--- a/k8s/runners/public/graviton/2/release.yaml
+++ b/k8s/runners/public/graviton/2/release.yaml
@@ -103,7 +103,7 @@ spec:
       imagePullPolicy: "if-not-present"
       locked: false
 
-      tags: "arm,aarch64,graviton,graviton2,small,medium,large,huge,public,aws,spack"
+      tags: "arm,aarch64,graviton,graviton2,small,medium,large,huge,public,aws,spack,service"
       runUntagged: false
       secret: gitlab-gitlab-runner-secret # from gitlab release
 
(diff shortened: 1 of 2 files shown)


scottwittenburg:
What do you think about this @AlmightyYakob?  The goal is to apply that `service` tag to some of the non-rebuild jobs that don't require much (if anything) in the way of resources.  This will have two benefits: 

1. It turns out we were relying on applying the `public` tag to those kinds of jobs to avoid them accidentally getting run on, eg,  the special signing runner that only allow the use of the `notary` image.  After the spack ci boiler plate reduction PR, it turns out we can no longer get the `public` applied to those jobs, so they are currently failing sometimes due to this.  Hacking in something to apply the `public` tag doesn't seem like a good solution, as choosing a regular vs signing runner seems orthogonal to public vs private.

2. As stated in the description, if we can restrict certain jobs so they don't run on the more optimized architectures, it can potentially save some money.

scottwittenburg:
fyi @kwryankrattiger

AlmightyYakob:
I don't see much issue with this, but if the goal is to target cheaper / more available instances, shouldn't we also tag the public graviton2 runners?

scottwittenburg:
> if the goal is to target cheaper / more available instances, shouldn't we also tag the public graviton2 runners?

Ok, that makes sense, thanks!

AlmightyYakob (approved):
lgtm

PR #467, opened 2023-04-14
Update opensearch TF to fix staging deployment

mvandenburgh:
The introduction of opensearch into terraform in #434 has broken our staging terraform setup, since we don't have an opensearch cluster provisioned for that. This went unnoticed because no one was actually using the staging cluster until now. This PR does a couple things to address this:

1) Makes deployment-specific aspects of opensearch (such as URL, cluster name, and cognito auth config) reliant on the deployment environment
2) Introduces a new input variable to our `spack` module, `provision_opensearch_cluster`, which allows disabling opensearch altogether.

Running this in production is currently a noop (aside from some tf state changes due to some resources being moved), so this shouldn't have any noticeable impact there.

### DIFF
--- a/terraform/modules/spack/opensearch.tf
+++ b/terraform/modules/spack/opensearch.tf
@@ -4,6 +4,8 @@ locals {
 }
 
 resource "aws_opensearch_domain" "spack" {
+  count = var.provision_opensearch_cluster ? 1 : 0
+
   domain_name = "spack${var.deployment_name == "production" ? "" : "-${var.deployment_name}"}"
 
   engine_version = "OpenSearch_1.3"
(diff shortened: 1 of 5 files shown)


terraform/modules/spack/opensearch.tf
```diff
@@ -23,17 +30,25 @@ resource "aws_opensearch_domain" "spack" {
...
+  # We want the existing Cognito set up for our production OpenSearch cluster to
+  # remain in use, so we use a dynamic block to ensure it's ignored for non-production
+  # OpenSearch deployments.
+  dynamic "cognito_options" {
+    for_each = local.cognito_enabled ? [1] : []
+
+    content {
+      enabled          = true
+      identity_pool_id = "us-east-1:ff2664d7-a403-42ba-8407-5d90b3eaa948" # TODO: encode this into terraform
+      role_arn         = aws_iam_role.opensearch_cognito_role[0].arn
```
AlmightyYakob, line 43: Any reason this went from `aws_iam_role.opensearch_cognito_role.arn` to `aws_iam_role.opensearch_cognito_role[0].arn` in this change?

AlmightyYakob (commented):
While I don't fully understand what's going on here, it roughly makes sense to me. Had one question

terraform/modules/spack/opensearch.tf
mvandenburgh: The `aws_iam_role.opensearch_cognito_role` went from a string to an array due to the `count` argument (which we're using to create the resource conditionally)

https://developer.hashicorp.com/terraform/language/meta-arguments/count

PR #471, opened 2023-04-18
[Pipeline Overview Dashboard] Add table showing cpu and pods per node

AlmightyYakob:
image

Here's how it would look. This removes the "CPU Total" and "RAM Total", as there's not really room otherwise. IMO those two don't really provide much valuable info, but if other disagree, I can find some other place to put the table.

The table defaults to sorting by CPU ascending, so you can immediately see idling nodes and how many pipeline pods are on them.


mvandenburgh (approved):
Nice!

PR #427, opened 2023-03-22
Add info about k8s pod to gitlab job failure record

mvandenburgh:
Adds pod "status" metadata for the k8s pod that the failed job is running on. This will allows us to identify jobs that failed due to their pod and/or one of their containers being OOMKilled, etc. 

Here's an example "status" object from a runner pod:

```
{
    "conditions": [
        {
            "last_probe_time": null,
            "last_transition_time": "2023-04-13 18:57:07+00:00",
            "message": null,
            "reason": null,
            "status": "True",
            "type": "Initialized"
        },
        {
            "last_probe_time": null,
            "last_transition_time": "2023-04-13 18:57:08+00:00",
            "message": null,
            "reason": null,
            "status": "True",
            "type": "Ready"
        },
        {
            "last_probe_time": null,
            "last_transition_time": "2023-04-13 18:57:08+00:00",
            "message": null,
            "reason": null,
            "status": "True",
            "type": "ContainersReady"
        },
        {
            "last_probe_time": null,
            "last_transition_time": "2023-04-13 18:56:55+00:00",
            "message": null,
            "reason": null,
            "status": "True",
            "type": "PodScheduled"
        }
    ],
    "container_statuses": [
        {
            "container_id": "containerd://136fc055dfaac36951afae6742476f155e3e7e8d739fa43f5f35de70c402a735",
            "image": "docker.io/ecpe4s/ubuntu20.04-runner-x86_64:2023-01-01",
            "image_id": "docker.io/ecpe4s/ubuntu20.04-runner-x86_64@sha256:026df548cce402c32de68c0a3eee645808c99dd0541e74af4ea3566026da099e",
            "last_state": {
                "running": null,
                "terminated": null,
                "waiting": null
            },
            "name": "build",
            "ready": true,
            "restart_count": 0,
            "started": true,
            "state": {
                "running": {
                    "started_at": "2023-04-13 18:57:07+00:00"
                },
                "terminated": null,
                "waiting": null
            }
        },
        {
            "container_id": "containerd://b9a83080b5b484419619bf9efca1a2d3d4512f7acf1016b6cff88d9795326500",
            "image": "registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper:x86_64-98daeee0",
            "image_id": "registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper@sha256:8706270c187b607189747b6700f76144abcc81b70c92b8aa295503f6121dfd99",
            "last_state": {
                "running": null,
[...]

PR #475, opened 2023-04-20
Record EC2 instance type for builder pods

zackgalbreath:
Update our OpenSearch records to include the type of instance that a job ran on.

### DIFF
--- a/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
+++ b/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
@@ -155,6 +155,14 @@ def collect_pod_status(job_input_data: dict[str, Any], job_trace: str):
             # In that case, we break out of the loop and return.
             break
 
+        # Attempt to get & record the type of EC2 instance that this pod ran on.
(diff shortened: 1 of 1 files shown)


mvandenburgh (approved):
It appears that no job has managed to grab pod info since 
#427 was merged yesterday. It seems that the runner pods are disappearing before the taxonomy job runs... but assuming I can get that working, this PR makes sense to me :+1:

zackgalbreath:
> It appears that no job has managed to grab pod info since #427 was merged yesterday. It seems that the runner pods are disappearing before the taxonomy job runs... but assuming I can get that working, this PR makes sense to me 👍

I was worried about that, but thanks for digging into it & letting us know. In the meantime I'll ponder alternate strategies to grab this info while it's available.

AlmightyYakob:
Any update on this?

mvandenburgh:
> It appears that no job has managed to grab pod info since #427 was merged yesterday. It seems that the runner pods are disappearing before the taxonomy job runs... but assuming I can get that working, this PR makes sense to me +1

I haven't revisited #427, so this is still the current status quo ^

PR #476, opened 2023-04-28
Add Cloned Gitlab Database and DMS change replication

AlmightyYakob:
This adds the necessary terraform to create a new RDS instance for analytics, and a DMS task which will copy and replicate new data into the databse (these infra changes are currently applied and running). It also includes updates to the gitlab failure log script, that will push job log records into the new database (as well as opensearch, for now).

Regarding the delay from data change in the prod database to the analytical one, it seems to replicate changes on the order of tens of minutes (but in most cases far less).


terraform/modules/spack/gitlab_db.tf
```diff
@@ -30,23 +88,350 @@ module "gitlab_db" {
...
+
+  allocated_storage     = 500
+  max_allocated_storage = 1000
+
+  vpc_security_group_ids = [module.postgres_security_group.security_group_id]
+}
+
+
+# Defines the Change Data Capture migration between the source and clone RDS instances
+module "database_migration_service" {
```
mvandenburgh, line 176: Should these two modules (the new RDS one and the DMS one) be in their own file? I think it's valuable to have a single .tf file that just represents "the gitlab database", without any of the analytics stuff which is technically not "required" for gitlab to work correctly. Maybe someone who's generally unfamiliar with how the analytics clone works might want to change something about the gitlab DB, and get confused by all the JSON blocks that configure the DMS replication. Any thoughts?

terraform/modules/spack/gitlab_db.tf
```diff
@@ -1,3 +1,61 @@
...
+  secret_id = data.aws_secretsmanager_secret.gitlab_db_clone_credentials.id
+}
+
+# Determine the corresponding EC2 instance size, in order to retrieve the memory and vcpu size
+locals {
+  gitlab_db_instance_ec2_class = replace(var.gitlab_db_instance_class, "db.", "")
+}
+data "aws_ec2_instance_type" "gitlab_db_instance_type" {
+  instance_type = local.gitlab_db_instance_ec2_class
+}
```
mvandenburgh, line 43: Why not just 
```suggestion
data "aws_ec2_instance_type" "gitlab_db_instance_type" {
  instance_type = replace(var.gitlab_db_instance_class, "db.", "")
}
```

images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
```diff
@@ -196,16 +257,9 @@ def main():
...
-        hosts=[OPENSEARCH_ENDPOINT],
-        http_auth=(
-            OPENSEARCH_USERNAME,
-            OPENSEARCH_PASSWORD,
-        ),
-    )
-    doc = JobPayload(**job_input_data)
-    doc.save()
+    # Push data to outputs
+    push_record_to_opensearch(job_input_data)
```
mvandenburgh, line 261: Just to clarify, the plan is to leave opensearch uploads in place for now until we have full confidence in the postgres replication, and then drop it, right?

mvandenburgh (commented):
Looks great! One more thing to note is that it looks like somewhere there are three new IAM roles being created by Terraform -
```
dms-access-for-endpoint
dms-cloudwatch-logs-role
dms-vpc-role
```

These are going to cause conflicts for prod vs staging, since IAM roles (and thus role names) are global. So somehow we will need to append `-${var.deployment_name}` or some other suffix to these roles.

terraform/modules/spack/gitlab_db.tf
AlmightyYakob: I think that's a good idea. Also, I'm realizing now that since this is at the module level it'll create it for both staging and prod, which we don't want. I can move this into a new file, and add the `count` meta argument so it only creates things for prod.

terraform/modules/spack/gitlab_db.tf
AlmightyYakob: No reason, thanks for the suggestion!

images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
AlmightyYakob: Correct, just playing it safe

terraform/modules/spack/gitlab_db.tf
mvandenburgh: Alternatively, you might also be able to move it into its own file in `terraform/production/xx.tf`, if this is for sure a "production-only" thing. Although if that requires adding several new outputs to the `spack` module, maybe it's not worth it, just throwing it out there. fwiw I did the `count` thing for OpenSearch because I thought maybe we may want to spin up a staging version of that someday to test upgrades, etc.

AlmightyYakob: Yeah I was trying to avoid polluting our outputs too much, but I'll see what works best.

AlmightyYakob:
@mvandenburgh This is good to go now. I resolved the issue I was having regarding foreign key constraints, and everything is running pretty smoothly now. Unfortunately the fix isn't easily encoded in terraform, but I documented the fix here: https://github.com/spack/spack-infrastructure/blob/7202c752b3b12657145695b83afd18a4cf74efa6/terraform/modules/spack/monitoring_db.tf#L99-L104

terraform/modules/spack/monitoring_db.tf
```diff
@@ -0,0 +1,440 @@
...
+            "object-locator" : {
+              "schema-name" : "public",
+              "table-name" : "%"
+            },
+            "rule-action" : "include",
+            "filters" : []
+          },
+          # ########################################
+          # Drop sensitive columns from users tables
+          # ########################################
```
mvandenburgh, line 141: What is the threat model driving this? Is there some security consideration with the analysis DB that doesn't apply to the real gitlab DB? Or are we just excluding this data to avoid storing a bunch of useless columns in the cloned DB?

terraform/modules/spack/monitoring_db.tf
```diff
@@ -0,0 +1,440 @@
...
+      tags          = { EndpointType = "source" }
+    }
+
+    destination = {
+      # Note: It's difficult to encode this into terraform at the moment (https://github.com/hashicorp/terraform-provider-aws/issues/23225),
+      # but the following setting needs to be entered into the "Endpoint Settings" of the target endpoint
+      # `AfterConnectScript="SET session_replication_role=replica"`
+
+      # This can also be done with the following command from the CLI:
+      # aws dms modify-endpoint --endpoint-arn 'ENDPOINT_ARN' --postgre-sql-settings='AfterConnectScript="SET session_replication_role=replica"'
```
mvandenburgh, line 104: :disappointed: that's unfortunate, guess we should keep an eye on that issue

mvandenburgh (commented):
Looks good, just had one question about the sensitive user data

terraform/modules/spack/monitoring_db.tf
AlmightyYakob: Yeah, aside from some very convoluted and brittle configuration, I'm not sure what else we can do.

mvandenburgh (approved):

AlmightyYakob:
This will not be merged due to the complications involved with doing full database replication. A separate and more simple approach will be taken.

PR #478, opened 2023-05-02
Terraform changes needed for cluster migration

mvandenburgh:
I'll be using the changes from this branch tomorrow when I spin the new cluster up. I've deployed this into an alternate AWS region and tested it extensively to make sure everything works as intended and that the clusters don't interfere with each others' Karpenter provisioning.

In the interest of minimizing the amount of things I need to think about regarding Terraform state and how we're going to spin up the new cluster alongside the existing one, I haven't touched the staging cluster in this PR; I'd like to postpone that until after everything is settled and we have the new cluster completely set up and working.


zackgalbreath (approved):

terraform/modules/spack/vpc.tf
```diff
@@ -6,25 +6,33 @@ module "vpc" {
   name = "spack-${var.deployment_name}"
   cidr = var.vpc_cidr
 
-  azs = var.availability_zones
-  public_subnets = var.public_subnets
-  private_subnets = var.private_subnets
-  database_subnets = var.database_subnets
-
-  # Create a DB subnet group for RDS (see rds.tf)
-  create_database_subnet_group = true
```
AlmightyYakob, line 15: Is `database_subnets` omitted because we don't use the `create_database_subnet_group` var? It seems this defaults to `true`, so wouldn't we want to explicitly set this `false`?

AlmightyYakob (commented):
Looks good overall, just one question

terraform/modules/spack/vpc.tf
mvandenburgh: Good catch, yes that should be explicitly set to false. I cherry-picked these changes out of a much messier branch and didn't run a plan with them yet, will update this

AlmightyYakob (approved):
LGTM

terraform/production/main.tf
```diff
@@ -96,24 +96,22 @@ provider "github" {
...
-    "10.0.160.0/19",
-  ]
-  database_subnets = [
-    "10.0.192.0/19",
-    "10.0.224.0/19",
+    "192.168.128.0/19",
+    "192.168.160.0/19",
+    "192.168.192.0/19",
+    "192.168.224.0/19",
   ]
```
AlmightyYakob, line 115: One thing I forgot to ask, why the change in these values?

mvandenburgh: I copied the old VPC config (from before the NAT gateway's were messed up) from [here](https://github.com/spack/spack-infrastructure/blob/91395c7d20a9bb50e48ec27133a3cc428a6df879/scripts/eks-vpc.yaml) as closely as possible. But yeah it wasn't actually necessary to change those values.

AlmightyYakob (approved):

PR #480, opened 2023-05-03
Update secret values

AlmightyYakob:
This updates and encodes new secret values for the runner registration token, as well as the values.yaml for gitlab and grafana


mvandenburgh (approved):

AlmightyYakob:
I'm going to wait to apply the `sealedsecrets.bitnami.com/managed: "true"` annotation until after this is merged. That way, I can be sure that the encrypted values match the existing secret values, and _then_ apply the labels, which will "release" all of the sealed secrets onto the existing secrets.

PR #481, opened 2023-05-03
signing script: Do not fail when there is nothing to sign

scottwittenburg:
Using the globbing pattern seems to produce an empty match when there are no spec files in the working directory.  Replacing the glob with the results of find prevents entering the loop unless there are any actual files in the directory.

Normally if there was nothing to rebuild, we would not schedule this job anyway.  So this is guarding an exceptional case, where all the scheduled rebuild jobs failed, something we saw when the version of gpg in one of the rebuild images was incompatible with one of the intermediate signing keys.


scottwittenburg:
Testing this is tricky since `sign.sh` can only run on protected pipelines in the `spack/spack` project on gitlab.spack.io.  But I tested a trimmed-down version locally to make sure this works with and without specs in the working directory.

zackgalbreath (approved):

PR #485, opened 2023-05-04
Encode Notary IAM role/policy and ServiceAccount into Terraform

AlmightyYakob:
Depends on #478, so please only look at the last commit when evaluating changes

### DIFF
--- a/k8s/production/runners/signing/serviceaccount.yaml
+++ b/k8s/production/runners/signing/serviceaccount.yaml
@@ -1,7 +1,0 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: notary
-  namespace: pipeline
-  annotations:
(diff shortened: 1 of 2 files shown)


scottwittenburg (commented):
Interesting, terraform is completely new to me.  So will this create these two resources (the notary role and associated policies) newly each time it's invoked?  And then it updates the kube yaml to match?

AlmightyYakob:
> Interesting, terraform is completely new to me. So will this create these two resources (the notary role and associated policies) newly each time it's invoked? And then it updates the kube yaml to match?

It won't create it _each_ time it's invoked, just the first time, and from there on if the resources are changed, it'll make sure it matches what's defined in code.

mvandenburgh (approved):
This looks great, and will definitely help mitigate these issues next time we do a cluster upgrade. I'm still cleaning up the TF state from last week, so after that is sorted out we can apply this.

mvandenburgh (approved):

PR #487, opened 2023-05-05
Add m7g and r7g to our list of graviton3 instance families

zackgalbreath:
https://aws.amazon.com/ec2/instance-types/m7g/
https://aws.amazon.com/ec2/instance-types/r7g/

### DIFF
--- a/k8s/production/karpenter/provisioners/runners/graviton/3/provisioners.yaml
+++ b/k8s/production/karpenter/provisioners/runners/graviton/3/provisioners.yaml
@@ -29,6 +29,8 @@ spec:
       operator: In
       values:
         - "c7g"
+        - "m7g"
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
Regarding the recent scheduling failures, are we assuming it was due to spot availability for `c7g` instances, and this is meant to widen the pool of nodes?

AlmightyYakob (approved):

zackgalbreath:
> Regarding the recent scheduling failures, are we assuming it was due to spot availability for `c7g` instances, and this is meant to widen the pool of nodes?

Precisely. I can't guarantee this will fix our issues with graviton3 runners failing to schedule, but in general giving Karpenter as many options as possible within its provisioners is a best practice.

PR #491, opened 2023-05-09
Update staging subnets

mvandenburgh:
This updates the staging cluster module to be in accordance with the changes from #478


AlmightyYakob (approved):

mvandenburgh:
Sadly this is not enough, since re-organizing subnets like this requires this requires the EKS cluster to be recreated.

PR #495, opened 2023-05-10
New public/private intermediate CI signing key

zackgalbreath:
Our old key expired on 2023-05-09


zackgalbreath:
I've locally downloaded a copy of the old public/private intermediate signing key, just in case we still need it for some reason 🤷

scottwittenburg (approved):
Thank you @zackgalbreath!

PR #498, opened 2023-05-10
Use editor to input sealed secret values

AlmightyYakob:
The secret update script will now open the user's configured editor (via the `$EDITOR` env var, defaults to `vim`) for inputting the secret value. This will solve the multi-line issue.

If anyone is on MacOS, please let me know how this works for you. I expect it to function the same, but I have no good way of testing it myself.

### DIFF
--- a/scripts/secrets/update.py
+++ b/scripts/secrets/update.py
@@ -126,7 +126,12 @@ def sealed_secret_cert_path(staging: bool) -> str:
     is_flag=True,
     help="Use the staging cert file.",
 )
-def main(secrets_file: str, staging: bool):
+@click.option(
+    "--value",
+    type=click.STRING,
+    help="Supply the value for the selected secret as an argument.",
+)
+def main(secrets_file: str, staging: bool, value: str):
     # Read in secrets file with comments
     yl = get_yaml_reader()
     with open(secrets_file) as f:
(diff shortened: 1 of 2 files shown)


mvandenburgh:
btw I just used this script to update the multiline secret in #502, and it worked great!

scripts/secrets/update.py
```diff
@@ -102,7 +104,11 @@ def represent_none(self, data):
 def sealed_secret_cert_path(staging: bool) -> str:
     env = "staging" if staging else "production"
     default_cert = (
-        Path(__file__).parent.parent / "k8s" / env / "sealed-secrets" / "cert.pem"
+        Path(__file__).parent.parent.parent
```
mvandenburgh, line 107: Something I learned about `pathlib` recently because I've found myself appending several `.parent`s to Path objects many times before :smile: 
```suggestion
        Path(__file__).parents[2]
```
Not that this is a big deal in this specific case, since three chained `.parents` is still readable imo, but thought i would share nonetheless

mvandenburgh (approved):
Works great! tested with both `vim` and `nano`.

scripts/secrets/update.py
AlmightyYakob: For some reason when it comes to usability I usually just do several `.parent` calls, but you're right that using `parents` is the better approach here.

mvandenburgh (approved):

PR #502, opened 2023-05-11
Apply NAT gateway changes to staging cluster

mvandenburgh:
This involves creating a new cluster, for the same reasons that the production cluster needed to be recreated. While I was at it, I also moved the staging cluster to `us-west-2` region, which has a similar number of availability zones as `us-east-1` which allows us to replicate our production NAT gateway setup in staging a bit more accurately.


AlmightyYakob (approved):
LGTM

PR #504, opened 2023-05-11
Update Karpenter AZs for new staging cluster region

mvandenburgh:
Forgot to update these after changing the region of the staging cluster in #502

### DIFF
--- a/k8s/staging/karpenter/kustomization.yaml
+++ b/k8s/staging/karpenter/kustomization.yaml
@@ -15,7 +15,7 @@ patches:
     patch: |-
       - op: replace
         path: /spec/requirements/2/values
-        value: ["us-west-1b", "us-west-1c"]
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
Should we also update this line? https://github.com/spack/spack-infrastructure/blob/10fd9e4b62da26470e38f5ce98848558d8775bab/terraform-bootstrap/staging/backend.tf#L13

mvandenburgh:
> Should we also update this line?
> 
> https://github.com/spack/spack-infrastructure/blob/10fd9e4b62da26470e38f5ce98848558d8775bab/terraform-bootstrap/staging/backend.tf#L13

The region of the S3 bucket storing the tfstate doesn't actually matter in relation to the rest of the resources. I suppose ideally that bucket should be in us-west-2 (or in us-east-1 with the other state bucket). But since moving it requires deleting and recreating it, I just left it alone for now.

AlmightyYakob (approved):

PR #505, opened 2023-05-12
Encode additional OIDC-dependent IAM roles/service accounts into TF, refactor existing roles/SAs

### LINKED ISSUE #486: Some roles broken after cluster upgrades
After the last cluster upgrade, and again after this one, some roles were no longer usable because under their "Trust relationships" in the AWS console, the oidc provider id from the old cluster was still listed.  This affected package signing, spackbot's PR binary graduation, and other things.  After the most recent cluster upgrade, I ran the following script provided earlier by @mvandenburgh:

```python
import boto3

client = boto3.client("iam")

old_oidc_id = "XXXXXXXXXXXXXXXXXXXXXXX"

for role in client.list_roles()["Roles"]:
    if old_oidc_id in str(role["AssumeRolePolicyDocument"]):
        role_name = role["RoleName"]
        role_arn = role["Arn"]
        print(f"{role_name} {role_arn}")
```

And that produced the following:

```console
FullCRUDAccessToBucketSpackBinariesPRs arn:aws:iam::588562868276:role/FullCRUDAccessToBucketSpackBinariesPRs
KarpenterControllerRole-spack arn:aws:iam::588562868276:role/KarpenterControllerRole-spack
KarpenterIRSA-spack-production-20230124204943653100000011 arn:aws:iam::588562868276:role/KarpenterIRSA-spack-production-20230124204943653100000011
PutObjectInPipelineStatistics arn:aws:iam::588562868276:role/PutObjectInPipelineStatistics
notary-role arn:aws:iam::588562868276:role/notary-role
```

I manually updated the trust relationships of those roles to use the new oidc provider id, but it would be good to encode this in TF or somewhere, so after the next upgrade it can be done automatically.

Also, @AlmightyYakob mentioned that maybe we no longer even need `KarpenterControllerRole-spack` and `KarpenterIRSA-spack-production-20230124204943653100000011`, so maybe those roles can be removed?  For now I just updated the oidc provider id, and will let @mvandenburgh decide whether they can be safely removed.

mvandenburgh:
This PR addresses the final two IAM roles I mentioned in [this comment ](https://github.com/spack/spack-infrastructure/issues/486#issuecomment-1542677935).

I also moved the notary role and service account out of the `modules/spack` TF module and into the top-level production cluster module. With it being in `modules/spack`, that means it is also being applied to the staging cluster as well, and I don't think granting the staging cluster read/write access to these KMS and S3 resources is ideal.

Closes #486


terraform/production/service_account_roles.tf
```diff
@@ -0,0 +1,200 @@
...
+          }
+        }
+      }
+    ]
+  })
+}
+
+resource "aws_iam_role_policy" "put_object_in_pipeline_statistics" {
+  name = "PutObjectInPipelineStatistics"
+  role = aws_iam_role.full_crud_access_spack_binaries_prs.id
```
AlmightyYakob, line 173:
```suggestion
  role = aws_iam_role.put_object_in_pipeline_statistics.id
```

AlmightyYakob (changes requested):
LGTM just one thing

terraform/production/service_account_roles.tf
mvandenburgh: Oh wow :sweat: , thanks for catching that!

terraform/production/service_account_roles.tf
```diff
@@ -0,0 +1,200 @@
...
+      {
+        "Effect" : "Allow",
+        "Action" : "s3:DeleteObject",
+        "Resource" : "arn:aws:s3:::spack-binaries-prs/*"
+      }
+    ]
+  })
+}
+
+resource "kubectl_manifest" "spackbot_service_account" {
```
AlmightyYakob, line 132: I think having the full role names in a comment here would be good just for searching through the code (as I tried that and failed as part of reviewing this PR), but otherwise LGTM

```suggestion

# Create the spackbot-spack-io and spackbotdev-spack-io ServiceAccounts
resource "kubectl_manifest" "spackbot_service_account" {
```

PR #508, opened 2023-05-12
Add sealed-secrets.yaml to runner staging kustomization

mvandenburgh:
#507 doesn't actually work, because I didn't add the `sealed-secrets.yaml` to the `resources` block in the staging kustomization. After doing that, I realized we also need to somehow exclude the other secrets in that YAML file from being included in staging. Documentation on how to remove an entire resources using kustomize seems very minimal, but I found this page that mentions how to do it and followed it https://pet2cattle.com/2022/11/kustomize-delete-resource. Unfortunately, the only way to find out if it works is to merge and see if Flux handles it ok.


AlmightyYakob (approved):
Looks good from what I can tell but have no way to really test/verify that.

PR #520, opened 2023-05-25
Add a dashboard viewing stats for specific refs

scottwittenburg:
Add a dashboard with panels taken from (or modeled after) existing panels in other dashboards, but restricted to a `ref` the viewer can specify.  This allows us to do a post-mortem on a release process, for example.

ref-specific-statistics_02

### DIFF
--- a/k8s/production/prometheus/custom/ref-specific-statistics.yaml
+++ b/k8s/production/prometheus/custom/ref-specific-statistics.yaml
@@ -1,0 +1,523 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: monitoring
+  name: kube-prometheus-stack-ref-specific-statistics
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@zackgalbreath I didn't see anywhere in the yaml pile where each dashboard needs to be referenced.  Do you know if some magic just reads everything in the `custom` directory and applies it?

scottwittenburg:
Just realized: the fact that these dashboards have a default time range they come with (usually the last 7 days, but in this case the last 12) may cause a little confusion down the road.  It means we'll have to make sure to enter a date range during which pipelines for the `targetRef` were actually running.  But I still think it's nice to be able to specify the `targetRef` as a variable.

zackgalbreath (approved):

PR #521, opened 2023-05-26
Add oauth2-proxy for prometheus installation

danlamanna:
This gates prometheus behind a Github OAuth login in favor of the broken mTLS workflow we currently have. This is currently configured on staging but will need some work to be applied to production. 

cc @mvandenburgh


k8s/production/ingress-nginx/oauth2-proxy.yml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
```
mvandenburgh, line 1: Should we move this `Deployment` and `Service` to `k8s/production/prometheus/ingress.yaml`? Or maybe just move this file itself to `k8s/production/prometheus/`? Since this is oauth2-proxy deployment is specific to prometheus.spack.io, and presumably we'll want to use this for alertmanager.spack.io in the future as well?

danlamanna: Done.

PR #523, opened 2023-05-30
Update sealed secrets for opensearch index job

mvandenburgh:
The `opensearch-index-build-logs` job is currently failing because the secrets for it have become outdated. This PR updates the corresponding `SealedSecret` resource with the correct values.

PR #524, opened 2023-05-30
Add missing json file to opensearch ingest image

mvandenburgh:
The cronjob that crawls the build cache and records timing info, etc. into opensearch is currently failing. I thought #523 would fix it, but even after applying those changes it still fails. The error logs indicated that this .json5 is missing; my guess is that the old cluster was using the image from my fork (https://github.com/mvandenburgh/docker-images/pkgs/container/opensearch-index-build-logs) which does have that file, and when we switched to the new cluster, Flux reset it to the old image and it started failing.


AlmightyYakob (approved):
LGTM but I have no idea what the json5 file represents

PR #526, opened 2023-05-31
Suspend the stuck pipelines cronjob

scottwittenburg:
This mitigation is currently not having the intended effect, as we have seen months-old pipelines still stuck in the "running" state, even with this cronjob running every three hours since forever.  We are also seeing multiple repeated develop pipelines (and pipelines for some PRs) running on the same sha.  Since this mitigation seems to be doing more harm than good, let's pause it until we can figure out what's wrong with it (or what's wrong with gitlab).

### DIFF
--- a/k8s/production/custom/unstick-pipelines/cron-jobs.yaml
+++ b/k8s/production/custom/unstick-pipelines/cron-jobs.yaml
@@ -4,6 +4,7 @@ metadata:
   name: cancel-and-restart-stuck-pipelines
   namespace: custom
 spec:
+  suspend: true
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@mvandenburgh @zackgalbreath Is this the best way to pause this cronjob for awhile?  This thing is a workaround for a gitlab bug that keeps finished pipelines marked as "running" forever.  But it seems there may be other bugs in gitlab preventing the workaround from actually working.  So I think it may be best to pause it until we can figure out what's going on.

mvandenburgh (approved):

PR #527, opened 2023-05-31
Proposed overhaul of building/deployment of custom pcluster AMIs

mvandenburgh:
## Background
Currently, we have a need to run a specific AMI on some of our gitlab runner nodes. This pcluster AMI was provided by Amazon and is not "[EKS-optimized](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-amis.html)". That is, it is not configured with the software needed for it to join an EKS cluster as a node. AWS provides a [Packer template](https://github.com/awslabs/amazon-eks-ami) that takes in a source AMI and builds a custom "EKS-optimized" version of that AMI. I did that using the pcluster AMI as the source AMI, and after pointing the pcluster Karpenter provisioner at the resulting "EKS-optimized" AMI, the nodes are able to join the cluster as expected. 

There is one caveat to this solution - the Packer template is coupled to the version of kubernetes running on the cluster; for the pcluster AMIs that I built manually, I had to specify `1.24` as the k8s version as part of the Packer build process.
This introduces some technical debt in that we will have to remember to rebuilt these AMIs when we move to a new version of kubernetes.

## Solution
This PR attempts to alleviate this by doing the following:

- The custom "EKS-optimized" pcluster AMIs are now built in CI. The kubernetes version is extracted from the Terraform state at build time so that we can ensure AMIs are always available for the current version of the cluster.
- The karpenter node template now refers to the AMIs by name instead of ID ([docs](https://karpenter.sh/preview/concepts/node-templates/#specamiselector)), and the name contains the k8s version string. This ensures that the karpenter node template is always referring to the corresponding pcluster AMI for the current kubernetes version, even if the AMI gets rebuilt.

### DIFF
--- a/.github/workflows/build_pcluster_amis.yml
+++ b/.github/workflows/build_pcluster_amis.yml
@@ -1,0 +1,97 @@
+name: Build pcluster AMIs
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+# Cancel any workflows that are already running
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  build:
+    name: Build AMI [${{ matrix.architecture }}, ${{ matrix.aws-region }}]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - architecture: x86_64
+            source-ami-id: ami-011244318c573c915
+            aws-region: us-east-1
+          - architecture: arm64
+            source-ami-id: ami-0d87b43677ef96ab9
+            aws-region: us-east-1
+          - architecture: x86_64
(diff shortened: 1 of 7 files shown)


.github/workflows/build_pcluster_amis.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: hashicorp/setup-terraform@v2
+        with:
+          # Disable the github actions "wrapper" for Terraform
+          # because it transforms the output such that it can't
+          # be read by `jq`.
+          terraform_wrapper: false
+
+      - name: Initialize terraform
+        working-directory: terraform/production
+        run: terraform init
```
danlamanna, line 69:
```suggestion
        run: terraform init -input=false
```
https://developer.hashicorp.com/terraform/tutorials/automation/automate-terraform#automated-terraform-cli-workflow

danlamanna (dismissed):

danlamanna (approved):

PR #529, opened 2023-06-05
Update prometheus to use a gp3 volume

zackgalbreath:


mvandenburgh (approved):

mvandenburgh:
I think we'll need to manually recreate the persistent volume in kubernetes in order for this to take effect.

PR #534, opened 2023-06-08
Terraform: Add S3 bucket policies for reading protected data

kwryankrattiger:
Add a terraform section for maintaining S3 bucket policies for protected access to data.

The purpose of this addition is to prevent the download of package binaries which have license or other legal restrictions on redistribution. This issue may potentially be resolved by other means for the current list of packages, but a method for maintaining bucket policies with Terraform makes it easier to adapt to these constraints in the future as the come up.

**Note** I think these policies require further review to ensure we are remaining in compliance with the current bucket policies. 

CC: @AlmightyYakob @mvandenburgh @zackgalbreath @scottwittenburg

### DIFF
--- a/terraform/production/s3.tf
+++ b/terraform/production/s3.tf
@@ -1,0 +1,66 @@
+locals {
+  s3_bucket_policies = {
+    "spack-binaries" = {
+      resources = [
+        "arn:aws:s3:::spack-binaries/*/armpl-*",
+        "arn:aws:s3:::spack-binaries/*/intel-*"
+      ]
+      allowed = "arn:aws:iam::588562868276:user/protected-binary-mirror"
+    }
+
+    "spack-binaries-prs" = {
+      resources = [
+        "arn:aws:s3:::spack-binaries-prs/*/armpl-*",
+        "arn:aws:s3:::spack-binaries-prs/*/intel-*"
+      ]
+      allowed = "arn:aws:iam::588562868276:user/pull-requests-binary-mirror"
+    }
+
+    "spack-binaries-cray" = {
+      resources = ["*"]
+      allowed = "arn:aws:iam::588562868276:user/cray-binary-mirror"
+    }
+  }
+}
+
+resource "aws_s3_bucket_policy" "protected_binaries_restricted" {
+  for_each = local.s3_bucket_policies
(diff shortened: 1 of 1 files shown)


terraform/modules/spack/s3.tf
```diff
@@ -0,0 +1,41 @@
...
+        "Effect": "Deny",
+        "Principal": {
+          "AWS": "${each.value.allowed}"
+        },
+        "Action": "s3:GetObject",
+        "Resource": each.value.resources
+      }
+    ]
+  })
+}
```
mvandenburgh, line 41: If this is only applicable to production, could we have these be a part of the root `production` module instead of the `spack` module? i.e. similar to how we have [these IAM roles and policies](https://github.com/spack/spack-infrastructure/blob/main/terraform/production/service_account_roles.tf) defined outside of the `spack` module since they are only used in production.

terraform/modules/spack/variables.tf
```diff
@@ -63,3 +63,11 @@ variable "provision_opensearch_cluster" {
...
   type        = bool
 }
+
+variable "s3_bucket_policies" {
+  description = "List of buckets mapped to object patterns that are protected and user/role that is allowed"
+  type        = map(object({
+    resources = list(string)
+    allowed   = string
+  }))
+}
```
mvandenburgh, line 73: Implementing the above change would also let us avoid adding this variable here

terraform/production/main.tf
```diff
@@ -125,4 +125,33 @@ module "production_cluster" {
   gitlab_db_master_credentials_secret = "arn:aws:secretsmanager:us-east-1:588562868276:secret:gitlab-prod-master-credentials-96P0Cl"
 
   provision_opensearch_cluster = true
+
+  # s3_bucket_policies = {
+  #   "spack-binaries-develop" = {
+  #     resources = ["arn:aws:s3:::spack-binaries-develop/*hello*"]
+  #     allowed = "arn:aws:iam::588562868276:user/develop-binary-mirror"
+  #   }
+  # }
```
mvandenburgh, line 134:
```suggestion
```

mvandenburgh (approved):

PR #536, opened 2023-06-09
Fix kube-event-exporter configuration

scottwittenburg:
This hopefully fixes the broken `kube-event-exporter` by setting different configuration values:

``` 
kubeQPS: 3
kubeBurst: 15
```

Also adds a couple tips to the `README.md` in case others run into the same issues I did.

PR #538, opened 2023-06-12
Fix event-exporter sealed-secret

mvandenburgh:
This updates the sealed secret for event-exporter to fix the wrongly encoded newlines. The secret now looks like this when decrypted:

```
kubeQPS: 100 # 1/5 of Burst
kubeBurst: 500 # estimate of events per minute
maxEventAgeSeconds: 60
logLevel: error
logFormat: json
route:
  # Main route
  routes:
    # This route allows dumping all events because it has no fields to match and no drop rules.
    - match:
        - receiver: "runner-events"
        - namespace: "pipeline|gitlab"
receivers:
  - name: "runner-events"
    opensearch:
      hosts:
        - <redacted>
      index: kube-events
      indexFormat: "kube-events-{2006-01-02}"
      username: <redacted>
      password: <redacted>
      # If set to true, it allows updating the same document in ES (might be useful handling count)
      useEventID: false
      deDot: true
```


AlmightyYakob (approved):
LGTM. Was there any issues with newlines when updating this? It looks like the last time this was touched was when I mass encoded existing cluster secrets, before multi-line secrets were properly addressed, so I'm hoping the issue that caused this doesn't still exist.

mvandenburgh:
> LGTM. Was there any issues with newlines when updating this? It looks like the last time this was touched was when I mass encoded existing cluster secrets, before multi-line secrets were properly addressed, so I'm hoping the issue that caused this doesn't still exist.

I didn't have any issues using the `scripts/secrets/update.py` script to update it. But yeah it looks like this was one of those secrets that got messed up in the initial encoding and got forgotten about until now.

mvandenburgh:
I didn't realize there's already a PR for this, #536, I'll close this one in favor of that

PR #546, opened 2023-06-13
Upgrade GitLab

mvandenburgh:
Note - this should not be merged yet. There are several intermediate upgrades that need to be performed with Flux turned off; this PR encodes what the final state of the cluster should be after the upgrade is complete.

### DIFF
--- a/k8s/production/gitlab/release.yaml
+++ b/k8s/production/gitlab/release.yaml
@@ -86,7 +86,7 @@ spec:
           key: postgres-password
         port: 5432
         database: gitlabhq_production
-        username: postgres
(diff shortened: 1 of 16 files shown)


k8s/production/gitlab/release.yaml
```diff
@@ -86,7 +86,7 @@ spec:
           key: postgres-password
         port: 5432
         database: gitlabhq_production
-        username: postgres
+        username: gitlab
```
AlmightyYakob, line 89: Why the username change?

AlmightyYakob (approved):

PR #549, opened 2023-06-14
Schedule oauth2-proxy on base nodes

danlamanna:


k8s/staging/prometheus/kustomization.yaml
```diff
@@ -38,7 +38,7 @@ patches:
   - target:
       kind: Ingress
       name: oauth2-proxy
-      namespace: monitoring
+      namespace: ingress-nginx
```
mvandenburgh, line 41: I think this wasn't supposed to be changed

mvandenburgh: Actually i'm wrong, disregard

PR #557, opened 2023-06-20
Remove patched gitlab images, gitops image

mvandenburgh:
Removes a bunch of docker images that are no longer used.


AlmightyYakob (approved):

AlmightyYakob:
@mvandenburgh For posterity can you remind us why we switched back to using the stock gitlab images? Was it simply that with the gitlab upgrade, the crafted images were rendered useless?

scottwittenburg:
> why we switched back to using the stock gitlab images?

The images we used before were patched to ignore the max yaml size limit.  Now we are increasing the limit with that cronjob which invokes the rails runner.

scottwittenburg (approved):
Approving, but only for the removal of the `gitlab-*` docker files.  Hopefully someone else can chime in on the `gitops` files.

mvandenburgh:
> Approving, but only for the removal of the `gitlab-*` docker files. Hopefully someone else can chime in on the `gitops` files.

Maybe @zackgalbreath can confirm. I believe that `gitops` image was a custom thing Omar made related to the old Flux v1 setup with the "staging" gitlab deployment (before we had an entire staging cluster). Either way, it isn't used elsewhere in the repo from what I can tell.

zackgalbreath:
> Maybe @zackgalbreath can confirm. I believe that `gitops` image was a custom thing Omar made related to the old Flux v1 setup with the "staging" gitlab deployment (before we had an entire staging cluster). Either way, it isn't used elsewhere in the repo from what I can tell.

This is correct and it can be safely removed.

zackgalbreath (approved):

PR #558, opened 2023-06-21
Fix staging email sending, encode SES configuration into TF

mvandenburgh:
Currently, the staging gitlab can't send emails to users because there's no SES configuration set up for it. This PR encodes the existing production SES config into Terraform, which also allows us to easily spin up an SES configuration for staging.


mvandenburgh:
I applied this to staging, and it works. Once this is approved and ready to merge, I will run the `terraform import` commands needed to integrate the existing SES setup in production.

AlmightyYakob (commented):
Looks good just one question

AlmightyYakob (approved):

PR #561, opened 2023-06-22
Add cronjob to create binary cache summary index

scottwittenburg:
This builds a new image containing a python script that iterates over `s3://spack-binaries` to build a summary index describing all the mirrors that the cache.spack.io website should contain.  Then the image is used in a nightly cron job that builds the index and pushes to the root of the `s3://spack-binaries` bucket.

This still needs some updates to the `service_account_roles.tf` file to create a role granting read/write permission to the spack-binaries bucket.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -48,6 +48,8 @@ jobs:
             image-tags: ghcr.io/spack/upload-gitlab-failure-logs:0.0.1
           - docker-image: snapshot-release-tags
             image-tags: ghcr.io/spack/snapshot-release-tags:0.0.2
+          - docker-image: cache-indexer
+            image-tags: ghcr.io/spack/cache-indexer:0.0.1
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/custom_docker_builds.yml
```diff
@@ -48,6 +48,8 @@ jobs:
             image-tags: ghcr.io/spack/upload-gitlab-failure-logs:0.0.1
           - docker-image: snapshot-release-tags
             image-tags: ghcr.io/spack/snapshot-release-tags:0.0.2
+          - docker-image: cache-indexer
+            image-tags: ghcr.io/spack/cache-indexer:0.0.1
```
danlamanna, line 52: Is there a reason we should pin this?

danlamanna:
@scottwittenburg I made a minor change to use `boto3` directly so we can get better error messages (if they happen) and don't have to muck with stdout/stderr. I added the service account. I'm planning on trying to deploy this later today unless it looks like there are any show stoppers that you can see?

mvandenburgh (approved):

PR #568, opened 2023-07-10
Make gp3 the default StorageClass

zackgalbreath:
This new StorageClass is a copy of our previous default except that `gp2` has been replaced with `gp3`.

### DIFF
--- a/k8s/production/kube-system/gp3-storage-class.yaml
+++ b/k8s/production/kube-system/gp3-storage-class.yaml
@@ -4,7 +4,7 @@ metadata:
   name: gp3
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
-provisioner: kubernetes.io/aws-ebs
+provisioner: ebs.csi.aws.com
 parameters:
   type: gp3
   fsType: ext4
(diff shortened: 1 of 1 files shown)


k8s/production/kube-system/gp3-storage-class.yaml
```diff
@@ -0,0 +1,11 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: gp3
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/aws-ebs
```
mvandenburgh, line 7: This provisioner is no longer supported (see [the docs](https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi-migration-faq.html#csi-migration-faq-aws-ebs-provisioner)). The new way is to use the Amazon EBS CSI driver - see https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html.

What do you think about migrating to that while we're here? Since we have the EBS CSI driver installed, it looks like the only change needed for that is
```suggestion
provisioner: ebs.csi.aws.com
```

PR #579, opened 2023-07-21
Standardize the number of GitLab runner pods and concurrency values

alecbcs:
This PR lowers the number of replica GitLab runner pods from 6 to 2 to reduce the number of idle pods in the k8s cluster. It also increases the concurrency of each runner to 75 and decreases the polling interval to 10s which mirrors the values we use internally at LLNL on login nodes, after we found those to be a good fit. Since our runners here mirror the functionality of login node runners, I thought it'd be a good fit to mirror those settings here.


zackgalbreath:
any thoughts on how we can test increased concurrency without just "going for it"?

PR #580, opened 2023-07-21
Add Job ID label to all build job pods

AlmightyYakob:
This change will ensure every pipeline build pod will have it's Job ID set as a pod label, which is useful for Prometheus metrics analysis.

cc @alecbcs


zackgalbreath (approved):

alecbcs:
Woo! Thanks @AlmightyYakob!

PR #581, opened 2023-07-21
Replace gitlab minio instance with S3 buckets

mvandenburgh:
I applied this to staging and it works.

In addition to disabling Minio and pointing gitlab at s3, this PR also
- Adds a bucket policy to each gitlab-managed bucket preventing deletion of the bucket.
- Replaces the "log_cleaner" cron job that runs in the cluster and deletes job artifacts older than 30 days with an S3 lifecycle policy that does the same thing.


k8s/production/gitlab/release.yaml
```diff
@@ -101,6 +103,10 @@ spec:
       appConfig:
         smartcard:
           enabled: false
+        lfs:
+          enabled: false
+        packages:
+          enabled: false
```
mvandenburgh, line 109: We don't use these services, and disabling them lets us avoid creating S3 buckets for them.

terraform/modules/spack/gitlab_object_stores.tf
```diff
@@ -0,0 +1,189 @@
...
+              bucket: ${aws_s3_bucket.gitlab_object_stores["backups"].id}
+              tmpBucket: ${aws_s3_bucket.gitlab_object_stores["tmp-storage"].id}
+        # https://docs.gitlab.com/charts/advanced/external-object-storage/#backups
+        gitlab:
+          toolbox:
+            backups:
+              objectStorage:
+                config:
+                  secret: ${local.backups_secret_name}
+                  key: ${local.backups_secret_key}
```
mvandenburgh, line 156: We don't use gitlab's backup service currently (although we should look into it i think), but gitlab still requires buckets for them.

k8s/production/gitlab/release.yaml
mvandenburgh: (@zackgalbreath or @scottwittenburg can you confirm that we don't use these?)

danlamanna (approved):

k8s/production/gitlab/release.yaml
zackgalbreath: That is my understanding. We don't host packages at gitlab.spack.io, and spack's GitHub repository doesn't use git lfs, so there's no need for it on the gitlab mirror.

PR #584, opened 2023-07-27
Create runner for service jobs

AlmightyYakob:
This creates a separate runner for service jobs, so that they can be placed on any node, instead of just nodes that are part of the `x86-64-v2` or `glr-graviton2` node pools. This will fix a situation that can occur where, if no nodes from these node pools are spun up (specifically, if the karpenter provisioner for these node pools don't have any nodes active at the moment), then karpenter will spin up a new node from that node pool _just_ for the service job, when in reality, it could have run on almost any node with capacity. The new runner has no node selector, and so will run on any node that's capable.

This is a symptom of a more general problem with our infrastructure setup, which is that the concept of "node pools" results in pods being placed in discrete piles of nodes. In reality, our pod placement should be more fluid than that (i.e. "minimum requirements" instead of "exact requirements"). My open PR #446 will address that once it's fixed up and ready.

CC @alecbcs, this slightly conflicts with #579

### DIFF
--- a/k8s/production/runners/public/x86_64/v2/release.yaml
+++ b/k8s/production/runners/public/x86_64/v2/release.yaml
@@ -105,7 +105,7 @@ spec:
       imagePullPolicy: "if-not-present"
       locked: false
 
-      tags: "x86_64,x86_64_v2,small,medium,large,huge,public,aws,spack,service"
+      tags: "x86_64,x86_64_v2,small,medium,large,huge,public,aws,spack"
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
Marking this as draft as I think it will in fact be rendered obsolete by #446

AlmightyYakob:
Superseded by #446, as with that PR, any service jobs would be scheduled on any pipeline node that matches their required architecture.

PR #585, opened 2023-07-27
Build ARM variants of docker images

mvandenburgh:
This PR updates our custom docker build action to also build ARM variants of the images. I also bumped the docker actions to the latest versions to ensure they work with these parameters.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -46,8 +46,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Log in to the Container registry
-        uses: docker/login-action@d398f07826957cd0a18ea1b059cf1207835e60bc
+        uses: docker/login-action@v2
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -55,12 +58,13 @@ jobs:
 
       - name: Build and push ${{ matrix.docker-image }}
         id: docker-build-push
-        uses: docker/build-push-action@67af6dc1d38f334ae6935c94587e8a5b45a81a0e
+        uses: docker/build-push-action@v4
         with:
           context: ./images/${{ matrix.docker-image }}
           file: ./images/${{ matrix.docker-image }}/Dockerfile
           push: true
           tags: ${{ matrix.image-tags }}
+          platforms: linux/amd64,linux/arm64
 
       - name: Image digest
         run: echo ${{ steps.docker-build-push.outputs.digest }}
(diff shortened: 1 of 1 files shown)


.github/workflows/custom_docker_builds.yml
```diff
@@ -46,21 +46,25 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Log in to the Container registry
-        uses: docker/login-action@d398f07826957cd0a18ea1b059cf1207835e60bc
+        uses: docker/login-action@v2
```
scottwittenburg, line 53: Should we stick with the practice of using the sha here, as github [recommends](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)?  

We could, however, include a line-end comment with the corresponding version.  I think dependabot can handle that, including updating the version comment, so if we're (justifiably) worried about falling behind, we might lean on dependabot for that.

I'm not insisting, maybe just curious what your thoughts are.

mvandenburgh: Ah i didn't know github recommends that. In that case, I agree we should do that - I've updated the PR.

mvandenburgh: Hmm I did a `git push`, but I'm not seeing it in the github ui...

mvandenburgh: Not sure whats going with github, but you can see the new changes if you view my branch directly https://github.com/spack/spack-infrastructure/tree/build-arm-images

.github/workflows/custom_docker_builds.yml
```diff
@@ -44,23 +44,27 @@ jobs:
...
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5 # v3.5.3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@2b82ce82d56a2a04d2637cd93a637ae1b359c0a7 # v2.2.0
 
       - name: Log in to the Container registry
-        uses: docker/login-action@d398f07826957cd0a18ea1b059cf1207835e60bc
+        uses: docker/login-action@a9794064588be971151ec5e7144cb535bcb56e36 # v2.2.0
```
scottwittenburg, line 53: To me, this looks like a [commit](https://github.com/docker/login-action/commit/a9794064588be971151ec5e7144cb535bcb56e36) on `master`, but maybe I'm not reading it right?  For the tagged release version `2.2.0` (I'm looking [here](https://github.com/docker/login-action/releases/tag/v2.2.0)), I would have thought you'd want `465a07811f14bebb1938fbed4728c6a1ff8901fc`.

mvandenburgh: You're right, my bad. I'm not sure where I got those hashes from, I must have mistakenly copied the latest commit to `master` instead. I pushed the correct ones.

scottwittenburg (approved):
Verified the versions/shas, looks good to me.  Thanks @mvandenburgh

PR #590, opened 2023-08-03
Fix broken python-aws-bash image build

scottwittenburg:
After a recent PR merge, I noticed the `python-aws-bash` image [failing](https://github.com/spack/spack-infrastructure/actions/runs/5753437333/job/15596635909#step:4:326) to build.

I think it was pinning `awscli` to an older version that was causing the build to hit this [issue](https://github.com/yaml/pyyaml/issues/601) with older versions of `pyyaml`.

As you can see in the deleted comment from this PR, there was a bug in older version of `boto3` that was preventing us from moving forward, but that bug has been fixed, and anyway spack had been [fixed](https://github.com/spack/spack/pull/31610) a year ago to support both older and newer versions of `boto3`.   So I thought we could move forward and remove all the pinning from these dependencies.

Just to be sure, I tested that I could `spack buildcache update-index <mirror-url>` with the new image resulting from removing all the version pinning, and it seems to work fine.  

However, this recent build failure makes me wonder if we should really be re-building, re-tagging, and pushing all these images with every merge to `main`.  If some breakage creeps in via dependencies, then overwriting the tags the way we do will cause us to lose the previously working version and start using the broken one immediately (and not be able to go back).  Or maybe I'm missing some detail that prevents this from happening.

### DIFF
--- a/images/python-aws-bash/Dockerfile
+++ b/images/python-aws-bash/Dockerfile
@@ -1,10 +1,7 @@
 FROM python:3
 
-# bug in spack s3 handling requires older botocore:
-#     https://github.com/spack/spack/issues/28830
-RUN pip install --upgrade \
-    awscli==1.22.46 \
-    boto3==1.20.35 \
-    botocore==1.23.46
+WORKDIR /scripts/
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
 
 ENTRYPOINT ["/bin/bash"]
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -32,7 +32,7 @@ jobs:
           - docker-image: ./images/notary
             image-tags: ghcr.io/spack/notary:latest
           - docker-image: ./images/python-aws-bash
-            image-tags: ghcr.io/spack/python-aws-bash:0.0.1
+            image-tags: ghcr.io/spack/python-aws-bash:0.0.2
           - docker-image: ./images/opensearch-index-build-logs
             image-tags: ghcr.io/spack/opensearch-index-build-logs:0.0.4
           - docker-image: ./images/gitlab-error-processor
--- a/images/python-aws-bash/requirements.txt
+++ b/images/python-aws-bash/requirements.txt
@@ -1,0 +1,3 @@
+awscli
+boto3
+botocore


mvandenburgh:
> However, this recent build failure makes me wonder if we should really be re-building, re-tagging, and pushing all these images with every merge to `main`. If some breakage creeps in via dependencies, then overwriting the tags the way we do will cause us to lose the previously working version and start using the broken one immediately (and not be able to go back). Or maybe I'm missing some detail that prevents this from happening.

This is mostly correct. We rebuild images and rewrite the tags whenever a file in the `images/` directory is modified (see [here](https://github.com/spack/spack-infrastructure/blob/main/.github/workflows/custom_docker_builds.yml#L3-L8)) in `main`. But that still has the potential to cause the problem you describe; I assume what happened here is _another_ image was modified, which caused the `python-aws-bash` image to get rebuilt with the modified dependencies.

The two options I can think of are 
1) Instrument the github action to only rebuild images that are actually modified,
2) Continue to pin the dependencies, and just update them in this PR.

I'm not sure what's better, although my gut feeling leans towards 1). Then if a breakage occurs due to an updated dependency, we can at least anticipate it since presumably that image was just recently rebuilt.

images/python-aws-bash/Dockerfile
```diff
@@ -1,10 +1,8 @@
...
 
-# bug in spack s3 handling requires older botocore:
-#     https://github.com/spack/spack/issues/28830
 RUN pip install --upgrade \
-    awscli==1.22.46 \
-    boto3==1.20.35 \
-    botocore==1.23.46
+    awscli \
+    boto3 \
+    botocore
```
mvandenburgh, line 6: Could these be moved to a `requirements.txt` file like we have for the rest of our custom docker images?

scottwittenburg: Yes definitely, I don't know why I didn't do that originally.

scottwittenburg:
> I assume what happened here is another image was modified, which caused the python-aws-bash image to get rebuilt

Yes, exactly.

> Instrument the github action to only rebuild images that are actually modified,

This would be nice and maybe not too hard, I'll take a stab at it in a new PR.

    Continue to pin the dependencies, and just update them in this PR.

I always lean toward pinning, a cognitive bias likely based on fear of untried things and failure.  But I'm not sure we want that if we can avoid it.  Picking up bugfixes is good, and as you said, if we do (1), we'll be less surprised when it breaks.

scottwittenburg:
> This would be nice and maybe not too hard, I'll take a stab at it in a new PR.

I did try this out in #618.

scottwittenburg:
This will fix one of the broken images we see [here](https://github.com/spack/spack-infrastructure/actions/runs/6190686244/job/16807426614), not sure what's going on with the opensearch one yet.

scottwittenburg:
@mvandenburgh mind taking another look at this?  I addressed the concern you raised about putting the requirements into their own file, and it fixes one of the broken builds.  I've rebased on your buildx fix too, so it should be good to go.

mvandenburgh (approved):

PR #591, opened 2023-08-03
Replace self-hosted redis with AWS elasticache

mvandenburgh:
To do this, I pretty much just followed this document - https://docs.gitlab.com/charts/advanced/external-redis/

I based the instance type off of the [gitlab reference architecture for 3,000 users](https://docs.gitlab.com/ee/administration/reference_architectures/3k_users.html), which is the minimum tier that is considered "high availability".

This is already applied in staging.


k8s/production/gitlab/release.yaml
```diff
@@ -85,8 +89,9 @@ spec:
         username: gitlab
 
       redis:
-        password:
-          enabled: true
+        host: <retrieved from TF-managed configmap>
```
danlamanna, line 92: This has to be defined here to be overridden later?

AlmightyYakob (commented):
Looks good to me overall, aside from the question Dan raised

k8s/production/gitlab/release.yaml
mvandenburgh: No, I included it as documentation of sorts to make it easier to see the structure of that field without having to look at the TF config map too. If it introduces more confusion I'm open to changing it.

AlmightyYakob: I think it would be sufficient to add a comment to the top of the `redis` block, explaining that it's configured via the terraform configmap.

AlmightyYakob: Sorry, I meant that we should remove this field and just rely on the comment (since that's essentially what the value of `host` is now).

mvandenburgh: Oh, my bad. That's fine with me, I updated it

danlamanna (dismissed):

k8s/production/gitlab/release.yaml
```diff
@@ -84,9 +88,11 @@ spec:
         database: gitlabhq_production
         username: gitlab
 
+      # This block will be overwritten by the TF-managed configmap
+      # specified in the `valuesFrom` block above.
       redis:
-        password:
-          enabled: true
+        auth:
+          enabled: false
```
AlmightyYakob, line 95: One more question, why does this need to be specified here? Isn't it specified in the config map as well?

mvandenburgh: Yeah, good catch, i missed that when I removed the hostname. We don't need this block at all.

AlmightyYakob (approved):

PR #594, opened 2023-08-14
Analytics Database and Build Time Processor

AlmightyYakob:
This PR adds an analytics database, with which arbitrary data can be stored and queried from, as well as a build timing processor, which processes succeeded jobs and their respective phase timings. To accomplish this, several components have been added/configured:

1. A terraform configuration to create the database
2. A very small Django app (no website or server) to be used for model/table migration and management, located at `analytics`
3. A build timing processor webhook listener, which mimics the structure of the `gitlab-error-processor`. This http server listens for received webhooks from gitlab, and dispatches a kubernetes job to run the `upload_build_timings` analytics management command.
4. A staging kustomization to support this configuration in the staging cluster
5. A github workflow for checking the django app migrations (fails if migrations need to be run).

**Note**: The Terraform changes are already applied to production and running correctly


images/build-timing-processor/app.py
```diff
@@ -0,0 +1,66 @@
...
+app = FastAPI()
+
+
+BUILD_STAGE_REGEX = r"^stage-\d+$"
+
+
+@app.post("/")
+async def gitlab_webhook_consumer(request: Request):
+    """
+    This endpoint receives the gitlab webhook for failed jobs and creates
```
scottwittenburg, line 22: Why is it just failed jobs?  Or maybe the comment is just copy-pasted from the error classifier webhook?

AlmightyYakob: Yep my bad, that's a leftover comment.

analytics/analytics/migrations/0001_initial.py
```diff
@@ -0,0 +1,81 @@
...
+                    models.PositiveBigIntegerField(primary_key=True, serialize=False),
+                ),
+                ("project_id", models.PositiveBigIntegerField()),
+                ("name", models.CharField(max_length=256)),
+                ("started_at", models.DateTimeField()),
+                ("duration", models.FloatField()),
+                ("package_name", models.CharField(max_length=128)),
+                ("microarch", models.CharField(max_length=32)),
+                ("aws", models.BooleanField(default=True)),
+            ],
```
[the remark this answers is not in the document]
AlmightyYakob: I believe you're correct, the ref is included as part of the `ProjectJob` object. I think it'd be perfectly fine to add it as a field on the `Job` model.

analytics/analytics/management/commands/upload_build_timings.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Get or create job record
+    job = Job.objects.filter(job_id=job_id).first()
+    if job is None:
+        job = create_job(gl_project, gl_job)
+
+    # Iterate through each timer and create timers and phase results
+    phases = []
+    for entry in get_timings_json(gl_job):
+        timer = Timer.objects.create(
```
scottwittenburg, line 84: Based on what we're seeing in actual aggregated timings files, it seems maybe bootstrapping something could be picking up old timings files with a different format.  So I think here if we have no `name` field in the `entry`, we need to skip this `Timer` and all its `Phases`.

analytics/analytics/management/commands/upload_build_timings.py
```diff
@@ -0,0 +1,111 @@
...
+            name=entry["name"],
+            hash=entry["hash"],
```
kwryankrattiger, line 86: some timers are not going to have `hash` or `cache` since they are not related to building packages. We should get `entry.get("hash", None)` to set these to null

AlmightyYakob: This has been fixed.

analytics/analytics/management/commands/upload_build_timings.py
AlmightyYakob: This has been done

.github/workflows/analytics_migrations.yml
```diff
@@ -0,0 +1,33 @@
+name: Check and Run Analytics Migrations
```
mvandenburgh, line 1: Does this workflow actually work? It seems like the DB is inside the private VPC, unless I'm missing something.

analytics/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+
+## Accessing Data
+
+To access and/or modify any data within the database, the django shell can be used as follows:
+```
+./manage.py shell_plus
+```
+
+This command will drop you into a Python shell with Django pre-loaded, so that models and data can be viewed, created, deleted, etc.
```
mvandenburgh, line 40: I think mentioning the `dbshell` command would also be useful here.

.github/workflows/analytics_migrations.yml
AlmightyYakob: You are correct. I spent some time coming up with a way around this, but have since backed it out of this PR, to avoid more unnecessary complexity and delays. I will address automatically running migrations in a follow up PR.

AlmightyYakob:
Okay, I've backed out all of the changes involving running migrations from the github workflow (will address in a follow up), and have just retained the very simple workflow of checking the migrations. So for the time being, if any migrations are made, they will need to be applied manually.

I've also backed out any changes that existed solely for testing, so this PR is ready to go.

scottwittenburg (approved):
This is working for me, so I'm approving :laughing: But it might make sense for @mvandenburgh to take a look from a kube/terraform perspective as well.

One question, well maybe with two parts: Will this just work when we merge Ryan's timing statistics PR and set up the webhook from `spack/spack`?  Will there be a way to keep triggering it and testing from `scott/pipeline-experiments` at that point as well?

AlmightyYakob:
> One question, well maybe with two parts: Will this just work when we merge Ryan's timing statistics PR and set up the webhook from `spack/spack`? Will there be a way to keep triggering it and testing from `scott/pipeline-experiments` at that point as well?

I believe the answer to both is yes. Flux should deploy the required services (might hiccup at first as the images will need to be built), and the webhook makes no distinction as to which project is invoking it.

mvandenburgh (approved):
I've been disconnected from the timing info stuff so I'll defer to Scott on the specifies of the DB schema and build processor server. But the Terraform for the database and k8s resources look good to me.

PR #597, opened 2023-08-16
Fix pipeline dashboard label selectors

AlmightyYakob:
This fixes a couple of panels in the "Pipeline Overview" dashboard, as they relied on a now out-dated label for nodes (removed in #446). 

The label selector has been changed to use a different node label, which accomplishes the same exact outcome, and has already been present since before the breaking change.


scottwittenburg (approved):
Thanks @AlmightyYakob, pasting those into the panel editor at least allows me to see data after 8/09.

PR #599, opened 2023-08-30
Add Spack S3 binary mirror resources to Terraform

mvandenburgh:
This PR adds a new Terraform module, `binary_mirror`, that represents a Spack binary mirror S3 bucket. This module is nested within the existing `spack` terraform module. It then creates two `binary_mirrors` in the `spack` module - a protected mirror, and a PR mirror. 

**Note**: I started off by just encoding the individual resources into the `spack` module, and then realized I could refactor them out to a module. I kept that initial work on the first commit to this PR (https://github.com/spack/spack-infrastructure/commit/5108bebad831985dcb03b69c38072e9ede45b041) in case it's helpful, but I think the module approach is cleaner.

These resources already exist in production, so we will need to run some `terraform import` statements to import those into our Terraform state; I've written a script to do this ([here](https://gist.github.com/mvandenburgh/8c105d7d49d121fd2d5d4dba65a04754)), and will run it after this is merged. I've already run `terraform apply` on staging to create staging versions of them.

For review, I would recommend looking at the new variables added to the `spack` module (`terraform/modules/spack/variables.tf`), then the two binary mirror instances (`terraform/modules/spack/binary_mirrors.tf`), and then the `binary_mirror` module itself (`terraform/modules/spack/modules/binary_mirror`).


terraform/staging/main.tf
```diff
@@ -133,4 +133,7 @@ module "staging_cluster" {
   provision_opensearch_cluster = false
 
   ses_email_domain = "staging.spack.io"
+
+  public_binary_mirrors       = false
```
danlamanna, line 137: Why is this configurable? And why is it false in staging?

terraform/staging/main.tf
```diff
@@ -133,4 +133,7 @@ module "staging_cluster" {
...
+  binary_mirror_bucket_suffix = "-staging"
```
AlmightyYakob, line 138: Could this not just be based on `deployment_name`, and avoid the variable altogether?

AlmightyYakob (changes requested):
One question/comment, but other than that nothing stands out

terraform/staging/main.tf
mvandenburgh: I figured there was no reason to allow public access as it's a staging bucket. If it's public, i think people could point their spack application at it and use it as a binary mirror. 
Maybe we don't care about that though, I don't have any strong opinions on it

terraform/staging/main.tf
mvandenburgh: Since the prod bucket already exists and is called `spack-binaries`, the alternative is a conditional like this https://github.com/spack/spack-infrastructure/blob/main/terraform/modules/spack/opensearch.tf#L2. My thinking was that an explicit input var is better than assigning some implicit special meaning to the `"prod"` deployment name string.

AlmightyYakob: Ultimately these modules are really a convenience for ourselves, and we already take some special care when it comes to the prod deployment. I think it's okay to handle the "prod" deployment in a special way like you highlighted, as that still generalizes to any more deployments we make, and I'd prefer to avoid adding superfluous variables to our terraform config.

But that's just my two cents. If you see your approach as less of a technical burden then I won't block on it.

terraform/staging/main.tf
danlamanna: I would vote for keeping prod/staging parity here since I think (hope?) we will be the people who point our spack installs at it. I think staging ought to be an identical-as-much-as-possible environment that just isn't very useful for non-developers.

terraform/staging/main.tf
mvandenburgh: That makes sense, I think it's fine to handle the "prod" deployment as a special case, as long as we're making an explicit decision to do that. And also, considering that I removed the other new variable based on Dan's suggestion above, it's nice to remove this one as well and avoid introducing any new vars in this PR.

terraform/staging/main.tf
mvandenburgh: Sounds good to me. I changed it

AlmightyYakob (approved):

mvandenburgh:
When I attempted to apply [`dd9c452` (#599)](https://github.com/spack/spack-infrastructure/pull/599/commits/dd9c4529e3aae56b05db8599f66a541972de6462) to staging, it failed due to these two resources missing [`e281f34` (#599)](https://github.com/spack/spack-infrastructure/pull/599/commits/e281f34b6867a7a6b0bf161ead8210ff6df23883), since S3 now disables ACLs by default. (Although I think we can remove the ACLs here since they do the same as the bucket policy, but the goal of this PR is to just encode what is already there.)

danlamanna (approved):

mvandenburgh:
I ran the script linked in the PR description, so all of these resources are imported into the production TF state now.

PR #601, opened 2023-09-01
Add gitlab terraform provider

danlamanna:


terraform/staging/main.tf
```diff
@@ -93,6 +97,10 @@ provider "github" {
   token = jsondecode(data.aws_secretsmanager_secret_version.flux_github_token.secret_string).flux_github_token
 }
 
+provider "gitlab" {
+  base_url = "https://gitlab.staging.spack.io"
+  token = jsondecode(data.aws_secretsmanager_secret_version.gitlab_token.secret_string).gitlab_terraform_provider_access_token
+}
```
mvandenburgh, line 103: I'm not suggesting that we need to do this since it might be an overcomplication, but I was playing around with this and found out you can actually extract the gitlab url from the cluster.
```suggestion
data "kubernetes_ingress_v1" "gitlab_webservice" {
  metadata {
    name      = "gitlab-webservice-default"
    namespace = "gitlab"
  }
}

locals {
  gitlab_url = "https://${data.kubernetes_ingress_v1.gitlab_webservice.spec[0].rule[0].host}"
}

provider "gitlab" {
  base_url = local.gitlab_url
  token = jsondecode(data.aws_secretsmanager_secret_version.gitlab_token.secret_string).gitlab_terraform_provider_access_token
}
```

mvandenburgh:
When I run `terraform init` locally on this branch, I get this error message
> Error: Failed to query available provider packages
	Could not retrieve the list of available versions for provider gitlabhq/gitlab: locked provider registry.terraform.io/gitlabhq/gitlab 16.2.0 does not match configured version
	constraint 16.3.0; must use terraform init -upgrade to allow selection of new versions

It looks like the version of the gitlab listed in the TF lockfile is 16.2.0, but you're specifying 16.3.0 here. I ran `terraform init -upgrade` and pushed the change to your branch.

mvandenburgh (approved):
If you're good with the lockfile changes, this LGTM

PR #602, opened 2023-09-05
Add GitLab OIDC provider

danlamanna:
Requires #601

This adds a GitLab OIDC provider to the spack module.

It also creates multiple resources for the protected and PR binary buckets*. The roles can only be assumed when the JWT matches the claims in `conditions`. This means that when a GitLab ID Token with `aud: protected_binary_mirror` is provided, it must also match the conditions here to be granted access.

This is a good illustration of how the overall flow works: https://docs.gitlab.com/ee/ci/cloud_services/index.html#authorization-workflow.

*The production bucket policies/roles aren't yet modularized in Terraform, so there's some logic to use the existing ones for production and create new ones for staging.

### DIFF
--- a/terraform/modules/spack/gitlab_runner_iam.tf
+++ b/terraform/modules/spack/gitlab_runner_iam.tf
@@ -1,0 +1,124 @@
+locals {
+  gitlab_domain = "gitlab${var.deployment_name == "prod" ? "" : ".staging"}.spack.io"
+
+  mirror_roles = {
+    "pr_binary_mirror" = {
+      "role_name_suffix"     = "PRBinaryMirror${var.deployment_name == "prod" ? "" : "-${var.deployment_name}"}",
+      "role_arn_ci_var_name" = "PR_BINARY_MIRROR_ROLE_ARN",
+      "conditions"           = ["project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:branch:ref:pr*"],
+    },
+    "protected_binary_mirror" = {
+      "role_name_suffix"     = "ProtectedBinaryMirror${var.deployment_name == "prod" ? "" : "-${var.deployment_name}"}",
+      "role_arn_ci_var_name" = "PROTECTED_BINARY_MIRROR_ROLE_ARN",
+      "conditions" = [
+        "project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:branch:ref:develop",
+        "project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:branch:ref:releases/v*",
+        "project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:tag:ref:develop-*",
+        "project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:tag:ref:v*"
+      ],
+    }
+  }
+}
+
+data "aws_caller_identity" "current" {}
+
+data "gitlab_project" "spack" {
+  path_with_namespace = "spack/spack"
+}
+
+data "tls_certificate" "gitlab" {
+  url = "https://${local.gitlab_domain}"
+}
+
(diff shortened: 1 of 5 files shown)


terraform/modules/spack/gitlab_runner_iam.tf
```diff
@@ -0,0 +1,125 @@
+locals {
+  # TODO: is it possible to get this from the gitlab provider?
+  gitlab_domain = "gitlab${var.deployment_name == "prod" ? "" : ".staging"}.spack.io"
+
+  mirror_roles = {
```
danlamanna, line 5: The conditions here should match these: https://github.com/spack/spack/blob/aa9eb331080a3edeb876dd80552bb59243c69783/share/spack/gitlab/cloud_pipelines/.gitlab-ci.yml#L71-L109 using the more limited IAM string matching. @scottwittenburg should probably also review these.

terraform/modules/spack/gitlab_runner_iam.tf
```diff
@@ -0,0 +1,125 @@
...
+resource "aws_iam_role" "gitlab_runner" {
+  for_each = data.aws_iam_policy_document.gitlab_oidc_assume_role
+
+  name                 = "GitLabRunner${local.mirror_roles[each.key].role_name_suffix}"
+  assume_role_policy   = each.value.json
+  max_session_duration = 3600 * 6 # only allow a max of 6 hours for a session to be active
+}
+
+data "aws_iam_policy_document" "gitlab_runner" {
+  for_each = var.deployment_name == "staging" ? local.mirror_roles : {}
```
danlamanna, line 75: @mvandenburgh This conditional and the 2 policy attachments can go away once the existing policies go into the spack module.

mvandenburgh: Why do the existing policies need to be preserved? Couldn't we just use the new `WriteBinariesTo` policy in prod as well?

danlamanna: I can but I would also have to update the currently in use IAM users to use the same policies. Those users/policies are going to be deleted once we start using this everywhere.

terraform/modules/spack/gitlab_runner_iam.tf
```diff
@@ -0,0 +1,125 @@
...
+      ],
+    }
+  }
+}
+
+data "aws_caller_identity" "current" {}
+
+data "gitlab_project" "spack" {
+  path_with_namespace = "spack/spack"
+}
```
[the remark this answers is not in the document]
danlamanna: Yeah, I'll make a follow up PR for making them resources and importing them.

terraform/modules/spack/gitlab_runner_iam.tf
```diff
@@ -0,0 +1,124 @@
+locals {
+  gitlab_domain = "gitlab${var.deployment_name == "prod" ? "" : ".staging"}.spack.io"
+
+  mirror_roles = {
+    "pr_binary_mirror" = {
+      "role_name_suffix"     = "PRBinaryMirror${var.deployment_name == "prod" ? "" : "-${var.deployment_name}"}",
+      "role_arn_ci_var_name" = "PR_BINARY_MIRROR_ROLE_ARN",
+      "conditions"           = ["project_path:${data.gitlab_project.spack.path_with_namespace}:ref_type:branch:ref:pr*"],
```
scottwittenburg, line 8: The rule we use in gitlab is a bit more restrictive about the name of pr branches, but I assume slightly fuller-featured regexes aren't supported here?

danlamanna: Correct, unfortunately https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_String is what we're limited to.

PR #605, opened 2023-09-06
Add a basic buildcache pruning application

scottwittenburg:
Similar to #415 but developed independently.  Currently used to cross-check pruning lists generated by that PR.


scottwittenburg:
fyi @kwryankrattiger

PR #606, opened 2023-09-07
Move cache indexer resources to spack module

danlamanna:
This moves us closer towards having staging/prod parity by moving the cache_indexer into the spack terraform module. It also parameterizes the cron job so it can run in both environments.

### DIFF
--- a/terraform/production/service_account_roles.tf
+++ b/terraform/production/service_account_roles.tf
@@ -77,72 +77,19 @@ resource "kubectl_manifest" "notary_service_account" {
   ]
 }
 
-# IAM Role for granting read access to spack-binaries and write access to the cache index
-data "aws_iam_policy_document" "cache_indexer_assume_role_policy" {
-  statement {
-    sid     = ""
-    effect  = "Allow"
-    actions = ["sts:AssumeRoleWithWebIdentity"]
-
-    condition {
-      test     = "StringEquals"
-      variable = "${module.production_cluster.oidc_provider}:aud"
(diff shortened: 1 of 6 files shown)


terraform/production/service_account_roles.tf
```diff
@@ -77,72 +77,19 @@ resource "kubectl_manifest" "notary_service_account" {
...
-    effect  = "Allow"
-    actions = ["s3:PutObject", "s3:DeleteObject"]
-
-    resources = [
-      "arn:aws:s3:::spack-binaries/cache_spack_io_index.json"
-    ]
-  }
+moved {
+  from = aws_iam_role.cache_indexer
+  to   = module.production_cluster.aws_iam_role.cache_indexer
```
mvandenburgh, line 82: How are you going to handle these moved blocks? Merge this PR, run `terraform apply`, and then make a follow up PR removing them?

mvandenburgh (approved):

terraform/production/service_account_roles.tf
danlamanna: Yes

mvandenburgh: :+1:

danlamanna:
@mvandenburgh @scottwittenburg I haven't been following #618 and the related PRs. Is this still safe to merge? It should create a new image for the cache-indexer.

scottwittenburg:
@danlamanna I think this should be safe to merge.  Go ahead, if it's ready, and I'll rebase mine if necessary.

PR #612, opened 2023-09-08
Refactor pre_build.py to avoid third party packages

danlamanna:


scottwittenburg (approved):
I'm not familiar with the internals you've had to dig into for this, but I appreciate the gymnastics you've done to avoid having to pip install anything for the pre_build script.  Thanks @danlamanna!

PR #613, opened 2023-09-08
Add dependency-free pre_build_script runner configuration

danlamanna:
This is the rewritten runner configuration that should run on any machine with python3.


scottwittenburg (approved):
Eventually we'll stop using that 2yr old ubuntu image that had problems with pip, but this looks like it should work everywhere we have python3, thanks!

PR #614, opened 2023-09-08
Switch pre_build.py to use JSON instead of XML

danlamanna:


scottwittenburg (approved):
No xml parsing using regex then?  In some ways I was looking forward to that, but nice that aws respects the `Accept` header.

PR #616, opened 2023-09-12
Upgrade to Gitlab 16.3.2

AlmightyYakob:
While performing this upgrade, we realized that the gitlab-runner helm chart images were pinned, which was causing the out-of-date runner versions we were seeing. 

@mvandenburgh According to the docs [here](https://gitlab.com/gitlab-org/charts/gitlab-runner/-/blob/main/values.yaml?ref_type=heads#L3-4), the pin is unnecessary. Is that consistent with why the change was made in the first place?

### DIFF
--- a/k8s/production/gitlab/release.yaml
+++ b/k8s/production/gitlab/release.yaml
@@ -19,7 +19,7 @@ spec:
   chart:
     spec:
       chart: gitlab
-      version: 7.3.0 # gitlab@16.3.0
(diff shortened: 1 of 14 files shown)


mvandenburgh (approved):
Oh, yeah we shouldn't specify that explicitly if it defaults to that.

mvandenburgh (approved):
Oh, yeah we shouldn't specify that explicitly if it defaults to that.

PR #625, opened 2023-09-15
Add vscode config and editorconfig

danlamanna:
@AlmightyYakob @mvandenburgh I know you're both vscode users, mind taking a look?


.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+    "recommendations": [
+        "editorconfig.editorconfig",
+        "hashicorp.terraform",
+        "ms-python.black-formatter",
+        "matangover.mypy",
```
AlmightyYakob, line 6: Is this needed if pylance is already recommended?

AlmightyYakob (changes requested):
One question but otherwise looks good

.vscode/extensions.json
danlamanna: No I don't think so. I'll remove it.

PR #626, opened 2023-09-15
add additional job labels to all build job pods

cmelone:
This change adds the following information to the job metadata:
- package version
- compiler name
- compiler version
- arch
- package variants

These labels were added to the environment in https://github.com/spack/spack/pull/39905 (pending merge).

let me know if i'm missing anything

cc: @alecbcs @scottwittenburg

### DIFF
--- a/k8s/production/runners/signing/release.yaml
+++ b/k8s/production/runners/signing/release.yaml
@@ -151,11 +151,6 @@ spec:
               "metrics/gitlab_ci_project_name" = "$CI_PROJECT_NAME"
               "metrics/gitlab_ci_job_stage" = "$CI_JOB_STAGE"
               "metrics/gitlab_ci_commit_ref_name" = "$CI_COMMIT_REF_NAME"
-              "metrics/spack_job_spec_pkg_version" = "$SPACK_JOB_SPEC_PKG_VERSION"
-              "metrics/spack_job_spec_compiler_name" = "$SPACK_JOB_SPEC_COMPILER_NAME"
-              "metrics/spack_job_spec_compiler_version" = "$SPACK_JOB_SPEC_COMPILER_VERSION"
-              "metrics/spack_job_spec_arch" = "$SPACK_JOB_SPEC_ARCH"
-              "metrics/spack_job_spec_variants" = "$SPACK_JOB_SPEC_VARIANTS"
               "metrics/spack_ci_stack_name" = "$SPACK_CI_STACK_NAME"
             [runners.kubernetes.node_selector]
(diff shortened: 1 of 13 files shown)


k8s/production/runners/signing/release.yaml
```diff
@@ -151,6 +151,11 @@ spec:
               "metrics/gitlab_ci_project_name" = "$CI_PROJECT_NAME"
               "metrics/gitlab_ci_job_stage" = "$CI_JOB_STAGE"
               "metrics/gitlab_ci_commit_ref_name" = "$CI_COMMIT_REF_NAME"
+              "metrics/spack_job_spec_pkg_version" = "$SPACK_JOB_SPEC_PKG_VERSION"
+              "metrics/spack_job_spec_compiler_name" = "$SPACK_JOB_SPEC_COMPILER_NAME"
+              "metrics/spack_job_spec_compiler_version" = "$SPACK_JOB_SPEC_COMPILER_VERSION"
+              "metrics/spack_job_spec_arch" = "$SPACK_JOB_SPEC_ARCH"
+              "metrics/spack_job_spec_variants" = "$SPACK_JOB_SPEC_VARIANTS"
```
scottwittenburg, line 158: These variables aren't applicable to this particular job, it runs at the end of each child pipeline and operates on all the specs in the mirror.

scottwittenburg (changes requested):
I requested to remove these labels from the `signing` runner, but otherwise this looks good to me, thanks @cmelone!

scottwittenburg (approved):

cmelone:
Hmm, this might be a problem for parsing. These variants: 

```
~bignuma~consistent_fpcsr+fortran~ilp64+locking+pic+shared
      build_system=makefile symbol_suffix=none threads=openmp
```

get displayed as this in prometheus:

```
bignuma_consistent_fpcsr_fortran_ilp64_locking_pic_shared_build
```

I think the main issue is that there is no way to distinguish between `~` and `+` when it gets converted to the underscore format.

scottwittenburg:
> but this adds an additional request per job to my collection script, slowing down the execution quite a bit

This also adds to the load on gitlab, which can already struggle to handle requests at busy times (like early this morning).

> Not sure if this is a limitation inherent to Prometheus, or if this is something I can help resolve.

I'm not at all a prometheus expert, if you can find a way to avoid the mangling of strings, that would be great.

We could also consider if some kind of encoding (eg base64 or url) could help here.  I'm just not sure what characters or formats prometheus can't handle, forcing it to add the underscores.

PR #627, opened 2023-09-15
Validate OIDC token audience in pre_build.py

danlamanna:


scottwittenburg (approved):
Looks good, thanks @danlamanna!

PR #628, opened 2023-09-18
Upgrade cdash.spack.io to v3.2.0-rc1

zackgalbreath:


zackgalbreath:
@mvandenburgh let me know if anything about this diff makes you nervous.

I used a very similar approach a while back to upgrade cdash-staging.spack.io and it went pretty smoothly.

mvandenburgh (approved):
Obviously I can't speak to the specifies of the cdash changes, but from an infrastructure/k8s standpoint nothing looks out of place here :+1:

PR #630, opened 2023-09-19
Move some job vars over to annotations rather than labels

cmelone:
https://github.com/spack/spack-infrastructure/pull/626 attempted to add some info to the pod labels, but there are limitations that wouldn't allow full specs to be displayed ([under 64 chars, only some tokens allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)). 

This change places that information under [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/), the more appropriate vessel for this data.

cc: @alecbcs

@scottwittenburg I'm not sure who the right person to assign for review is.


scottwittenburg:
@cmelone This seems reasonable to me, though you're right, I'm probably not the best person to review.  

Maybe @AlmightyYakob or @mvandenburgh can take a look.  I think @danlamanna probably knows more than I do on this topic as well.

cmelone:
Just following up, thanks

@danlamanna

cmelone:
Putting this on hold for now 

see #635

cmelone:
Since the main goal of this PR is to reduce the load on the Gitlab API and shift it over to Prometheus, I'm wondering if it's architecturally possible to include the following fields: job status (running, fail, success), start time, end time/duration. Would these fields be difficult to include in labels/annotations since they can change during the course of the build?

k8s/production/prometheus/release.yaml
```diff
@@ -139,6 +139,8 @@ spec:
       metricLabelsAllowlist:
         - pods=[*]
         - nodes=[*]
+      metricAnnotationsAllowList:
+        - pods=[gitlab/ci_job_id,metrics*]
```
AlmightyYakob, line 143: Is there any specific reason `gitlab/ci_job_id` is exposed as the only "gitlab" annotation?

AlmightyYakob (approved):
Have one question but otherwise looks good to me. I checked and it doesn't seem like we use the removed labels anywhere in our dashboards, or otherwise. The removed labels are technically used in https://github.com/spack/spack-infrastructure/pull/697, but since they're being moved to annotations, I can just as easily retrieve that data from the pod annotations.

k8s/production/prometheus/release.yaml
cmelone: I did that so I could query the annotation by the job ID, and the other gitlab variables are more about the runner itself rather than the job imo

PR #631, opened 2023-09-20
Add dashboard tracking pressure on selected  groups of runners

scottwittenburg:
Adds a dashboard with panels attempting to show pressure on selected groups of runners over time.  The queries are addressed to the gitlab database, and find jobs that ran on UO runners using three different `WHERE` clauses:

1. darwin/arm64: `ci_runner.platform LIKE 'darwin' AND ci_runner.architecture LIKE 'arm64'`
2. ppc64le: `ci_runners.architecture LIKE 'ppc64le'`
3. cray:  `ci_runners.description LIKE 'uo-gary%'`
4. aws: `ci_runners.description LIKE '%-gitlab-runner-%'`

Only jobs with a non-null `started_at` are selected, and results are binned up by that value into 5 minute intervals.  Average seconds in pending state is computed for the jobs in each bin and plotted as well.

uo-runner-pressure

### DIFF
--- a/k8s/production/prometheus/custom/runner-pressure.yaml
+++ b/k8s/production/prometheus/custom/runner-pressure.yaml
@@ -1,0 +1,860 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: monitoring
+  name: kube-prometheus-stack-runner-pressure-dashboard
+  labels:
+    grafana_dashboard: "1"
+    app: kube-prometheus-stack-grafana
+    release: "kube-prometheus-stack"
+data:
+  gitlab-runner-pressure-dashboard.json: |-
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": {
+              "type": "grafana",
+              "uid": "-- Grafana --"
+            },
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@AlmightyYakob In [`fc3e8ad`](https://github.com/spack/spack-infrastructure/commit/fc3e8ad942ffd8011653c3c4d320a4fe8627e04c) I incorporated your improved query (timeseries generation).  Thanks!

You can see the live charts [here](https://grafana.spack.io/d/yS3bBliIz/uo-runner-pressure?orgId=1&from=now-24h&to=now).

scottwittenburg:
@AlmightyYakob Unless you object, let's merge this as-is.  Then if you want, you can iterate a bit on the queries in a follow-on PR.  I've updated it (and grafana), so they currently match, in case you want to poke at it live before you approve.

scottwittenburg:
Maybe we can eventually figure out to run the spack-ci-bridge test only when we change that thing, instead of every time we touch anything in this repo.  It looks to me like the tests actually passed, but then it maybe failed because it was taking too long to communicate success to cdash?  Not sure, maybe @zackgalbreath has an idea.

AlmightyYakob:
> @AlmightyYakob Unless you object, let's merge this as-is. Then if you want, you can iterate a bit on the queries in a follow-on PR. I've updated it (and grafana), so they currently match, in case you want to poke at it live before you approve.

No objections 👍

AlmightyYakob (approved):

PR #632, opened 2023-09-21
Install Sentry on production

mvandenburgh:
This PR installs self-hosted sentry on production. This consists of

1) Moving the sentry `HelmRelease` and related resources into the `k8s/production` directory, and refactoring the existing staging configuration to be a ` Kustomization`
2) Creating a k8s `Job` to automatically create an initial user for Sentry + an API key
3) Feeding that API key into the sentry Terraform provider, and using that to create resources on sentry


terraform/modules/spack/sentry.tf
```diff
@@ -0,0 +1,207 @@
...
+        node_selector = {
+          "spack.io/node-pool" = "base"
+        }
+      }
+    }
+  }
+  # Setting this to true forces Terraform to wait for the API
+  # key to be created before creating any sentry resources.
+  wait_for_completion = true
+}
```
mvandenburgh, line 125: Any thoughts on this approach @AlmightyYakob @danlamanna ? This lets us avoid the chicken-and-egg problem that comes with using the Sentry terraform provider. For reference, I got the Python code from https://github.com/getsentry/sentry/issues/46911

danlamanna: Works for me 🤷‍♂️

danlamanna (approved):

terraform/modules/spack/sentry.tf
AlmightyYakob: >  This lets us avoid the chicken-and-egg problem that comes with using the Sentry terraform provider.

What is the exact chicken-and-egg problem? Is it that you need to have users or api tokens created before you can use sentry, yet we're trying to spin up sentry from scratch?

mvandenburgh: The former.

mvandenburgh: Well, to be more specific - you need a user/API key in order to configure Sentry with the Terraform provider. But to create an API key, you need to create a user on Sentry. Ideally, Terraform takes care of provisioning the API key, and then feeds that into the sentry provider.

terraform/modules/spack/gitlab_ses.tf
```diff
@@ -0,0 +1,42 @@
...
+    kind: ConfigMap
+    metadata:
+      name: gitlab-ses-config
+      namespace: gitlab
+    data:
+      values.yaml: |
+        global:
+          email:
+            from: admin@${local.gitlab_email_domain}
+            reply_to: noreplay@${local.gitlab_email_domain}
```
AlmightyYakob, line 19: Is this a typo?
```suggestion
            reply_to: noreply@${local.gitlab_email_domain}
```

AlmightyYakob (approved):
I think this looks good (aside from my one comment), although I'll just lament the fact that we have a chicken and egg problem between sentry and terraform. Maybe in the future we could revisit this, and somehow separate out the two distinct steps here. Namely:

1. Run terraform initially to bring all necessary infra up, which will in turn run flux and apply all of the k8s files
2. Run a separate terraform command after the first one is run, that is responsible for the configuration of existing resources.

This would apply for any resources which have this kind of chicken-and-egg problem. If we documented this somewhere (maybe a terraform README), the main downside would be running terraform in two commands instead of one.

Idk, just a thought for the future :man_shrugging:

terraform/modules/spack/gitlab_ses.tf
mvandenburgh: This typo wasn't introduced by this PR, so I'll address this in a follow up

mvandenburgh:
> I think this looks good (aside from my one comment), although I'll just lament the fact that we have a chicken and egg problem between sentry and terraform. Maybe in the future we could revisit this, and somehow separate out the two distinct steps here. Namely:
> 
> 1. Run terraform initially to bring all necessary infra up, which will in turn run flux and apply all of the k8s files
> 2. Run a separate terraform command after the first one is run, that is responsible for the configuration of existing resources.
> 
> This would apply for any resources which have this kind of chicken-and-egg problem. If we documented this somewhere (maybe a terraform README), the main downside would be running terraform in two commands instead of one.
> 
> Idk, just a thought for the future 🤷‍♂️

I agree, something like this would be nice and is probably a prerequisite to getting to a place where we can spin up the entire infrastructure from scratch with terraform.

PR #633, opened 2023-09-21
gitlab-error-processor: Process successful jobs as well as failures

scottwittenburg:
Trigger the upload_failure_logs module on successes as well as failures, allowing it to notice and upload records about jobs that were incorrectly scheduled, i.e. jobs which ultimately found there was nothing to rebuild.

### DIFF
--- a/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
+++ b/images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
@@ -19,7 +19,7 @@ config.load_config()
 v1_client = client.CoreV1Api()
 
 
-class JobPayload(Document):
+class FailedJobPayload(Document):
     timestamp = Date()
 
     class Index:
@@ -35,6 +35,22 @@ class JobPayload(Document):
         return super().save(**kwargs)
 
 
+class UnnecessaryJobPayload(Document):
+    timestamp = Date()
+
+    class Index:
+        name = "gitlab-unnecessary-jobs-*"
+
+    def save(self, **kwargs):
+        # assign now if no timestamp given
+        if not self.timestamp:
+            self.timestamp = datetime.now()
+
+        # override the index to go to the proper timeslot
+        kwargs["index"] = self.timestamp.strftime("gitlab-unnecessary-jobs-%Y%m%d")
+        return super().save(**kwargs)
+
+
 GITLAB_TOKEN = os.environ["GITLAB_TOKEN"]
 GITLAB_POSTGRES_DB = os.environ["GITLAB_POSTGRES_DB"]
 GITLAB_POSTGRES_USER = os.environ["GITLAB_POSTGRES_RO_USER"]
@@ -44,6 +60,8 @@ GITLAB_POSTGRES_HOST = os.environ["GITLAB_POSTGRES_HOST"]
 OPENSEARCH_ENDPOINT = os.environ["OPENSEARCH_ENDPOINT"]
 OPENSEARCH_USERNAME = os.environ["OPENSEARCH_USERNAME"]
 OPENSEARCH_PASSWORD = os.environ["OPENSEARCH_PASSWORD"]
+
+UNNECESSARY_JOB_REGEX = re.compile(r"No need to rebuild [^,]+, found hash match")
 
 
 # Instantiate gitlab api wrapper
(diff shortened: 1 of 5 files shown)


images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
```diff
@@ -35,6 +35,22 @@ def save(self, **kwargs):
...
+        name = "gitlab-unnecessary-jobs-*"
+
+    def save(self, **kwargs):
+        # assign now if no timestamp given
+        if not self.timestamp:
+            self.timestamp = datetime.now()
+
+        # override the index to go to the proper timeslot
+        kwargs["index"] = self.timestamp.strftime("gitlab-unnecessary-jobs-%Y%m%d")
+        return super().save(**kwargs)
```
scottwittenburg, line 51: @mvandenburgh I'm not thrilled about having to duplicate most of this class, but I found little or no documentation on how the `Document` class is supposed to be used.  It would be nice to just be be able to change the index patterns since everything else is the same, but I'm not sure what bits of the class are used under the hood and how.  Some questions you may or may not know the answers to:

1. Why do we have the `timestamp` class member?  Do we want all instances to have the same value?
2. Is it intentional that there is also an instance member of the same name, but seemingly a different type?  One is a `opensearch_dsl.Date`, one is a `datetime` object.
3. How is the inner `Index` class used by the parent class?

scottwittenburg:
Also @mvandenburgh I have a question about how to test this change in staging.  I *think* what could work is:

- keep the change bumping the versions of the images we build on push to `main`
- remove the changes that use the new image versions
- add two new directories under `k8s/staging/custom/` for the error processor and log uploader
- in those, add `kustomization.yaml` files to pick the new image versions
- merge all the above into `main`

Does that seem like it would work?

mvandenburgh:
> Also @mvandenburgh I have a question about how to test this change in staging. I _think_ what could work is:
> 
> * keep the change bumping the versions of the images we build on push to `main`
> * remove the changes that use the new image versions
> * add two new directories under `k8s/staging/custom/` for the error processor and log uploader
> * in those, add `kustomization.yaml` files to pick the new image versions
> * merge all the above into `main`
> 
> Does that seem like it would work?

Yup, that's exactly how I would do it.

AlmightyYakob:
I believe we discussed this a bit in our meeting, but considering a task I had for the build time processor to include failed jobs instead of only successful jobs, we'd be essentially saving the same data in two locations (three if you include the gitlab DB). I wonder what the intent for using this data is, and if it might make more sense to store it in the analytics DB instead?

scottwittenburg:
> I wonder what the intent for using this data is, 

The intent here is just to track rebuild jobs we schedule erroneously.  For some background: this is most likely because the buildcache index was out of date (typically it might not reflect recently-built specs), and when we generate a pipeline we tell it to only look at the index.  Then when the scheduled job runs, there is a separate check that looks directly for the target spec, and if it finds one, it prints "Nothing to rebuild..." and exits.  I haven't come up with another way to catch those other than by scraping the traces.

> and if it might make more sense to store it in the analytics DB instead

Yeah, it very well might be better to do that.  Should I add connection details for the analytics DB to this PR and push the data there, or should I move this stuff into the analytics DB processor?

cmelone:
This looks great! +1 for adding it to the analytics database.

images/upload-gitlab-failure-logs/upload_gitlab_failure_logs.py
mvandenburgh: 1/2) No, we don't want all instances to have the same value. I'm not sure why I put that there, maybe it was an oversight from a copy paste from an example like this https://elasticsearch-dsl.readthedocs.io/en/latest/#persistence-example. 

3) I believe the API of `opensearch_dsl` is the same as `elasticsearch_dsl`. So, the docs for the latter should help here - https://elasticsearch-dsl.readthedocs.io/en/latest/api.html#index

AlmightyYakob:
> Yeah, it very well might be better to do that. Should I add connection details for the analytics DB to this PR and push the data there, or should I move this stuff into the analytics DB processor?

If we decide to go that route, I think moving it into the analytics db processor would probably work better. We would probably want to process both in the same file, and rename the `upload_build_timings` management command to something more general, like `upload_build_data` or something (someone else come up with a better name please).

danlamanna:
This info is now being stored in the jobs table with the `unnecessary` column (see #747). Also opensearch can show this info for recent jobs by searching `gitlab-job-logs-*` for `"No need to rebuild"`.

PR #635, opened 2023-09-22
Add new fields to analytics database

cmelone:
I took a quick pass at adding some additional information about each job to the analytics db, in the interest of having these fields in a relational format, rather than just in Prometheus. 

cc: @scottwittenburg


scottwittenburg:
Thanks for taking a stab at this @cmelone!  Unless you object, one or more of us may make our own pass on this PR as well.  For example, I'd like to try to consolidate the download/extract of artifacts to happen in one place for both existing purposes.  And maybe @AlmightyYakob will have thoughts on table structure/layout.  In general though, it looks like you covered all the important bits that are needed.  Thanks again!

cmelone:
sounds good 👍

AlmightyYakob:
I'll take a look at this as soon as I can.

analytics/analytics/management/commands/upload_build_timings.py
```diff
@@ -22,11 +20,48 @@
...
+                            "SPACK_JOB_SPEC_COMPILER_VERSION"
+                        ),
+                        "arch": job_vars.get("SPACK_JOB_SPEC_ARCH"),
+                        "package_variants": job_vars.get("SPACK_JOB_SPEC_VARIANTS"),
+                        "build_jobs": job_vars.get("SPACK_BUILD_JOBS"),
+                        "job_size": job_vars.get("CI_JOB_SIZE"),
+                        "stack": pipeline_vars.get("SPACK_CI_STACK_NAME"),
+                    }
+        except KeyError:
+            pass
```
AlmightyYakob, line 56: Where in the try block is this catching against? Is it the `zfile.open`?

analytics/analytics/management/commands/upload_build_timings.py
```diff
@@ -22,11 +20,48 @@
...
+                        "package_version": job_vars.get("SPACK_JOB_SPEC_PKG_VERSION"),
+                        "compiler_name": job_vars.get("SPACK_JOB_SPEC_COMPILER_NAME"),
+                        "compiler_version": job_vars.get(
+                            "SPACK_JOB_SPEC_COMPILER_VERSION"
+                        ),
+                        "arch": job_vars.get("SPACK_JOB_SPEC_ARCH"),
+                        "package_variants": job_vars.get("SPACK_JOB_SPEC_VARIANTS"),
+                        "build_jobs": job_vars.get("SPACK_BUILD_JOBS"),
+                        "job_size": job_vars.get("CI_JOB_SIZE"),
+                        "stack": pipeline_vars.get("SPACK_CI_STACK_NAME"),
```
AlmightyYakob, line 53: It seems most of these are always present, is that correct?

analytics/analytics/models.py
```diff
@@ -14,6 +14,15 @@ class Job(models.Model):
 
     # Derived fields
     package_name = models.CharField(max_length=128)
+    # these fields can be null because the data was not previously collected
+    package_version = models.CharField(max_length=128, null=True)
+    compiler_name = models.CharField(max_length=128, null=True)
+    compiler_version = models.CharField(max_length=128, null=True)
+    arch = models.CharField(max_length=128, null=True)
+    package_variants = models.TextField(null=True)
```
AlmightyYakob, line 22: What's the reason for making this a text field? Does it just have a truly arbitrary length?

cmelone:
just wondering if there is any update? @AlmightyYakob

AlmightyYakob:
> just wondering if there is any update? @AlmightyYakob

I've left questions for you on this PR that I'm waiting for a response on.

cmelone:
> > just wondering if there is any update? @AlmightyYakob
> 
> I've left questions for you on this PR that I'm waiting for a response on.

Sorry, I'm not used to reviewing on github...I had no idea that "pending" meant you couldn't see them 🤦🏻‍♂️

AlmightyYakob:
> Sorry, I'm not used to reviewing on github...I had no idea that "pending" meant you couldn't see them 🤦🏻‍♂️

No worries! I've done the same thing before.

AlmightyYakob:
@cmelone I pushed a commit that reorganizes/refactors things a bit. I think it's good the way it is, but I'd like someone like @scottwittenburg or @mvandenburgh to sanity check me. If they think it looks good then I think it's good to go.

analytics/analytics/management/commands/upload_build_timings.py
```diff
@@ -36,26 +76,19 @@ def create_job(project: Project, job: ProjectJob) -> Job:
...
-            with zipfile.ZipFile(artifacts_file) as zfile:
-                with zfile.open(timing_filename) as timing_file:
-                    return json.load(timing_file)
-        except KeyError:
-            pass
+    try:
+        with get_job_artifacts_file(job, timing_filename) as file:
+            return json.load(file)
+    except KeyError:
+        pass
```
mvandenburgh, line 91: Under what circumstance would a `KeyError` be thrown here? I see there are places where a regular `Exception` gets thrown, but I can't find where a `KeyError` would potentially happen.
Also, what would be the reason that such an error gets thrown? i.e. is there a legitimate scenario where a `KeyError` would happen, or would it only happen if data is somehow malformed? Just trying to figure out if silently suppressing this error rather than letting sentry catch it is the way to go here or not.

AlmightyYakob: `KeyError` would be thrown if the filename that's attempted to be read from the zip file archive is not present. If we feel confident that a timings file should pretty much _always_ exist in a job's artifacts, then we could remove the try/except. Although I expect that some number of jobs will either not have artifacts, or won't have the timings file (e.g. if something in the job fails unexpectedly).

mvandenburgh: > If we feel confident that a timings file should pretty much always exist in a job's artifacts, then we could remove the try/except.

I'm not familiar with how the timings file gets generated and what would cause it to not be generated, so I can't speak confidently either way. But it seems like having some degree of visibility into files that fail would be good. Since we have sentry enabled, having an exception get thrown would raise an alert there and would at least give us some auditability regarding which jobs have missing timings files and the reasons for them being missing. And maybe down the line, if we determine that there's specific scenarios where a timings file is not expected to be present, we could add logic to handle those gracefully. 

But it seems like that's not where we are now (i.e., we don't have a solid understanding of all the different scenarios where a timings file would be legitimately missing), so I guess taking a "catch-all" approach where we suppress any errors arising from a missing file seems risky to me and could potentially cause legitimate issues to go unnoticed.

AlmightyYakob: @kwryankrattiger or @scottwittenburg might want to chime in here.

To clarify, the situation I'm trying to prevent is that a job fails, and due to that no timing data is uploaded. In that case, we'd potentially get a sentry error somewhere else in the code base, _as well as here_, even though the error here is really a red herring. The error being raised here can't actually help us at all, since it's not pointing us to the root cause (rather, it's a side effect of the real error).

We don't currently process failed jobs as a part of this webhook, but we've discussed the desire/need to do so (relevant [thread](https://github.com/spack/spack-infrastructure/pull/633#issuecomment-1730747929)), so that might be a real concern soon.

Having said all that, unless Ryan or Scott object, I'm happy to structure this so that an exception is raised to sentry. I just want to make sure we're aware of the potential noise / error duplication in sentry. If we do decide to handle failed jobs as well, we'll probably need to gracefully handle this error, at least against failed jobs.

AlmightyYakob: Relevant change: 03dd182

mvandenburgh: > To clarify, the situation I'm trying to prevent is that a job fails, and due to that no timing data is uploaded. 

In that case, maybe keeping the exception handler and doing something like this would make more sense? https://docs.sentry.io/platforms/python/usage/#capturing-errors

AlmightyYakob: At this point, I'd prefer to just get this merged in, and worry about the ramifications about processing failed jobs on the PR that introduces that functionality.

mvandenburgh: Sounds good to me.

AlmightyYakob (approved):

mvandenburgh (commented):
LGTM!

AlmightyYakob:
Hold off on merging, I just realized we need to bump the container image

AlmightyYakob (approved):

cmelone:
I'm not too familiar with Metabase, but how can I find out why the new columns have appeared, but no values are being propagated for new jobs?

https://metabase.spack.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjozLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjE5MTUsIm9yZGVyLWJ5IjpbWyJkZXNjIixbImZpZWxkIiwzMjg2MyxudWxsXV1dfSwidHlwZSI6InF1ZXJ5In0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9LCJvcmlnaW5hbF9jYXJkX2lkIjpudWxsfQ==

AlmightyYakob:
I'm taking a look into it

AlmightyYakob:
@cmelone I believe I've diagnosed the issue: https://github.com/spack/spack-infrastructure/pull/691

PR #644, opened 2023-09-29
Increase `maxReplicas` for ingress-nginx controller

mvandenburgh:
The nginx autoscaler is still scaling up the max number of pods, even after #640 and #641, so this PR doubles it again. After this is merged, we should keep an eye on the "NGINX Ingress Controller" grafana dashboard to see what effect it has.


mvandenburgh:
This is the wrong approach and will likely have no impact, we need to look at nginx resource requests first.

PR #647, opened 2023-10-05
Add retries to pre_build.py

danlamanna:
This adds retries to the OIDC assume role script in the event that a transient error occurs, or worse, gitlab.spack.io is down momentarily while AWS is fetching the identity information.

@scottwittenburg PTAL. In particular, I'm not sure how long we should be sleeping and retrying.

### DIFF
--- a/scripts/gitlab_runner_pre_build/pre_build.py
+++ b/scripts/gitlab_runner_pre_build/pre_build.py
@@ -6,7 +6,7 @@ sourced by the gitlab runner in its pre_build configuration option.
 In the case of a PR build, the temporary credentials are scoped down to only allow access to the
 S3 bucket prefix for the relevant PR.
 """
-import os, sys, json, base64
+import os, sys, json, base64, time
 import urllib.request, urllib.parse, urllib.error
 
 TEMPORARY_CREDENTIALS_DURATION = 3600 * 6  # 6 hours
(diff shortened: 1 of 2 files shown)


scripts/gitlab_runner_pre_build/pre_build.py
```diff
@@ -48,6 +48,42 @@ def _token_to_sts_request(raw_jwt, decoded_jwt):
...
+        except urllib.error.HTTPError as e:
+            print(e.read().decode("utf-8"), file=sys.stderr)
+
+            if e.code == 400:
+                attempts += 1
+                print(
+                    f"Assume role request failed with 400, retrying ({attempts}/{max_attempts})",
+                    file=sys.stderr,
+                )
+                time.sleep(2 ** (1 + attempts))
```
scottwittenburg, line 80: Assuming I'm reading this correctly, it allows around 2 minutes of "retry grace", assuming all attempts are used.  Is that long enough for the length of outtage we have been seeing?

scripts/gitlab_runner_pre_build/pre_build.py
```diff
@@ -48,6 +48,42 @@ def _token_to_sts_request(raw_jwt, decoded_jwt):
...
+                + urllib.parse.urlencode(assume_role_kwargs),
+                headers={"Accept": "application/json"},
+            )
+
+            with urllib.request.urlopen(req) as response:
+                if response.getcode() == 200:
+                    response = json.loads(response.read().decode("utf-8"))
+                    return response
+                else:
+                    raise ValueError(
```
scottwittenburg, line 68: Is this new check based on some failure mode you've seen recently?

scripts/gitlab_runner_pre_build/test_pre_build.py
```diff
@@ -97,3 +104,36 @@ def read(self):
...
+                "someurl", 400, "Internal Server Error", None, None
+            )
+        else:
+            return MockResponse(200)
+
+    flaky_urlopen.has_failed = False
+
+    mocker.patch("urllib.request.urlopen", side_effect=flaky_urlopen)
+    mocker.patch("urllib.request.Request")
+    mocker.patch("time.sleep", side_effect=lambda x: None)
```
scottwittenburg, line 123: Good idea :grin: No need to make the testing take longer than necessary.

scottwittenburg (dismissed):
Looks good to me.  As long as my questions don't inspire any code changes, go ahead and merge.

scripts/gitlab_runner_pre_build/pre_build.py
danlamanna: It's tough to say, because a lot of the outages we're seeing aren't 100% down, but very flaky responsiveness. What do you think? Is it worth sleeping for longer to potentially save a job?

scripts/gitlab_runner_pre_build/pre_build.py
danlamanna: It's based on being overlay paranoid. AWS doesn't enumerate the non-error statuses possible so I would rather explicitly catch it.

scripts/gitlab_runner_pre_build/pre_build.py
scottwittenburg: I'm not sure either.  We already try jobs three times, and once a job fails, there's already some delay before it is retried. Typically I see a failed job go into a paused state, and other jobs that have been queued run first before the next try.  If you added two more attempts using the same backoff computation, that would only be around 8.5 minutes of retrying, total.  What do you think about that?

danlamanna: I added one more attempt so we met in the middle 😅.

scottwittenburg: :ok_hand:

scottwittenburg (approved):

PR #660, opened 2023-10-12
Add gitlab-runner-pod-cleanup pod

AlmightyYakob:
@mvandenburgh found [this](https://gitlab.com/gitlab-org/ci-cd/gitlab-runner-pod-cleanup#gitlab-runner-pod-cleanup) project (created by the gitlab org itself), which is a service that can be run in your kubernetes cluster to automatically clean up pods that have overstayed their welcome. The original intent of that project is aligned with our own - to clean up orphaned runner pods (for us, pods that run builds in the `pipeline` namespace).

The service is configured in this PR to clean up any runner pods that are still running after 7 hours. Since our timeout in gitlab is 6 hours, this seemed like a natural  choice. I also configured the service to run in the `custom` namespace. I'm open to suggestions on either of these choices, if people have them.


AlmightyYakob:
I have tested this in staging by deploying the service, and manually creating a test pod with the appropriate annotation. I can confirm that it deletes the pod as expected. I was able to see the log message of the deleted pod in the pod running the cleanup service, and I also observed it get deleted.

k8s/production/custom/gitlab-pod-cleanup/pod.yaml
```diff
@@ -0,0 +1,49 @@
...
+kind: ServiceAccount
+metadata:
+  name: pod-cleanup-sa
+  namespace: custom
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gitlab-runner-pod-cleanup
+  namespace: custom
```
mvandenburgh, line 33: Should we have this in the `gitlab` namespace instead? It seems like we generally reserve `custom` for custom scripts/jobs, whereas this is a tool maintained by gitlab.

mvandenburgh (approved):
LGTM. Seems like this just does the exact same thing that @zackgalbreath was going to write a custom job for, but with less work.  Just left one question about the namespace but I don't feel strongly either way.

k8s/production/custom/gitlab-pod-cleanup/pod.yaml
AlmightyYakob: I was thinking something similar as well. I'll change it.

PR #661, opened 2023-10-12
Add sentry_dsn for gitlab runners

danlamanna:
This should catch any system errors on gitlab runners in sentry. @mvandenburgh It turns out this is a global config, not in the runners toml, so we can use a config map for it.


mvandenburgh (approved):
Cool! I missed that global option when i looked at this.

PR #665, opened 2023-10-13
Enable Prometheus metrics for Karpenter

mvandenburgh:
@AlmightyYakob do you remember why this was disabled? I suspect it wasn't working before due to the namespace issue mentioned in #664.


terraform/modules/spack/karpenter.tf
```diff
@@ -70,6 +70,12 @@ resource "helm_release" "karpenter" {
     value = "1Gi"
   }
 
+  # Enable service monitor for prometheus metrics
+  set {
+    name = "serviceMonitor.enabled"
+    value = true
+  }
```
AlmightyYakob, line 77: Have you tested this at all to confirm there's no other config required?

AlmightyYakob:
> @AlmightyYakob do you remember why this was disabled? I suspect it wasn't working before due to the namespace issue mentioned in https://github.com/spack/spack-infrastructure/pull/664.

I didn't put too much time into it, I just noticed that it wasn't configured correctly but wasn't sure how to fix it.

terraform/modules/spack/karpenter.tf
mvandenburgh: Yes, I did a terraform apply in staging and it works as expected.

AlmightyYakob (approved):

PR #675, opened 2023-10-19
Add Sentry integration to sync script

mvandenburgh:
Adds Sentry to the sync script for error reporting. I've parameterized things such that if (1) the sentry sdk is not installed, or (2) there's no SENTRY_DSN environment variable set, then the sync script still works as usual.

### DIFF
--- a/images/gh-gl-sync/SpackCIBridge.py
+++ b/images/gh-gl-sync/SpackCIBridge.py
@@ -15,6 +15,14 @@ import tempfile
 import urllib.parse
 import urllib.request
 
+if 'SENTRY_DSN' in os.environ:
+    import sentry_sdk
+
+    sentry_sdk.init(
+        dsn=os.environ['SENTRY_DSN'],
+        traces_sample_rate=0.01,
+    )
+
 
 class SpackCIBridge(object):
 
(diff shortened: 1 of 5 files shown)


images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -15,6 +15,14 @@
 import urllib.parse
 import urllib.request
 
+if 'SENTRY_DSN' in os.environ:
+    import sentry_sdk
+
+    sentry_sdk.init(
+        dsn=os.environ['SENTRY_DSN'],
+        traces_sample_rate=0.01,
+    )
```
danlamanna, line 24:
```suggestion
import sentry_sdk

sentry_sdk.init(traces_sample_rate=0.1)
```

See https://docs.sentry.io/platforms/python/configuration/options/#dsn.

I bumped up the performance sampling rate too, I think 10% is reasonable for something that only runs ~12/hr.

danlamanna: Is there any easy way to pass staging/prod as the environment?

mvandenburgh: > Is there any easy way to pass staging/prod as the environment?

There might be, I have an idea I could try. but since we have the separate sentry instance for staging, why would we need to differentiate between them?

danlamanna: 🤦‍♂️ d'oh, nevermind.

danlamanna (approved):

PR #678, opened 2023-10-20
Instrument other cluster python jobs with Sentry

mvandenburgh:
Adds Sentry integration to the rest of the custom Python jobs that run in the cluster.

I didn't instrument any of the build time processor stuff in this PR because I'm not as familiar with how that one works. I'll make a follow up PR for that and consult with @AlmightyYakob on the best way to do that.


AlmightyYakob:
> I didn't instrument any of the build time processor stuff in this PR because I'm not as familiar with how that one works. I'll make a follow up PR for that and consult with @AlmightyYakob on the best way to do that.

It's setup very similarly to the gitlab error processor. I think we could include it in this PR unless you have any objections.

AlmightyYakob (approved):

PR #680, opened 2023-10-23
Fix crash in gitlab error processor upload job

mvandenburgh:
If the `runner` key is nonexistant in the job payload, the python script will crash and no record will be uploaded to opensearch. Since adding Sentry to this job last week, we've seen this happen several times. It seems the `runner` key may be missing when there is a "runner system failure", but unfortunately gitlab does not appear to provide an exact spec for the structure of this JSON blob.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -36,9 +36,9 @@ jobs:
           - docker-image: ./images/opensearch-index-build-logs
             image-tags: ghcr.io/spack/opensearch-index-build-logs:0.0.4
           - docker-image: ./images/gitlab-error-processor
-            image-tags: ghcr.io/spack/gitlab-error-processor:0.0.2
+            image-tags: ghcr.io/spack/gitlab-error-processor:0.0.3
           - docker-image: ./images/upload-gitlab-failure-logs
-            image-tags: ghcr.io/spack/upload-gitlab-failure-logs:0.0.2
+            image-tags: ghcr.io/spack/upload-gitlab-failure-logs:0.0.3
           - docker-image: ./images/snapshot-release-tags
             image-tags: ghcr.io/spack/snapshot-release-tags:0.0.3
           - docker-image: ./images/cache-indexer
--- a/images/gitlab-error-processor/job-template.yaml
+++ b/images/gitlab-error-processor/job-template.yaml
@@ -15,7 +15,7 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: gitlab-error-processing-job
-        image: ghcr.io/spack/upload-gitlab-failure-logs:0.0.2
+        image: ghcr.io/spack/upload-gitlab-failure-logs:0.0.3
         imagePullPolicy: Always
         env:
           - name: GITLAB_TOKEN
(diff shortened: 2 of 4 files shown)


mvandenburgh:
I'm working on adding some tests to this script, but I'm pushing up this fix first so we can get it deployed ASAP and stop losing data.

scottwittenburg (dismissed):
I'm ok with this as long as missing the pod status isn't going to cause issues later on in the submission workflow.  I do have a question though, is it possible we don't really have any error taxonomy to process when the job failed due to a runner system failure?  I think it makes sense to guard against the failure in the processor though.

danlamanna (dismissed):

mvandenburgh:
> I'm ok with this as long as missing the pod status isn't going to cause issues later on in the submission workflow. 

UO jobs also have no pod status, so this shouldn't cause any issues there.

> is it possible we don't really have any error taxonomy to process when the job failed due to a runner system failure? I think it makes sense to guard against the failure in the processor though.

It looks like job logs are still produced in some cases of a runner system failure. See https://gitlab.spack.io/spack/spack/-/jobs/8826259.

mvandenburgh:
It looks like this is not going to be enough to fix things. I wrote some integration tests for the upload script, and while this PR does help "runner failure" payloads make it further into the script, they eventually crash again due to the absence of the `project_id` field. It's kind of bizarre that that field would _ever_ be missing from a job webhook payload, as a job will always have an associated projected regardless of whether or not it passes or fails. For example, [the latest error in sentry](https://sentry.spack.io/organizations/sentry/issues/360/?project=6&query=is%3Aunresolved&referrer=issue-stream&stream_index=0) had this as a payload:
```
{
    before_sha: "a6466b9dddf59cc185800ac428bd4ba535b96c2e",
    build_id: 8826259,
    build_name: "py-scipy@1.5.4 /mscpuhz %gcc@11.3.0 arch=linux-ubuntu22.04-x86_64_v3 Machine Learning",
    build_stage: "stage-13",
    build_status: "failed",
    object_kind: "build",
    ref: "develop",
    retries_count: 1,
    sha: "a452e8379e12bd46925df30b99cf4b30edf80457",
    tag: False,
}
```

This seems like more of a problem with gitlab, but it does leave us in a bad place. I don't like the idea of having a default project ID to fall back on, but I can't think of anything better than that.

AlmightyYakob:
> This seems like more of a problem with gitlab, but it does leave us in a bad place. I don't like the idea of having a default project ID to fall back on, but I can't think of anything better than that.

Do you know how common it is that the `project_id` field (or any other fields we "require") is missing? If it's not a large amount, I think we could just ignore events to the webhook that don't contain the fields we need. If it _is_ a decent percentage of the events we recieve, we might need to look into more direct ways of solving the issue.

danlamanna (approved):

mvandenburgh:
> Do you know how common it is that the `project_id` field (or any other fields we "require") is missing? If it's not a large amount, I think we could just ignore events to the webhook that don't contain the fields we need. 

It's impossible to know at this point, since none if these records are being uploaded properly. Now that this PR is merged, we should get a better idea through sentry.

PR #691, opened 2023-10-31
Fix build timing script and image versions

AlmightyYakob:
There are a couple of things that are causing the analytics DB to not contain the added fields from #635 

1. I forgot to add the `@contextmanager` decorator to the `get_job_artifacts_file` function. The lack of this decorator _should_ have caused sentry errors that would point to the issue. However, that didn't happen because...
2. Due to the fallout of #635 regarding version collision, the hastily made follow-up PR #690 didn't actually bump the version of `upload-build-timings` that was used in the build timing webhook. So events that were being sent to the webhook were _still_ using the old version of the script, and not adding in those fields.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -46,9 +46,9 @@ jobs:
           - docker-image: ./images/cache-indexer
             image-tags: ghcr.io/spack/cache-indexer:0.0.3
           - docker-image: ./images/build-timing-processor
-            image-tags: ghcr.io/spack/build-timing-processor:0.0.3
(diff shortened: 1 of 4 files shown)


AlmightyYakob:
> Do you need to bump this image version as well? [`fix-upload-build-timing`/k8s/production/custom/build-timing-processor/deployments.yaml#L26](https://github.com/spack/spack-infrastructure/blob/fix-upload-build-timing/k8s/production/custom/build-timing-processor/deployments.yaml?rgh-link-date=2023-11-01T14%3A43%3A38Z#L26)

Yes, good catch

mvandenburgh (approved):

PR #693, opened 2023-11-01
Add opensearch staging deployment

danlamanna:
This adds a basic auth deployment of opensearch to staging instead of using cognito. Setting up a separate cognito deployment would have been significantly more work. This opensearch deployment is also a bit different in specific configurations that still have to be fully codified in terraform, but it gets us closer to staging/production parity.

I think deployment for this change will need special attention paid to the password resource to avoid changing it in production, see https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password#import.

### DIFF
--- a/terraform/modules/spack/opensearch.tf
+++ b/terraform/modules/spack/opensearch.tf
@@ -1,11 +1,14 @@
 locals {
   domain_endpoint_name = "opensearch.${var.deployment_name == "prod" ? "" : "${var.deployment_name}."}spack.io"
-  cognito_enabled = var.deployment_name == "prod"
+  cognito_enabled      = var.deployment_name == "prod"
+}
+
+resource "random_password" "opensearch_password" {
+  length  = 64
+  special = true
 }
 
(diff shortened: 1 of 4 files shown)


terraform/modules/spack/opensearch.tf
```diff
@@ -1,28 +1,40 @@
...
     enabled                        = true
     internal_user_database_enabled = true
+
+    dynamic "master_user_options" {
+      for_each = var.deployment_name == "prod" ? [] : [1]
+
+      content {
+        master_user_name     = "admin"
+        master_user_password = random_password.opensearch_password.result
+      }
```
mvandenburgh, line 26: We use the existing production password in some places, such as https://github.com/spack/spack-infrastructure/blob/main/images/gitlab-error-processor/job-template.yaml#L54. So, we'll need to be careful of that. Since we're specifying `length = 64` and `special = true`, I'm not sure if the existing prod password is compatible.

mvandenburgh (approved):

PR #695, opened 2023-11-08
Update runner pressure db with new cray runner names

scottwittenburg:
Update the existing UO cray panel so it shows only Cray RHEL jobs, and add  a new panel that shows only Cray SLES jobs.


scottwittenburg:
fyi @eugeneswalker

eugeneswalker (approved):
This is great! Thank you!

PR #696, opened 2023-11-09
Analytics DB: Add error taxonomy table

mvandenburgh:
Adds a table for storing error taxonomy data. The idea is that this will be used to store error taxonomy data derived from failed gitlab job webhooks instead of OpenSearch. 

@AlmightyYakob I just added the model + migration for the new table; I figured it would make sense to get consensus on that first, and then I'll make a follow up PR that updates the upload script to make use of the new table.


analytics/analytics/models.py
```diff
@@ -66,3 +66,23 @@ class Meta:
...
+    kubernetes_job = models.BooleanField(
+        help_text="True if this job is a Kubernetes job, False otherwise."
+    )
+
+    error_taxonomy = models.CharField(max_length=64)
+    error_taxonomy_version = models.CharField(max_length=32)
+
+    webhook_payload = models.JSONField(
+        help_text="The JSON payload received from the GitLab job webhook."
+    )
```
AlmightyYakob, line 88: These webhooks payloads are decently large, correct? I'm somewhat concerned about storing lots of data here, but I don't think we should block on this. Just might be something to keep an eye on.

mvandenburgh: They're not small, but I don't think they're anything too big for postgres (especially relative to the amount of timing data we're already storing). Here's an example of one https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#job-events

analytics/analytics/models.py
```diff
@@ -66,3 +66,23 @@ class Meta:
...
+    created = models.DateTimeField(auto_now_add=True)
+    job = models.OneToOneField(
+        Job, related_name="error_taxonomy", on_delete=models.CASCADE
+    )
+
+    attempt_number = models.PositiveSmallIntegerField()
+    retried = models.BooleanField()
+    kubernetes_job = models.BooleanField(
+        help_text="True if this job is a Kubernetes job, False otherwise."
+    )
```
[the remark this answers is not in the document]
mvandenburgh: Ah, yes it looks like it. I'll remove this field

analytics/analytics/models.py
AlmightyYakob: Yeah, I think we should just keep an eye on it. We only have ~550k records in the job table atm, so the amount of data this is likely to represent isn't huge anyway, but it's definitely dwarfing most other fields we store.

AlmightyYakob (approved):
LGTM. I usually run the database migration after the relevant PR is merged, but since this PR doesn't update the container images, I'll hold off until the follow-up.

I need to solve the issue of running db migrations more generally, but for now I'm running them manually.

PR #701, opened 2023-11-30
Switch to new `kubectl` Terraform provider

mvandenburgh:
The current provider we're using (https://github.com/gavinbunney/terraform-provider-kubectl/) is now unmaintained and contains some critical bugs. We should fix this by switching to the official k8s provider, but that will require a refactor of a significant part of our Terraform code base. To address the immediate issues being caused by the buggy provider, this PR switches us to [this fork](https://github.com/gavinbunney/terraform-provider-kubectl/), which is actively maintained and fixes the bugs we're seeing while being effectively a drop-in replacement.

### DIFF
--- a/terraform/modules/spack/main.tf
+++ b/terraform/modules/spack/main.tf
@@ -9,8 +9,8 @@ terraform {
       version = "~> 2.5"
     }
     kubectl = {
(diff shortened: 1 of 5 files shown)


danlamanna (approved):
🙏

PR #704, opened 2023-12-04
Add django webhook handler to save job traces to opensearch

danlamanna:
This adds a k8s service to host our django application internally to handle webhook requests for saving job traces. In the near future I think we should aim to consolidate our FastAPI apps to be similar endpoints under django. At the moment this PR functions but has a few question marks. It should be ready for general feedback.


k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,91 @@
...
+      labels:
+        app: webhook-handler
+        svc: web
+    spec:
+      restartPolicy: Always
+      serviceAccountName: webhook-handler
+      containers:
+        - name: webhook-handler
+          #image: ghcr.io/danlamanna/webhook-handler:0.0.11
+          image: ghcr.io/spack/upload-build-timings:0.0.5
```
AlmightyYakob, line 28: Should we rename this image?

AlmightyYakob (commented):
Most of this looks good to me

analytics/analytics/views.py
```diff
@@ -0,0 +1,18 @@
...
+import sentry_sdk
+
+from analytics.job_log_uploader import upload_job_log
+
+
+def webhook_handler(request: HttpRequest) -> HttpResponse:
+    job_input_data: dict[str, Any] = json.loads(request.body)
+
+    if job_input_data.get("object_kind") != "build":
+        raise ValueError("Not a build event")
```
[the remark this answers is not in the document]
danlamanna: This code is copied from other webhooks, but AFAICT it's there so that it doesn't attempt to process events that aren't related to jobs.

terraform/modules/spack/gitlab_webhooks.tf
```diff
@@ -16,3 +17,53 @@ resource "gitlab_project_hook" "job_webhook" {
 
 // TODO: Once https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/issues/1350 is resolved the
 // gitlab_application_settings resource should be used to whitelist the domains in job_webhooks.
+
+
+data "gitlab_user" "spackbot" {
+  username = "spackbot"
```
danlamanna, line 23: Does it make sense to use spackbot here, or something else? @AlmightyYakob @mvandenburgh

k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,91 @@
...
+  labels:
+    app: webhook-handler
+    svc: web
+spec:
+  selector:
+    matchLabels:
+      app: webhook-handler
+      svc: web
+      # TODO: evaluate
+  replicas: 2
```
danlamanna, line 16: I tried reading the documentation, but I'm not sure if replicas does anything here? Would it be load balanced or something? Or should this be removed? @AlmightyYakob @mvandenburgh

analytics/analytics/settings.py
```diff
@@ -44,3 +45,14 @@
 # Default primary key field type
 # https://docs.djangoproject.com/en/4.2/ref/settings/#default-auto-field
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+SECRET_KEY = os.environ["SECRET_KEY"]
```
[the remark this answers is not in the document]
danlamanna: I suppose they'll continue to work because they're still running 0.4 of the docker image for now 😬.

k8s/production/custom/webhook-handler/deployments.yaml
mvandenburgh: Yeah it would be load balanced. If there are multiple replicas, the assocated `Service` will dispatch requests to one randomly.

AlmightyYakob: Since this is set up behind a service, I think the effect would be that this load balances the incoming requests. According to [this](https://stackoverflow.com/questions/48789227/does-clusterip-service-distributes-requests-between-replica-pods) StackOverflow thread that's the behavior, although I couldn't find it explicitly mentioned in the docs as they claim. However, based on that thread and our past usage of this configuration, I believe this is the case.

All that being said, I think 2 is a fine value for this.

analytics/analytics/views.py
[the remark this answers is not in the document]
AlmightyYakob: The main reason I ask is because if we don't want this to be treated as an error to be dealt with (and so we don't get this error in sentry), we could return a 400 here instead of raising an exception.

analytics/analytics/settings.py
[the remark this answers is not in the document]
AlmightyYakob: Should we update the job-template for the build-time-processor to include this new secret value?

analytics/analytics/views.py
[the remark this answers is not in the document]
danlamanna: I changed it to return a 400 but also catch a sentry error, since I think a misconfiguration of our webhooks is indicative of a larger problem.

k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,190 @@
...
+  name: webhook-handler-worker
+  namespace: custom
+  labels:
+    app: webhook-handler-worker
+    svc: web #TODO
+spec:
+  selector:
+    matchLabels:
+      app: webhook-handler-worker
+      svc: web # TODO
```
mvandenburgh, line 110:
```suggestion
```

k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,190 @@
...
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: webhook-handler-worker
+  namespace: custom
+  labels:
+    app: webhook-handler-worker
+    svc: web #TODO
```
mvandenburgh, line 105:
```suggestion
```

k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,190 @@
...
+    matchLabels:
+      app: webhook-handler-worker
+      svc: web # TODO
+      # TODO: evaluate
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        app: webhook-handler-worker
+        svc: web # TODO
```
mvandenburgh, line 117:
```suggestion
```

k8s/production/custom/webhook-handler/deployments.yaml
```diff
@@ -0,0 +1,190 @@
...
+      containers:
+        - name: webhook-handler-worker
+          image: ghcr.io/spack/upload-build-timings:0.0.5
+          command: ["celery", "-A", "analytics.celery", "worker", "-l", "info"]
+          imagePullPolicy: Always
+          resources:
+          # TODO: evaluate
+            requests:
+              cpu: 350m
+              memory: 1G
```
mvandenburgh, line 130: I think these make sense. With the default concurrency of 4, we can potentially have 4 job logs loaded into memory at once on a single worker, which seems like 1GB of memory should always be able to handle.

analytics/analytics/settings.py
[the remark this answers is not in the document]
danlamanna: We decided to fix the build timings script in a follow up PR, it'll still work since it uses the old docker image.

k8s/production/custom/webhook-handler/deployments.yaml
danlamanna: Renamed to django.

images/build-timing-processor/job-template.yaml
```diff
@@ -18,6 +18,7 @@ spec:
         - name: build-timing-processing-job
           image: ghcr.io/spack/upload-build-timings:0.0.4
           imagePullPolicy: Always
+          command: ["./manage.py", "upload_build_timings"]
```
jjnesbitt, line 21: Just to be clear, the plan is to keep the build timings uploader running separately from the django app at first, and to then later fold that into the django webhook handler?

terraform/modules/spack/gitlab_webhooks.tf
```diff
@@ -16,3 +17,56 @@ resource "gitlab_project_hook" "job_webhook" {
...
+     kind: Secret
+     metadata:
+       name: webhook-secrets
+       namespace: custom
+     data:
+       gitlab-endpoint: ${base64encode("${var.gitlab_url}")}
+       gitlab-token: ${base64encode("${gitlab_personal_access_token.webhook_handler.token}")}
+       sentry-dsn: ${base64encode("${data.sentry_key.webhook_handler.dsn_public}")}
+       secret-key: ${base64encode("${random_password.webhook_handler.result}")}
+       celery-broker-url: ${base64encode("redis://${aws_elasticache_replication_group.pr_binary_graduation_task_queue.primary_endpoint_address}:6379/1")}
```
jjnesbitt, line 54: It feels a bit weird that the gitlab webhook processor would rely on this specific redis cache. Would it not make sense to create a new (small) cache instance just for this webhook processor? Or to just deploy something like rabbitmq as part of the two existing deployments (if not its own deployment)?

My main concern is that these two seemingly unrelated parts of the infrastructure could end up affecting one another in some way.

images/build-timing-processor/job-template.yaml
danlamanna: Yeah, I already have that change made locally. I just want to verify this PR is deployed and working first before moving more to it.

terraform/modules/spack/gitlab_webhooks.tf
danlamanna: Yeah I discussed this with @mvandenburgh before since it seems a bit awkward. I think we're on the same page that we want to consolidate infrastructure so using a differently namespaced database `/1` on the same instance of redis makes sense. I'll watch the redis utilization after merge but we're almost comically overprovisioned at this point.

danlamanna: FWIW, I pushed a change to explicitly listen on the celery queue and double checked that the spackbot queue uses database 0.

jjnesbitt (dismissed):

jjnesbitt (approved):

PR #706, opened 2023-12-11
Fix pod annotations

cmelone:
#630 added some job information to be exposed through attributes. In that PR, I tried including the entire `metrics` section by doing this:

```
     metricAnnotationsAllowList:
        - pods=[gitlab/ci_job_id,metrics*]
```

However, I checked the [docs](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/cli-arguments.md#available-options) and it seems like the only options are specifying each annotation manually or including all (performance isn't great). See [this prometheus query](https://prometheus.spack.io/api/v1/query_range?query=kube_pod_annotations%7Bannotation_gitlab_ci_job_id=%279359381%27%7D&start=1701780555&end=1701800555&step=10) for an example of this not working.

This PR lists each annotation we would like to have available in prometheus.

Is there a way to test if this would work beforehand in a staging area before pushing to production? No worries if not

### DIFF
--- a/k8s/production/prometheus/release.yaml
+++ b/k8s/production/prometheus/release.yaml
@@ -149,16 +149,7 @@ spec:
         - pods=[*]
         - nodes=[*]
       metricAnnotationsAllowList:
-        - > pods=[
-                gitlab/ci_job_id,
-                metrics/spack_job_spec_pkg_name,
-                metrics/spack_job_spec_pkg_version,
-                metrics/spack_job_spec_compiler_name,
-                metrics/spack_job_spec_compiler_version,
-                metrics/spack_job_spec_arch,
-                metrics/spack_job_spec_variants,
-                metrics/spack_job_build_jobs
-            ]
+        - pods=[gitlab/ci_job_id,metrics/spack_job_spec_pkg_name,metrics/spack_job_spec_pkg_version,metrics/spack_job_spec_compiler_name,metrics/spack_job_spec_compiler_version,metrics/spack_job_spec_arch,metrics/spack_job_spec_variants,metrics/spack_job_build_jobs]
 
   # Configure OpenSearch datasource
   valuesFrom:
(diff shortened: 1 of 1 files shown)


k8s/production/prometheus/release.yaml
```diff
@@ -149,7 +149,7 @@ spec:
         - pods=[*]
         - nodes=[*]
       metricAnnotationsAllowList:
-        - pods=[gitlab/ci_job_id,metrics*]
+        - pods=[gitlab/ci_job_id,metrics/spack_job_spec_pkg_name,metrics/spack_job_spec_pkg_version,metrics/spack_job_spec_compiler_name,metrics/spack_job_spec_compiler_version,metrics/spack_job_spec_arch,metrics/spack_job_spec_variants,metrics/spack_job_build_jobs]
```
AlmightyYakob, line 152: Is it possible to format this like so?
```suggestion
        - > pods=[
                gitlab/ci_job_id,
                metrics/spack_job_spec_pkg_name,
                metrics/spack_job_spec_pkg_version,
                metrics/spack_job_spec_compiler_name,
                metrics/spack_job_spec_compiler_version,
                metrics/spack_job_spec_arch,
                metrics/spack_job_spec_variants,
                metrics/spack_job_build_jobs
            ]
```

I'm not sure if this will work, but according to yaml syntax I _think_ it would.

AlmightyYakob:
I'm having issues with the staging Prometheus instance, but I can say for sure that the syntax in [c68c6a1](https://github.com/spack/spack-infrastructure/pull/706/commits/c68c6a15a246a2acc70a69303e079a3c38e29a80) is invalid. It can be fixed to preserve the whitespace, but I have no way to test if the annotation allow list works correctly even _with_ properly formatted whitespace.

If you just revert [c68c6a1](https://github.com/spack/spack-infrastructure/pull/706/commits/c68c6a15a246a2acc70a69303e079a3c38e29a80) I think this is good to go.

cmelone:
done, thanks

AlmightyYakob (approved):

PR #708, opened 2023-12-14
Enable fluent-bit for pipeline pods

mvandenburgh:
We're excluding pipeline pods from fluent-bit, so when a job pod fails, we don't have any logs to debug.


mvandenburgh:
We discussed the rationale behind the original decision to disable this - it's because fluent-bit sends each log line separately, making it very difficult/impossible to get any meaningful info from them.

PR #710, opened 2023-12-18
Update runners and signing sealed secrets

jjnesbitt:
Note: All entries in both the affected secrets have been re-sealed, even though each one only had a new data field added. This is due to how sealed secrets work, and isn't an issue (the other entries in the secret data object haven't actually been changed).


zackgalbreath (approved):

jjnesbitt:
Holding off on merging until the new year.

PR #721, opened 2024-01-05
Run django migrations before running server

jjnesbitt:
My only remaining concern is if we should have some check in place to make sure we're not running an old image version? I was thinking of using the migration system to check for that, although I don't know how exactly.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -48,7 +48,7 @@ jobs:
           - docker-image: ./images/build-timing-processor
             image-tags: ghcr.io/spack/build-timing-processor:0.0.4
           - docker-image: ./analytics
-            image-tags: ghcr.io/spack/django:0.0.1
+            image-tags: ghcr.io/spack/django:0.0.2
     steps:
       - name: Checkout
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
--- a/analytics/entrypoint.sh
+++ b/analytics/entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+set -e
+
+# Always run migrate when invoking this image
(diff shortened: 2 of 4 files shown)


.github/workflows/custom_docker_builds.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           - docker-image: ./images/build-timing-processor
             image-tags: ghcr.io/spack/build-timing-processor:0.0.4
           - docker-image: ./analytics
-            image-tags: ghcr.io/spack/django:0.0.1
+            image-tags: ghcr.io/spack/django:0.0.2
```
danlamanna, line 51: Don't the references need to be updated for this as well?

danlamanna: The celery runners also override this command and probably need to run migrate as well. What do you think?

jjnesbitt: Good catch, I totally forgot about the other references. Regarding the command overwriting in celery as well, would it make sense to move the call to `migrate` into the entrypoint? Essentially using the entrypoint script pattern shown [here](https://stackoverflow.com/a/41518225).

jjnesbitt: (testing that pattern locally seems to work as intended)

danlamanna: Sounds good to me.

analytics/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Always run migrate when invoking this image
+python manage.py migrate
+
+# This will exec the CMD passed to docker, i.e. "gunicorn" or "celery"
+exec "$@"
```
danlamanna, line 7: Will this exec even if migrate fails?

danlamanna (approved):

PR #722, opened 2024-01-10
Add job info to sentry tags for webhook jobs

danlamanna:


analytics/analytics/build_timing_processor.py
```diff
@@ -88,6 +90,7 @@ def get_timings_json(job: ProjectJob) -> list[dict]:
 def upload_build_timings(job_input_data_json: str):
     # Read input data and extract params
     job_input_data = json.loads(job_input_data_json)
+    setup_gitlab_job_sentry_tags(job_input_data)
```
jjnesbitt, line 93: Should we not just put the invocation to this function in the webhook endpoint itself?

danlamanna: It'll be run on a different machine. I briefly tried looking into how to reduce the duplication, I'd probably use a custom celery base task, but it didn't really look easier 😕.

jjnesbitt: Right, I see.

.github/workflows/custom_docker_builds.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           - docker-image: ./images/cache-indexer
             image-tags: ghcr.io/spack/cache-indexer:0.0.3
           - docker-image: ./analytics
-            image-tags: ghcr.io/spack/django:0.0.3
+            image-tags: ghcr.io/spack/django:0.0.4
```
jjnesbitt, line 49: This needs to be bumped in https://github.com/spack/spack-infrastructure/blob/main/k8s/production/custom/webhook-handler/deployments.yaml as well right?

danlamanna: 😩

jjnesbitt (approved):

PR #697, opened 2023-11-09
Analytics DB: Annotate pod/node data onto Job model

jjnesbitt:
This PR will allow us to derive "cost per job" data

Adds the following records to the `Job` model:

* job_cpu_request
* job_memory_request
* node_name
* node_uid
* node_cpu
* node_mem
* node_capacity_type
* node_instance_type
* node_instance_type_spot_price

PR #723, opened 2024-01-10
Use prometheus to annotate pod/node data onto Job model

jjnesbitt:
Supersedes #697

This PR collects much more data every time job data is collected by the job webhook. This allows us to derive "cost per job" metrics.

The following notable changes are included in this PR:

* `build_timings_processor.py` has been moved/renamed to the `job_processor` folder, with several files, including one for build timings. This is because at this point, a lot more is going on than just uploading build timings.
* A new `Node` model, which stores data relating to a node that a job ran on, including available cpu/memory, its spot price, etc.
* A new `JobPod` model, which stores data about the pod a job ran on in the cluster, including resource requests, limits, and usage.

The above two models are referenced by `Job`, but are null-able, since jobs that weren't run in the cluster (uo, etc.), as well as historical aws jobs, won't have any available data.

There are a few more minor changes worth mentioning:

* The `Job.duration` field has been changed from a `FloatField` to a `DurationField`, to better represent the data and make it easier to work with.
* The `Job.aws` field has been changed to be no longer null-able. The field was made null-able in https://github.com/spack/spack-infrastructure/pull/724, to get around a quirk of the gitlab API. This PR circumvents that part of the API entirely, and so null values shouldn't be allowed. The amount of rows in the database that have a null value in the `aws` column is only a handful at the moment, and so those rows are deleted as a part of the migration, to clean things up.


analytics/analytics/job_processor/prometheus.py
```diff
@@ -0,0 +1,293 @@
...
+from datetime import datetime
+from urllib.parse import urlencode
+
+import requests
+from kubernetes.utils.quantity import parse_quantity
+
+from analytics.models import Job, JobPod, Node
+
+CLUSTER_INTERNAL_PROMETHEUS_URL = (
+    "kube-prometheus-stack-prometheus.monitoring.svc.cluster.local:9090"
```
danlamanna, line 12: Can this be parameterized?

jjnesbitt: Parameterized in what way? The `PrometheusClient` already takes an argument to specify the url to use, so this just serves as a constant for the value we use in prod.

If you're referring to the fact that this value is hard-coded at all, I agree, however I couldn't find a clear way to retrieve this value from the cluster via secrets, config-maps, etc.

Did you have something specific in mind?

danlamanna: Is this value the same in staging and production? If so, I guess this is fine - though it is a shame we can't populate it into the settings somehow.

analytics/analytics/job_processor/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    job_input_data = json.loads(job_input_data_json)
+    setup_gitlab_job_sentry_tags(job_input_data)
+
+    # Retrieve project and job from gitlab API
+    gl = gitlab.Gitlab(settings.GITLAB_ENDPOINT, settings.GITLAB_TOKEN)
+    gl_project = gl.projects.get(job_input_data["project_id"])
+    gl_job = gl_project.jobs.get(job_input_data["build_id"])
+
+    # Get or create job record
+    job = create_job(gl, gl_project, gl_job)
```
danlamanna, line 65: Should these 2 statements be made transactional, and retries added to the celery task? I'm particularly worried about repeated calls to the prometheus API that are bound to fail eventually.

analytics/analytics/job_processor/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    # Save and return new job
+    job.save()
+    return job
+
+
+@shared_task(name="process_job")
+def process_job(job_input_data_json: str):
+    # Read input data and extract params
+    job_input_data = json.loads(job_input_data_json)
+    setup_gitlab_job_sentry_tags(job_input_data)
```
danlamanna, line 57: 🙏

analytics/analytics/job_processor/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Retrieve project and job from gitlab API
+    gl = gitlab.Gitlab(settings.GITLAB_ENDPOINT, settings.GITLAB_TOKEN)
```
danlamanna, line 60: This can be a future project, but we should look at setting `retry_transient_errors` and maybe using a global gitlab client (https://python-gitlab.readthedocs.io/en/stable/api-usage-advanced.html#transient-errors). We're already seeing how occasional failures creep in to the GitLab API with other jobs.

analytics/analytics/job_processor/prometheus.py
```diff
@@ -0,0 +1,293 @@
...
+            end=job.finished_at,
+            step=step,
+        )
+
+        # First, get the cpu utlization by the pod we care about
+        # To do this, just get the last value from the response, since that'll be the total of the counter
+        pod_results = next(
+            (res for res in results if res["metric"]["pod"] == pod), None
+        )
+        assert pod is not None, f"Pod {pod} not found in query"
```
danlamanna, line 164: I was recently scolded for using `assert` instead of an if statement, rightfully so I think 😞.

analytics/analytics/models.py
```diff
@@ -8,12 +58,21 @@ class Job(models.Model):
...
     started_at = models.DateTimeField()
-    duration = models.FloatField()
+    duration = models.DurationField()
     ref = models.CharField(max_length=256)
     tags = ArrayField(base_field=models.CharField(max_length=32), default=list)
     package_name = models.CharField(max_length=128)
 
-    # Fields allow null to accomodate historical data
+    # Whether this job ran in the cluster or not
+    aws = models.BooleanField(default=True)
```
danlamanna, line 67: Is there a reason to model this as `aws` instead of a choices field for `aws/uo` and others in the future?

danlamanna: Also, I think the comment could be migrated to `help_text` in the field.

analytics/analytics/settings.py
```diff
@@ -67,3 +67,5 @@
 
 GITLAB_ENDPOINT: str = os.environ["GITLAB_ENDPOINT"]
 GITLAB_TOKEN: str = os.environ["GITLAB_TOKEN"]
+
+SPACK_PROMETHEUS_ENDPOINT: str = os.environ.get("SPACK_PROMETHEUS_ENDPOINT")
```
danlamanna, line 71: Should this use `.get` or require the endpoint be set?

analytics/analytics/tests/test_build_timings.py
```diff
@@ -21,7 +21,7 @@ def build_json_string():
 
 @pytest.mark.django_db
 def test_upload_build_timings(build_json_string):
-    upload_build_timings(build_json_string)
+    process_job(build_json_string)
```
danlamanna, line 24: Out of curiosity, does this exercise a lot of your new code?

analytics/analytics/job_processor/artifacts.py
```diff
@@ -0,0 +1,54 @@
...
+    pipeline_yml_filename = "jobs_scratch_dir/reproduction/cloud-ci-pipeline.yml"
+    with get_job_artifacts_file(gljob, pipeline_yml_filename) as pipeline_file:
+        raw_pipeline = yaml.safe_load(pipeline_file)
+
+    pipeline_vars = raw_pipeline.get("variables", {})
+    job_vars = raw_pipeline.get(gljob.name, {}).get("variables", {})
+    if not job_vars:
+        raise Exception(f"Empty job variables for job {gljob.id}")
+
+    job.package_name = job_vars["SPACK_JOB_SPEC_PKG_NAME"]
```
cmelone, line 44: isn't most of the info you're grabbing here in the annotations metric?

analytics/analytics/job_processor/__init__.py
cmelone, line 60: just for my own edification, is the purpose of this function to handle the webhook payload from gitlab? If so, what info is not available in that data that necessitates an API call?

analytics/analytics/tests/test_build_timings.py
jjnesbitt: This change was mainly done to keep the existing behavior consistent, although `process_job` does pretty much exercise all of my new code.

analytics/analytics/job_processor/__init__.py
jjnesbitt: Great idea, I agree.

analytics/analytics/job_processor/__init__.py
jjnesbitt: I think that makes sense. What's the motivation for a global gitlab client? Is it just so it's initialized once when django is loaded, re-using the same client session in all parts of the application?

danlamanna: Yeah, and it's just reusing the client in order to add convenience and reduce the chance that we're using differently-configured clients in different jobs.

analytics/analytics/models.py
jjnesbitt: The main reason is that if `aws=True` (the job ran in the cluster), then we have extra data for it. If not, we don't have that data. The rest doesn't _really_ matter at the moment, since the data we're gathering is mainly focused on the things we control and can optimize (jobs/runners in the cluster). 

Where exactly a runner ran outside the cluster also might not be easy to determine. For example, in the case of UO runners, you could just look at the runner name/description to see if it starts with UO. But sometimes you can't access the runner, and that might not generalize to other "classes" of runner.

If in the future we have a concrete way to determine where a runner ran, I'd be happy to make the change.

analytics/analytics/job_processor/__init__.py
jjnesbitt: The main things not available from the webhook payload seem to be the job trace/artifacts and the job tag list.

analytics/analytics/job_processor/prometheus.py
jjnesbitt: It is the same in prod/staging, as it's just the internal IP to the service. I agree it's a shame.

analytics/analytics/job_processor/__init__.py
jjnesbitt: We talked offline and agreed to punt on adding retries, so that we can know better what would actually need to be retried specifically.

analytics/analytics/job_processor/prometheus.py
jjnesbitt: [2fb0190](https://github.com/spack/spack-infrastructure/pull/723/commits/2fb0190ef37d54034200acfd82211f98ba8733e8)

analytics/analytics/settings.py
jjnesbitt: [cd8879d](https://github.com/spack/spack-infrastructure/pull/723/commits/cd8879d15f3e0990b842a7de2808cafced7e3f47)

analytics/analytics/job_processor/__init__.py
jjnesbitt: I addressed half of this in [7bb549b](https://github.com/spack/spack-infrastructure/pull/723/commits/7bb549b05eb8b1a49b877489d63d2bfd48ff88ae). The global gitlab client thing I would also like to implement at some point, but it's probably easier to do that in a follow-on PR, after all the concurrent work around the django app has been settled. I'll file an issue.

analytics/dev/.env.docker-compose
```diff
@@ -10,3 +10,4 @@ CELERY_BROKER_URL=redis://localhost:6379/0
 SECRET_KEY=deadbeef
 GITLAB_ENDPOINT="http://fakeurl"
 GITLAB_TOKEN="bar"
+PROMETHEUS_URL="http://localhost:9090
```
danlamanna, line 13: Does this work with the quoting? And I think it needs to be added to the native version of this file too.

jjnesbitt: Whoops, that's a typo. Also, there's no (existing) native version of this file.

danlamanna: Oh right, that's my PR 🤦‍♂️

danlamanna (approved):

danlamanna (dismissed):

danlamanna (approved):

danlamanna (approved):

jjnesbitt:
I've been testing this and am seeing times of around 3-5 seconds for the `process_job` task.

danlamanna (approved):

PR #724, opened 2024-01-11
Handle jobs with missing runners

jjnesbitt:
Sometimes the runner of a job can't be found or is otherwise unable to be retrieved from the gitlab API. This PR handles those cases, also updating the `aws` field of a Job to be null-able, to accommodate these situations.


analytics/analytics/build_timing_processor.py
```diff
@@ -64,7 +65,13 @@ def get_job_metadata(job: ProjectJob) -> dict:
 
 def create_job(gl: gitlab.Gitlab, project: Project, job: ProjectJob) -> Job:
     # grab runner tags
-    runner_tags = gl.runners.get(job.runner["id"]).tag_list
+    aws = None
+    if job.runner is not None:
+        # For various reasons, sometimes the runner is missing
+        try:
+            aws = "aws" in gl.runners.get(job.runner["id"]).tag_list
+        except GitlabHttpError:
```
danlamanna, line 73: Does this catch only the type of errors we want to catch? What if there's a transient 500?

jjnesbitt: Good point. I'll look into how to narrow it to just the "runner not found" error.

danlamanna (approved):

PR #725, opened 2024-01-11
Enhance docker-compose setup and django configuration

danlamanna:
This adds some boilerplate to the django install that D&A uses on other projects. It should streamline developer setup and testing for the webhooks we have configured.


analytics/setup.py
```diff
@@ -0,0 +1,67 @@
...
+        "Framework :: Django :: 3.0",
+        "Framework :: Django",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python",
+    ],
+    python_requires=">=3.10",
```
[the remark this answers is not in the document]
mvandenburgh: Discussed this in-person, Girder 4 doesn't support Python 3.12.

analytics/setup.py
```diff
@@ -0,0 +1,67 @@
...
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    license="Apache 2.0",
+    author="Kitware, Inc.",
+    author_email="kitware@kitware.com",
+    keywords="",
+    classifiers=[
+        "Development Status :: 3 - Alpha",
+        "Environment :: Web Environment",
+        "Framework :: Django :: 3.0",
```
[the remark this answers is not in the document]
danlamanna: We're also stuck on 4.2 for now.

danlamanna:
@mvandenburgh I rebased this on main.

@jjnesbitt Could you check out this PR and attempt to spin it up and verify it works in some basic capacity?

analytics/dev/.env.docker-compose
```diff
@@ -1,13 +1,11 @@
...
-DB_HOST=localhost
-DB_PORT=5432
-OPENSEARCH_ENDPOINT=http://localhost:9200
+DJANGO_SETTINGS_MODULE=analytics.settings.development
+DJANGO_DATABASE_URL=postgres://postgres:postgres@postgres:5432/django
+DJANGO_CELERY_BROKER_URL=redis://redis:6379/
+OPENSEARCH_ENDPOINT=http://elasticsearch:9200
 OPENSEARCH_USERNAME=elastic
 OPENSEARCH_PASSWORD=elastic
 CELERY_BROKER_URL=redis://localhost:6379/0
```
jjnesbitt, line 7: Is there a reason for this being included alongside `DJANGO_CELERY_BROKER_URL`? When I run `docker-compose up`, it picks up this env var and fails to connect to redis, as it's not trying to connect to `localhost` instead of `redis`.

danlamanna: Probably a bad merge conflict or just a mistake.

analytics/analytics/core/tests/test_build_timings.py
```diff
@@ -1,9 +1,11 @@
 import json
 import pytest
-from analytics.job_log_uploader import upload_job_log
+from analytics.core.job_log_uploader import upload_job_log
 from analytics.job_processor import process_job
 
-from analytics.models import Timer
+from analytics.core.job_log_uploader import upload_job_log
+from analytics.core.models import Timer
+from analytics.core.build_timing_processor import upload_build_timings
```
jjnesbitt, line 8: This is old
```suggestion
```

jjnesbitt:
Seems to work for me locally. I ran `migrate` both against my existing local database, as well as a fresh db, without any issue. Testing local queries against the database works fine as well.

I tried running some of the tox commands, and left a change suggestion I found from `tox -e test`. Besides that, `lint` and `format` fail with lots of suggestions, but I don't think those changes need to necessarily happen now. Plus, if we wanted to switch to Ruff as a follow up, there may be additional changes anyway.

analytics/requirements.txt
```diff
@@ -1,11 +1,97 @@
...
+idna==3.6
+iniconfig==2.0.0
+ipython==8.20.0
+jedi==0.19.1
+kombu==5.3.5
+kubernetes==29.0.0
+markdown-it-py==3.0.0
+matplotlib-inline==0.1.6
+mdurl==0.1.2
+minio==7.2.3
```
jjnesbitt, line 50: Why do we have both a `requirements.txt` and a `setup.py` file?

danlamanna: One specifies the concrete requirements and the other the abstract ones.

jjnesbitt: And that's used for the prod django image? I'm a bit worrisome of having two independently managed lists of dependencies (i.e. if we change a dependency in `setup.py` and go to re-deploy, it will fail / not work unless we remember to update `requirements.txt`). Do you see this as an issue?

jjnesbitt: I'll leave the settling of this up to you. If you don't think it's an issue or would prefer to address it separately, that's fine.

jjnesbitt (approved):

PR #726, opened 2024-01-12
Disable gitlab aws access key rotation

danlamanna:
All runners should now be using OIDC for authentication. I'm going to disable the secret updating and verify that nothing breaks. After a little while I'll remove all the associated key rotation code.

### DIFF
--- a/k8s/production/custom/rotate-keys/cron-jobs.yaml
+++ b/k8s/production/custom/rotate-keys/cron-jobs.yaml
@@ -15,7 +15,7 @@ spec:
           restartPolicy: Never
           containers:
           - name: rotate-gitlab-aws-access-keys
-            image: ghcr.io/spack/ci-key-rotate:0.0.5
+            image: ghcr.io/spack/ci-key-rotate:0.0.6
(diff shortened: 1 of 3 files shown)


zackgalbreath (approved):
LGTM

Noting here that you correctly did **not** disable the "clear-admin-keys" cron job, which happens to live in the same YAML file.

k8s/production/custom/rotate-keys/cron-jobs.yaml
```diff
@@ -23,6 +23,8 @@ spec:
                 secretKeyRef:
                   name: rotate-keys
                   key: gitlab-token
+            - name: SKIP_GITLAB_VARIABLE_UPDATE
+              value: "true"
```
scottwittenburg, line 27: Does kube automatically convert this to a boolean type so the check of the environment variable will work as written?

danlamanna: I'm not sure, but `"true"` is truthy regardless.

mvandenburgh (approved):

PR #731, opened 2024-01-16
Give Stephen Sachs write permissions to an s3 sub-bucket

becker33:
@stephenmsachs creates binaries as part of the process to bootstrap the images for the pcluster CI stack. This will allow him to store those binaries in the Spack s3.

### DIFF
--- a/terraform/production/s3.tf
+++ b/terraform/production/s3.tf
@@ -32,19 +32,6 @@ resource "aws_s3_bucket_policy" "protected_binaries_restricted" {
         "Principal": "*",
         "Action": "s3:GetObject",
         "Resource": "arn:aws:s3:::${each.key}/*"
-      },
-      {
-        "Sid": "StesachsAtAWSWriteSubFolder",
-        "Effect": "Allow",
-        "Principal": {
-          "AWS": "arn:aws:iam::679174810898:root"
-        },
-        "Action": [
-          "s3:GetObject*",
-          "s3:PutObject*",
-          "s3:DeleteObject*"
-        ],
-        "Resource": "arn:aws:s3:::spack-binaries/pcluster-bootstrap/*"
       }
     ]
   })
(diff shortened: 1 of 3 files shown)


terraform/production/s3.tf
```diff
@@ -32,6 +32,19 @@ resource "aws_s3_bucket_policy" "protected_binaries_restricted" {
         "Principal": "*",
         "Action": "s3:GetObject",
         "Resource": "arn:aws:s3:::${each.key}/*"
+      },
+      {
+        "Sid": "StesachsAtAWSWriteSubFolder",
+        "Effect": "Allow",
+        "Principal": {
+          "AWS": "arn:aws:iam::679174810898:root"
```
kwryankrattiger, line 40: I am not able to verify this, where does this ARN come from?

becker33: @stephenmsachs sent me this ARN for himself.

kwryankrattiger: Then that is probably fine then. If he can just verify that is indeed correct here then I think this looks okay to me.

stephenmsachs: Double checked, it's the correct arn.

danlamanna:
👋 @becker33 after discussing this a bit, I think it would probably be easier if this was switched to a separate bucket rather than a special prefix in the `spack-binaries` bucket. I pushed a commit to this PR to add that change. If that sounds good to you let me know and we can merge and deploy this.

becker33:
@danlamanna that's fine by me, I don't remember why we had talked about a special prefix but I'm pretty confident there was no technical reason for it.

@stephenmsachs that work for you?

stephenmsachs:
Will this bukcet be mirrored to a https page as well? I have asked for a sub-key in the spack-binaries bucket as this can be consumed without installing `boto3`.

danlamanna:
> Will this bukcet be mirrored to a https page as well? I have asked for a sub-key in the spack-binaries bucket as this can be consumed without installing `boto3`.

Yes you should be able to download over a browser, curl, etc.

mvandenburgh (approved):

danlamanna:
This should be all set.

e.g. uploading:
```
# set credentials
aws s3 cp somefile.txt s3://spack-bootstrap/pcluster/somefile.txt
```

and downloading:
```
curl https://bootstrap.spack.io/pcluster/somefile.txt
```

stephenmsachs:
It works. Thank you!

PR #733, opened 2024-01-18
Add stack name to annotations

cmelone:
This is a follow-up to #630. With the exception of stack name, all the information I need to retrieve for the dynamic allocation web hook is available in the annotations. Rather than add load to the Prometheus server with an extra request to the labels, it would be most convenient to have the stack name available here as well.

@jjnesbitt


jjnesbitt (approved):

cmelone:
@jjnesbitt would you mind merging when you get the chance? Thanks!

PR #738, opened 2024-01-21
Allow fallback to use AWS environment variable for authentication

kwryankrattiger:
Not sure if this is the best strategy, but it hopefully removes having to patch my PRs every time I want to test something.


danlamanna:
Closing this. Discussed an alternate approach with @kwryankrattiger and @scottwittenburg that involves creating separate sets of buckets for experiment repos. I'll be pushing a pr soon.

PR #747, opened 2024-01-25
Store whether a rebuild job is unnecessary

danlamanna:
Supercedes #633 

I have to test that this works against a couple of sample job logs (ANSI codes might cause a little chaos), but I wanted to get this in front of @jjnesbitt to make sure it looks sane. And @scottwittenburg to see how it's done.


analytics/analytics/core/models.py
```diff
@@ -81,6 +81,10 @@ class Job(models.Model):
     job_size = models.CharField(max_length=128, null=True)
     stack = models.CharField(max_length=128, null=True)
 
+    unnecessary = models.BooleanField(
+        default=False, help_text="Whether this job has 'No need to rebuild' in its trace."
+    )
```
jjnesbitt, line 86: @scottwittenburg Is there a more domain specific term for this?

scottwittenburg: Hmmm :thinking:  

If you were thinking of any specific alternatives, please suggest.  Otherwise, I think `unnecessary` is fine.

analytics/analytics/job_processor/__init__.py
```diff
@@ -19,7 +20,12 @@
 from analytics.core.models import Job
 
 
+UNNECESSARY_JOB_REGEX = re.compile(r"No need to rebuild [^,]+, found hash match")
+
+
 def create_job(gl: gitlab.Gitlab, project: Project, gljob: ProjectJob) -> Job:
+    job_trace: str = gljob.trace().decode()
```
scottwittenburg, line 27: Just curious if this is possibly a second or more retrieval of the job trace resulting from the web hook firing.  And if so, is that anything we should be concerned with?

scottwittenburg:
> And @scottwittenburg to see how it's done.

Thanks for the heads up, I like how simple it is to do something like this now!

analytics/analytics/job_processor/__init__.py
danlamanna: It is. I'm not super worried, but we can always consolidate jobs in the future if it becomes an issue.

FWIW, this is a rough distribution of our job log size (91% of logs are between 0-1MB):
```
0                   58,606     [91.0%] ██████████████████████████████████████████████████
1                   2,476      [ 3.8%] ██▏
2                   1,062      [ 1.6%] █
3                   819        [ 1.3%] ▊
4                   240        [ 0.4%] ▏
5                   73         [ 0.1%]
6                   66         [ 0.1%]
7                   41         [ 0.1%]
8                   78         [ 0.1%]
9                   125        [ 0.2%]
10                  823        [ 1.3%] ▊
```

scottwittenburg: Ok, perfect, thanks for the details :+1:

jjnesbitt:
Obligatory reminder that we need to bump the image version

mvandenburgh (dismissed):
Just need to bump the image version for the `Deployments` in k8s/production/custom :upside_down_face:

PR #751, opened 2024-01-29
Rewrite migrations to fix table renames

danlamanna:
This is some migration surgery to hopefully fix what was missed in #744.


mvandenburgh (approved):
This seems to fix the migration error I was seeing locally :+1:

PR #754, opened 2024-02-01
Fix helper_memory_request units

zackgalbreath:
megabytes not millibytes


danlamanna (approved):
Do we have an idea of what it's set to now?

zackgalbreath:
> Do we have an idea of ~what it's set to now~ what it's defaulting to now?

Describing a running pipeline pod shows the current faulty value in millibytes:

```
Init Containers:
  init-permissions:
    Requests:
      memory:     512m
```

PR #756, opened 2024-02-01
Increase the job log limit to 20MB

danlamanna:
This should allow larger job logs to get through and let us retrieve more detail about what went wrong. Assuming there's no stability issues after rollout, we might want to ask 3rd party runners to do the same.


scottwittenburg (approved):
Yeah, I think we can do this safely.  We don't have that many jobs that hit this limit, so I don't think this doubles the amount of storage required for job traces.  Pretty far from it, I'm guessing.

But as you say, we can keep an eye on stability and either cut it back later, or else make the change on external runners.

PR #757, opened 2024-02-08
Enhance retry metadata tracking

danlamanna:
This removes the `retried` field and adds the following fields to opensearch:
```
    is_retry: bool
    is_manual_retry: bool
    attempt_number: int
    final_attempt: bool
```

Computing these is a little trickier than it ought to be but this should account for less common situations e.g.
1. Jobs with unique retry conditions unlike the regular rebuild job (like protected publish)
2. Multiple pipelines (`commit_id`s) executing the same named job multiple times
3. Webhooks being received/processed out of order

@mvandenburgh I know this is going to go in postgres soon but I wanted to get the logic committed so it can be ported when that happens.

### DIFF
--- a/analytics/analytics/core/job_failure_classifier/__init__.py
+++ b/analytics/analytics/core/job_failure_classifier/__init__.py
@@ -1,3 +1,4 @@
+from dataclasses import asdict, dataclass
 from datetime import datetime
 import json
 from pathlib import Path
@@ -9,6 +10,7 @@ from django.conf import settings
 from django.db import connections
 import gitlab
 import opensearch_dsl
+import sentry_sdk
 import yaml
 from opensearchpy import ConnectionTimeout
 from urllib3.exceptions import ReadTimeoutError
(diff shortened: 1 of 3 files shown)


analytics/analytics/core/job_failure_classifier/__init__.py
```diff
@@ -132,13 +177,18 @@ def upload_job_failure_classification(job_input_data_json: str) -> None:
...
+    try:
+        retry_info = _job_retry_data(
+            job_id=job_input_data["build_id"],
+            job_name=job_input_data["build_name"],
+            job_commit_id=job_input_data["commit"]["id"],
+            job_failure_reason=job_input_data["build_failure_reason"],
+        )
+        job_input_data.update(asdict(retry_info))
+    except Exception:
+        sentry_sdk.capture_exception()
```
danlamanna, line 191: I tested this locally but this is just out of an abundance of caution for the initial deployment. Assuming it's not triggered within the first few days we can remove it.

mvandenburgh (approved):

mvandenburgh (approved):

PR #767, opened 2024-02-13
Add timing information for GitLab CI sections

danlamanna:
This is for dumping the timing information for sections of the GitLab CI run that can be grabbed from the trace, an example is:
https://gitlab.spack.io/spack/spack/-/jobs/10033515
```
{'resolve_secrets': 0,
 'prepare_executor': 0,
 'prepare_script': 228,
 'get_sources': 43,
 'download_artifacts': 64,
 'step_script': 141,
 'after_script': 92,
 'upload_artifacts_on_success': 1,
 'cleanup_file_variables': 1}
```

Of particular note is probably the `download_artifacts` stage, which could prove useful for comparing aws/uo runners and cloudfront/s3 implementations.

The format is probably suboptimal - note that it uses `JobAttempt` instead of `Job` since that doesn't store non-rebuild jobs (for now?). Either way, I think it makes sense to start tracking this in some format before we lose the information due to lifecycle rules.


analytics/analytics/core/models.py
```diff
@@ -72,6 +72,15 @@ class JobAttempt(models.Model):
...
 
+class SectionTimer(models.Model):
+    """
+    A model for storing the timer values for each section in GitLab.
+    """
+
+    job_attempt = models.OneToOneField(JobAttempt, on_delete=models.CASCADE)
+    timers = models.JSONField(default=dict)
+
+
```
jjnesbitt, line 83: Would it make more sense to store this as a possibly nullable `timers` field on `JobAttempt`? Or is there a more concrete reason to have this as a separate model?

jjnesbitt (approved):

PR #773, opened 2024-02-29
Fix Build Timing Dashboard

zackgalbreath:
The tables used by this dashboard were renamed from `analytics_*` to `core_*`.


danlamanna (approved):
My mistake, thanks for catching this.

PR #780, opened 2024-03-19
Update terraform state with reality, enable RDS performance insights

mvandenburgh:
Some AWS resources got changed out from under Terraform, some due to AWS upgrades and some due to some manual actions in the AWS console UI, so this is just bringing the Terraform state up to date with reality.

One exception to this is enabling performance insights for each RDS instance. I had enabled it a while back for the prod gitlab instance for debugging, and I think it's worth enabling for all of our instances since it's free (at least for the default retention rate of 7 days), and might be helpful someday.

### DIFF
--- a/terraform/modules/spack/analytics_db.tf
+++ b/terraform/modules/spack/analytics_db.tf
@@ -34,6 +34,7 @@ module "analytics_db" {
   backup_window                   = "03:00-06:00"
   enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
   create_cloudwatch_log_group     = true
+  performance_insights_enabled    = true
 
   backup_retention_period = 7
   skip_final_snapshot     = false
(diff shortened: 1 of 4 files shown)


jjnesbitt (approved):

jjnesbitt:
My only hesitation is the database version upgrades, but they're pretty small bumps, so I'm not too concerned.

mvandenburgh:
> My only hesitation is the database version upgrades, but they're pretty small bumps, so I'm not too concerned.

Actually, the upgrades have already happened, AWS does them occasionally for us. This is just updating the version encoded in Terraform; otherwise, running `terraform apply` results in a plan to downgrade the DB versions.

jjnesbitt:
> Actually, the upgrades have already happened, AWS does them occasionally for us. This is just updating the version encoded in Terraform; otherwise, running `terraform apply` results in a plan to downgrade the DB versions.

Right. Somehow I forgot that even though it's the first sentence in the PR description :grimacing:

PR #781, opened 2024-03-19
Add github workflow for linting + validating TF correctness

mvandenburgh:
This should help us catch Terraform mistakes before anything gets merged.


danlamanna (approved):

jjnesbitt (commented):
This all looks good to me. I see we're running lint and validate, I wonder if it's worth going a step further and trying to do a plan or dry run? That way, if something is incorrect that a simple validate wouldn't catch, we can catch that here.

danlamanna:
To tack on to the future requests - if we could get a periodic plan with some form of alerting that would give us makeshift drift-detection and prevent the problem from earlier today.

mvandenburgh:
> This all looks good to me. I see we're running lint and validate, I wonder if it's worth going a step further and trying to do a plan or dry run? That way, if something is incorrect that a simple validate wouldn't catch, we can catch that here.

We'd need to configure AWS credentials to do that (which is not difficult, we already do it in the pcluster action), but i think that would be a good thing to have. I'll make a follow up PR with that + a cron job that runs it once a day or something

PR #794, opened 2024-04-01
Add timeout handling to pre_build.py

danlamanna:
This should handle the (fairly rare) case of requests timing out e.g. https://gitlab.spack.io/spack/spack/-/jobs/10466875.

### DIFF
--- a/scripts/gitlab_runner_pre_build/pre_build.py
+++ b/scripts/gitlab_runner_pre_build/pre_build.py
@@ -68,18 +68,19 @@ def _durable_assume_role_request(assume_role_kwargs):
                     raise ValueError(
                         f"Unexpected response code from assume role request: {response.getcode()}"
                     )
-        except urllib.error.HTTPError as e:
-            print(e.read().decode("utf-8"), file=sys.stderr)
+        except urllib.error.URLError as e:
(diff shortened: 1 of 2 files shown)


scripts/gitlab_runner_pre_build/pre_build.py
```diff
@@ -68,19 +68,20 @@ def _durable_assume_role_request(assume_role_kwargs):
...
-                )
-                time.sleep(2 ** (1 + attempts))
-            else:
+        except urllib.error.URLError as e:
+            should_retry = type(e) is urllib.error.URLError or isinstance(e, urllib.error.HTTPError) and e.code == 400
+
+            if not should_retry:
                 raise e
 
+            fail_reason = "HTTP 400" if isinstance(e, urllib.error.HTTPError) and e.code == 400 else e.reason
```
kwryankrattiger, line 77: The job you linked doesn't seem to have a 400 error, it was 110 if I read it correctly but maybe I am misreading the trace.

danlamanna: This is just a message for the log, the logic for changing when we retry is in the `except` clause. tl;dr is I caught `HTTPError` but not the parent `URLError` which catches timeouts.

mvandenburgh (approved):

PR #796, opened 2024-04-04
Prioritize execution_timeout over 5xx

danlamanna:
I'm not sure if this is the perfect place, but it improves jobs like https://gitlab.spack.io/spack/spack/-/jobs/10615522 which get categorized as 5xx.


scottwittenburg:
Yeah, the order makes sense given the linked job, thanks @danlamanna!

As you mentioned elsewhere, this still needs the image tag versions updated, but I'm happy to approve after that.

mvandenburgh (approved):

PR #798, opened 2024-04-11
Fix typo in classifier name

kwryankrattiger:
I think `rcp` was meant to be `rpc`


mvandenburgh (approved):

mvandenburgh (changes requested):
You also have to bump the `django` image version.

PR #816, opened 2024-04-16
Add retries/exp. backoff to TF init

mvandenburgh:
Hopefully this cuts down on the amount of false positives we get from the drift detection job due to transient network failures on the `terraform init` step.


mvandenburgh:
After some testing, it seems the `retry` util doesn't provide exponential backoff, but just a static backoff of whatever amount of seconds you give in between retries. Still, I imagine that's sufficient for our case.

jjnesbitt:
Would it be more concise to use something like https://github.com/marketplace/actions/retry-step?

mvandenburgh:
Oh nice, didn't even think of locking for an action that does it, I'll switch to that.

jjnesbitt (approved):
I think this looks good. If we see false alarms in the plan stage as well, we may need to adjust this.

PR #819, opened 2024-04-16
Add `c5.4xlarge` instance type to `gitlab` provisioner

mvandenburgh:
This is the instance type recommended by gitlab for the webservice. https://docs.gitlab.com/ee/administration/reference_architectures/3k_users.html#cluster-topology

It seems Karpenter is having trouble scheduling the webservice pods after #817 . Giving it this instance type should fix that.

### DIFF
--- a/k8s/production/karpenter/provisioners/gitlab/provisioner.yaml
+++ b/k8s/production/karpenter/provisioners/gitlab/provisioner.yaml
@@ -14,7 +14,11 @@ spec:
   requirements:
     - key: "node.kubernetes.io/instance-type"
       operator: In
-      values: ["t3.xlarge", "m5.xlarge", "m5.4xlarge"]
+      values:
+        - "t3.xlarge"
+        - "m5.xlarge"
+        - "m5.4xlarge"
+        - "c5.4xlarge"  # recommended by gitlab - https://docs.gitlab.com/ee/administration/reference_architectures/3k_users.html#cluster-topology
 
     # Always use on-demand
     - key: "karpenter.sh/capacity-type"


jjnesbitt (approved):

mvandenburgh:
:(
```
incompatible with provisioner "gitlab", 
daemonset overhead={"cpu":"315m","memory":"825829120","pods":"7"}, 
no instance type satisfied resources {"cpu":"16415m","memory":"15325829120","pods":"8"}
and requirements karpenter.sh/capacity-type In [on-demand], karpenter.sh/provisioner-name In [gitlab], kubernetes.io/arch In [amd64],
kubernetes.io/os In [linux], node.kubernetes.io/instance-type In [c5.4xlarge m5.4xlarge m5.xlarge t3.xlarge],
spack.io/node-pool In [gitlab] (no instance type which had enough resources and the required offering met the scheduling requirements); 
```

So we actually need an instance type with 16.4 vCPUs and at least 15.3 GB of memory. I think that means we need to go up to a 32 vCPU instance type.

jjnesbitt:
> So we actually need an instance type with 16.4 vCPUs and at least 15.3 GB of memory. I think that means we need to go up to a 32 vCPU instance type.

Could we not just artificially lower the resource requests to slightly under that of what gitlab is recommending?

mvandenburgh:
> > So we actually need an instance type with 16.4 vCPUs and at least 15.3 GB of memory. I think that means we need to go up to a 32 vCPU instance type.
> 
> Could we not just artificially lower the resource requests to slightly under that of what gitlab is recommending?

The additional 0.4 is needed for daemonsets. It would work if we were to lower it to exactly 15.6 vCPUs, but we would then have to mindful of this when adding any new daemonsets to the cluster. imo just using a slightly more expensive instance is preferable to the increased maintenance cost that would introduce (and hopefully the "more expensive" part doesn't even matter, since karpenter can fill the remaining cpu/memory on that instance with other gitlab pods).

PR #827, opened 2024-04-24
Add `spack-gantry` S3 bucket/ServiceAccount

mvandenburgh:
Adds an S3 bucket, IAM role/policy, and associated k8s ServiceAccount for #826. I also reworked the `litestream-config.yaml` file into a ConfigMap managed by Terraform so we don't have to hardcode the bucket name.


alecbcs:
Looks good to me!

PR #828, opened 2024-04-26
New cron job to delete stale branches from GitLab

zackgalbreath:


images/gitlab-delete-stale-branches/delete_stale_branches.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+if __name__ == "__main__":
+    if "GITLAB_TOKEN" not in os.environ:
+        raise Exception("GITLAB_TOKEN environment is not set")
+    try:
+        delete_stale_branches()
+    except Exception as inst:
+        print("Caught unhandled exception:")
+        print(inst)
```
mvandenburgh, line 109:
```suggestion
    delete_stale_branches()
```

Catching all exceptions here will result in it not being reported to sentry.

zackgalbreath: Good call. What do you think is the better thing to do here?
1. don't attempt to handle exceptions at all here
2. catch, report, re-raise
3. something else entirely?

mvandenburgh: I would vote for 1, so the original exception/stack trace is preserved in Sentry.

scottwittenburg: I think option two, but using a `raise ... from` (see [PEP 3134](https://legacy.python.org/dev/peps/pep-3134/)) would support reporting and preserve the exception for sentry, wouldn't it?

mvandenburgh: I believe that would work too. Although either way, the exception would be both logged to stderr and sent to sentry, right? Is there an advantage to explicitly catching it, logging it with `print()`, and then re-raising it?

images/gitlab-delete-stale-branches/delete_stale_branches.py
```diff
@@ -0,0 +1,106 @@
...
+def paginate(query_url):
+    """Helper method to get all pages of paginated query results"""
+    results = []
+
+    while query_url:
+        resp = requests.get(query_url, headers=AUTH_HEADER)
+
+        if resp.status_code == 401:
+            print(" !!! Unauthorized to make request, check GITLAB_TOKEN !!!")
+            return []
```
mvandenburgh, line 34: I would also suggest raising an exception here instead of silently failing with an empty list -
```suggestion
        resp = requests.get(query_url, headers=AUTH_HEADER)

        if resp.status_code == 401:
            print(" !!! Unauthorized to make request, check GITLAB_TOKEN !!!")
            resp.raise_for_status()
```

That would give us a sentry alert if the gitlab token doesn't work (if it expires, for example). Basically I think we should want these internal services to fail as loudly as possible so that we can find out as soon as possible. I could see a scenario where it starts to fail and no one notices for a while because you'd need to be checking the pod logs (or notice the branches on gitlab growing).

k8s/production/custom/delete-stale-branches/cron-jobs.yaml
```diff
@@ -0,0 +1,33 @@
...
+        spec:
+          restartPolicy: Never
+          containers:
+          - name: delete-stale-branches
+            image: ghcr.io/spack/gitlab-delete-stale-branches:0.0.1
+            imagePullPolicy: IfNotPresent
+            resources:
+              requests:
+                cpu: 500m
+                memory: 50M
```
jjnesbitt, line 22: This memory value seems pretty small, I could see it using more than this easily. Maybe something like 200M would be better? What do you think?

jjnesbitt: For the record I know it's not a limit, but this is more of a scheduling concern (making sure there's enough memory for all containers on a given node).

images/gitlab-delete-stale-branches/delete_stale_branches.py
```diff
@@ -0,0 +1,106 @@
...
+def paginate(query_url):
+    """Helper method to get all pages of paginated query results"""
+    results = []
+
+    while query_url:
+        resp = requests.get(query_url, headers=AUTH_HEADER)
+
+        if resp.status_code == 401:
+            print(" !!! Unauthorized to make request, check GITLAB_TOKEN !!!")
+            resp.raise_for_status()
```
jjnesbitt, line 34: Should we just call `raise_for_status` here unconditionally? Any `4xx` or `5xx` response would be invalid and not processable.  

```suggestion
        resp = requests.get(query_url, headers=AUTH_HEADER)
        if resp.status_code == 401:
            print(" !!! Unauthorized to make request, check GITLAB_TOKEN !!!")

        resp.raise_for_status()

```

mvandenburgh (approved):

PR #836, opened 2024-05-08
Add precondition for expired gitlab token

mvandenburgh:
Something useful I learned about Terraform recently. This will cause any plans or applys to fail if the token is expired.

### DIFF
--- a/terraform/modules/spack/gitlab_webhooks.tf
+++ b/terraform/modules/spack/gitlab_webhooks.tf
@@ -22,13 +22,23 @@ data "gitlab_user" "spackbot" {
   username = "spackbot"
 }
 
+locals {
+  webhook_handler_token_expires_at = "2024-12-03"
+}
+
 resource "gitlab_personal_access_token" "webhook_handler" {
-  user_id = data.gitlab_user.spackbot.id
-  name    = "Webhook handler token"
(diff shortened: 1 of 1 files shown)


terraform/modules/spack/gitlab_webhooks.tf
```diff
@@ -22,13 +22,23 @@ data "gitlab_user" "spackbot" {
...
+  expires_at = local.webhook_handler_token_expires_at
 
   scopes = ["read_api", "read_repository"]
+
+  lifecycle {
+    precondition {
+      condition     = timecmp(timestamp(), "${local.webhook_handler_token_expires_at}T00:00:00Z") == -1
+      error_message = "The token has expired. Please update the expires_at date."
+    }
+  }
```
jjnesbitt, line 41: This is great, although I'm concerned at what happens when encounter this expiration date. The terraform plan/apply will fail loudy about it, but besides that how would we know? Let's say if we haven't run terraform plan/apply for a while during the expiration date, would we just be relying on sentry at that point?

This is sort of secondary to this PR, which I think looks good itself, but just thought I'd raise this question.

jjnesbitt (approved):

mvandenburgh:
Superseded by #1104

PR #841, opened 2024-05-09
Add protected/signing runners for staging

mvandenburgh:


jjnesbitt:
This all looks good but is there any concern with having _all_ of the runner variants in staging? Could we just do the most general (graviton2, x86_64_v2)?

mvandenburgh:
> This all looks good but is there any concern with having _all_ of the runner variants in staging? Could we just do the most general (graviton2, x86_64_v2)?

We already have all the variants for public runners, so I don't think it's a problem. iirc we used to limit it as you suggested but were running into issues running some pipelines in staging

jjnesbitt (approved):

PR #843, opened 2024-05-14
Enforce exit on failure of native commands on Windows

johnwparent:
Native commands don't cause powershell to stop execution on failure because powershell's exception handling does not consider native command exit codes. 
Add deliberate handling for these cases (pretty much just the calls to the prebuild script).


johnwparent:
attn @mvandenburgh @kwryankrattiger

mvandenburgh (approved):

PR #844, opened 2024-05-15
Increase memory request/limit for gitlab webservice

mvandenburgh:
As of #838, we request 7GB for each gitlab webservice pod. We've been seeing sporadic downtime of a few seconds on gitlab.spack.io for the last week, and it looks like memory usage has been at about 14 GB for each gitlab webservice pod that period of time
https://grafana.spack.io/d/d249cc23-6930-403e-91f6-c9c30389b88d/kubernetes-deployment-cpu-and-memory-metrics

I verified that Karpenter can spin up necessary nodes to accomodate this without adding new instance types.

### DIFF
--- a/k8s/production/gitlab/release.yaml
+++ b/k8s/production/gitlab/release.yaml
@@ -143,10 +143,10 @@ spec:
           # https://docs.gitlab.com/ee/administration/reference_architectures/3k_users.html#cluster-topology
           limits:
             cpu: 12
-            memory: 7G
+            memory: 20G
           requests:
             cpu: 4
-            memory: 5G
+            memory: 14G
         nodeSelector:
           spack.io/node-pool: gitlab
       gitlab-exporter:


zackgalbreath:
Thanks for the link to the relevant grafana dashboard, @mvandenburgh 

Screenshot 2024-05-15 at 10 22 31 AM

This is for the whole deployment, right?  i.e. the sum of the usage for our four webservice pods? If so, then the updated memory request of 14G seems right to me.

Should we be concerned that this memory usage seems to be gradually increasing over time?

mvandenburgh:
> This is for the whole deployment, right? i.e. the sum of the usage for our four webservice pods? If so, then the updated memory request of 14G seems right to me.

Yes. You can also drill down into each of the 4 individual pods underneath that one.

> Should we be concerned that this memory usage seems to be gradually increasing over time?

Yeah I think it's worth keeping an eye on/investigating, not sure what the cause of that could be yet.

zackgalbreath (approved):

PR #845, opened 2024-05-15
Cron job to start rebuild everything pipelines

kwryankrattiger:
Cron job that starts rebuild everything on a new branch dedicated to rebuilding everything. The branch gets tagged with a date stamp for tracking purposes.

I think we can add these to the list of pruning activities we do with snapshot pruning. 

Pre-requisite spack CI change, https://github.com/spack/spack/pull/44200

cc: @scottwittenburg


images/git-remote-sync/sync-git-remotes.py
```diff
@@ -0,0 +1,116 @@
...
+    subprocess.run(["git", "remote", "add", "origin", pull_remote], check=True)
+    subprocess.run(["git", "remote", "set-url", "--push", "origin", push_remote], check=True)
+
+    fetch_args = ["git", "fetch", "-q", "origin", branch]
+    subprocess.run(fetch_args, check=True, stdout=subprocess.PIPE).stdout
+    subprocess.run(["git", "checkout", branch], check=True)
+
+
+@contextmanager
+def chworkingdir(path):
```
scottwittenburg, line 72: Did you look at the [one](https://docs.python.org/3/library/contextlib.html#contextlib.chdir) provided by `contextlib`?  Maybe you don't need to write your own.

images/git-remote-sync/sync-git-remotes.py
```diff
@@ -0,0 +1,116 @@
...
+    ssh_key_base64 = os.getenv("GITLAB_SSH_KEY_BASE64")
+    if ssh_key_base64 is None:
+        raise Exception("GITLAB_SSH_KEY_BASE64 environment is not set")
+    setup_ssh(ssh_key_base64)
+
+    with chworkingdir(args.working_dir):
+        setup_repo(args.pull_remote, args.push_remote, args.pull_branch)
+
+        if args.tag:
+            subprocess.run(["git", "tag", args.push_branch], check=True)
```
scottwittenburg, line 111: I think you may need an annotated tag object.  I feel like Zack ran into something like that when pushing the develop snapshot tags.

k8s/production/custom/rebuild-everything/sealed-secrets.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  name: rebuild-everything
+  namespace: custom
+spec:
+  encryptedData:
+    # This token must have repo:status permissions
```
scottwittenburg, line 8: Is this comment accurate?  Isn't this supposed to be an ssh key?

scottwittenburg (commented):
Thanks for taking a stab at this @kwryankrattiger!  Just curious if you saw the similar thing @zackgalbreath did [here](https://github.com/spack/spack-infrastructure/blob/main/images/snapshot-release-tags/snapshot_release_tags.py), which does much the same, but seems a little simpler.  I know in this case you're pushing directly to gitlab, so you'd have to verify whether a similar api is available in the `python-gitlab` package.

images/git-remote-sync/sync-git-remotes.py
kwryankrattiger: It would have to be something like `with tempdir.TemporaryDirectory() as workingdir, contextmanager.chdir(workingdir)` which I didn't like. The one I have lets be choose whether to chdir to a temp dir or a persistent dir which was nice to test this locally.

images/git-remote-sync/sync-git-remotes.py
kwryankrattiger: It is here for potentially changing this to tag develop and push the new tag on the currently running develop, or push a new develop. I am not sure what is better.

k8s/production/custom/rebuild-everything/sealed-secrets.yaml
kwryankrattiger: ssh key registered with the gitlab API with that access level

kwryankrattiger: I stole the ssh stuff from the `gh-gl-sync`. I am basically doing the same thing, but I only want to do one tiny part and I also don't want to add more to that script if I don't need to.

kwryankrattiger:
I decided not to use the API because I wanted to try pushing the latest develop on github and it seemed simpler to me to just use git commands directly. I was back and forth on making this a bash script, but it was easier to just use some of the existing bits from `gh-gl-sync`. That is where I got the repo setup for spackbot running and the ssh key registration.

images/git-remote-sync/sync-git-remotes.py
scottwittenburg: I did the something similar for testing protected publish locally, [this](https://github.com/scottwittenburg/spack-infrastructure/blob/protected-publish/images/protected-publish/publish.py#L444-L448) is how I avoided forcing use of a temporary directory.

images/git-remote-sync/sync-git-remotes.py
scottwittenburg: I'm not sure I know what you mean.  I'm was just wondering if the `git tag` needed a `-a`.

kwryankrattiger: oh, this isn't for a release or anything so I don't think there is a need to add any annotations. These also only get pushed to gitlab, so they are opaque to the world.

images/git-remote-sync/sync-git-remotes.py
kwryankrattiger: I see, the way yield works is still a bit of magic to me, I didn't know I could construct it, return the yielded result, and use it in a with clause. 🪄

kwryankrattiger:
> Just curious if you saw the similar thing @zackgalbreath did [here](https://github.com/spack/spack-infrastructure/blob/main/images/snapshot-release-tags/snapshot_release_tags.py),

I didn't see that, but I used the gitlab API for the pruner. I didn't for this because I didn't need the pagination or API requests and this works no matter where the remotes are located since it just takes source, dest, from branch, to branch/tag. Not direct deps on gh or gl.

PR #848, opened 2024-05-21
Publish stacks within a ref to the top-level

scottwittenburg:
This PR adds a container image with tooling to support publishing binary mirrors from the stack-specific mirrors in which pipelines build them, to a top-level mirror, for easier consumption by end users.  See the `README.md` for complete details.  

The goal here is to have an idempotent way to publish binary packages from gitlab pipelines outside of the gitlab pipeline workflow.  This will speed up protected pipeline turnaround time, since the current `protected-publish` job can take a long time, depending on the number of specs in the top-level mirror.


k8s/production/custom/protected-publish/cron-jobs.yaml
```diff
@@ -0,0 +1,52 @@
...
+                mountPath: "/tmp"
+            env:
+              # Tell tempfile.mkdtemp to use the ephemeral storage
+              #     https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp
+              - name: TMPDIR
+                value: "/tmp"
+          volumes:
+            - name: ephemeral
+              emptyDir:
+                sizeLimit: 50G
```
mvandenburgh, line 50: This part looks correct to me.

images/protected-publish/validate_index.py
```diff
@@ -0,0 +1,120 @@
...
+import io
+import json
+import re
+import sys
+from datetime import datetime
+
+import botocore.exceptions
+import boto3.session
+
+
```
mvandenburgh, line 11: Should we also include sentry in this file? Or, is this never run in the cluster?

scottwittenburg: I'll add it back, since as you pointed out, it doesn't hurt to have it in there, even if there is no cronjob defined to run it.

k8s/production/custom/protected-publish/cron-jobs.yaml
scottwittenburg: Thanks!

scottwittenburg:
fyi @jjnesbitt

images/protected-publish/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+When pipeline generation runs, it produces a manifest of everything new the pipeline will build, and this was provided via artifacts to the `protected-publish` job, so that it could only copy the new things to the top level.  One problem with the previous approach is that holes were produced in the top level mirror when any of the child pipelines failed, and thus the `protected-publish` job failed to run.  Information about exactly what was in those "holes" was scattered around in the artifacts of skipped `protected-publish` jobs.
+
+Another problem with the previous approach is that it added too much time to each protected pipeline.  Rebuilding a mirror index can take a long time, not to mention the time to copy potentiall large files from one prefix to another.  This resulted in hours extra delay before another protected pipeline could run on the same ref.
```
zackgalbreath, line 90:
```suggestion
Another problem with the previous approach is that it added too much time to each protected pipeline.  Rebuilding a mirror index can take a long time, not to mention the time to copy potentially large files from one prefix to another.  This resulted in hours extra delay before another protected pipeline could run on the same ref.
```

k8s/production/custom/protected-publish/cron-jobs.yaml
```diff
@@ -0,0 +1,49 @@
...
+          nodeSelector:
+            spack.io/node-pool: base
```
zackgalbreath, line 49: Noting here that currently the biggest instance in our base pool is t3.medium which has 2 vCPU and 4 GiB of memory. I think we'll be fine but it's never quite clear to me ahead of time how much overhead kubelet claims for itself.

zackgalbreath (commented):
Looks great to me!

zackgalbreath (approved):

images/protected-publish/validate_index.py
```diff
@@ -0,0 +1,123 @@
+import argparse
+import io
+import json
+import re
+import sys
+from datetime import datetime
+
+import botocore.exceptions
+import boto3.session
+import sentry_sdk
```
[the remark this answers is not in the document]
scottwittenburg: I think just install deps from the requirements file, and there's no need to try/except.

k8s/production/custom/protected-publish/cron-jobs.yaml
scottwittenburg: Yeah, I agree.  I picked `1500m` because I noticed that spackbot worker picks the same when using the base node pool.  :crossed_fingers:

images/protected-publish/validate_index.py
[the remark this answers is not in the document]
kwryankrattiger: I think the sentry stuff failed for me locally even with the module. But I can just comment it out to test locally nbd.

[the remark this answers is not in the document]
scottwittenburg: I ran both python scripts in the container locally many times, I never encountered an error.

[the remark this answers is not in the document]
kwryankrattiger: awesome, I know you said you did quite a bit, just something I kept running across in my testing which was annoying. I hope I never have to run this manually, if so something has gone terribly wrong 🙃

PR #851, opened 2024-05-22
Add S3 lifecycle rule for non-current versions

scottwittenburg:
Following this [example](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration#creating-a-lifecycle-configuration-for-a-bucket-with-versioning), encode our existing lifecycle rule expiring objects 30 days after they become non-current.


scottwittenburg:
@mvandenburgh It's likely this isn't quite right yet, but hopefully serves to remind us to eventually encode the lifecycle rule I set up a couple weeks ago.

PR #853, opened 2024-05-22
Prune old develop snapshots

scottwittenburg:
Old develop snapshots have been accumulating for about a year now, this PR provides a cronjob to clean up all but the most recent few on a periodic basis.

### DIFF
--- a/images/snapshot-release-tags/prune_snapshots.py
+++ b/images/snapshot-release-tags/prune_snapshots.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import re
+import subprocess
+
+import sentry_sdk
+from github import Github
+
+
+sentry_sdk.init(
+    # This cron job only runs once weekly,
+    # so just record all transactions.
+    traces_sample_rate=1.0,
+)
+
+TAG_REF_REGEX = re.compile(r"^refs/tags/(develop-\d{4}-\d{2}-\d{2})$")
+
+
+def main():
+    if "GITHUB_TOKEN" not in os.environ:
+        raise Exception("GITHUB_TOKEN environment is not set")
+
+    parser = argparse.ArgumentParser(
+        prog="prune_snapshots.py",
+        description="Prune expired snapshots",
+    )
+
+    parser.add_argument(
+        "-k",
+        "--keep-last-n",
+        type=int,
+        default=8,
+        help="Prune all but most recent --keep-last-n",
+    )
+    parser.add_argument(
+        "-m",
+        "--mirror-root",
+        default="s3://spack-binaries",
+        help=("Root url of mirror where snapshot binaries are mirrored"),
+    )
+
+    args = parser.parse_args()
+
+    keep_n = args.keep_last_n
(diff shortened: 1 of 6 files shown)


scottwittenburg:
@zackgalbreath @kwryankrattiger I know we discussed this several times and never really came to agreement about how and when this should be done to be minimally disruptive.  I'm just getting the ball rolling on it again with some concrete implementation we can poke at.

images/snapshot-release-tags/prune_snapshots.py
```diff
@@ -0,0 +1,85 @@
...
+            continue
+
+        mirror_prefix = m.group(1)
+        url_to_prune = f"{mirror_root_url}/{mirror_prefix}"
+
+        print(f"  Ref: {tag.ref}, Mirror: {url_to_prune}")
+
+        # First, try to delete the mirror associated with the snapshot
+        try:
+            subprocess.run(["aws", "s3", "rm", "--recursive", url_to_prune], check=True)
```
kwryankrattiger, line 75: An idea for a process for making making sure the cache.spack.io and the mirrors are in sync.

check if mirror has `index.json`
- if yes -> remove `index.json`
- if no -> delete the entire prefix

This will give a buffer time between dropping the tag and deleting the mirror contents. If we run this weekly that should translate to the cache.spack.io page being updated, it uses the the `index.json` to create a global index file.

The timeline for snapshot pruning could be:

-- Prune Cron Runs @ 2023/01/01 0100 UTC
- Delete Tag `develop-XXXX`
- Delete `develop-XXXX/build_cache/index.json`

-- Generate cache.spack.io @ 2023/01/02 0100 UTC
- Create global index
- push new website without mirrors containing no `index.json`

-- Prune Cron Runs @ 2023/01/08 0100 UTC
- Delete prefix `develop-XXXX/`

images/snapshot-release-tags/prune_snapshots.py
```diff
@@ -0,0 +1,85 @@
...
+    keep_n = args.keep_last_n
+    mirror_root_url = args.mirror_root
+
+    # Use the GitHub API to create a tag for this commit of develop.
+    github_token = os.environ.get("GITHUB_TOKEN")
+    py_github = Github(github_token)
+    py_gh_repo = py_github.get_repo("spack/spack", lazy=True)
+
+    # Get a list of all the tags matching the develop snapshot pattern
+    snapshot_tags = py_gh_repo.get_git_matching_refs("tags/develop-")
```
kwryankrattiger, line 55: The method described below would break this query for mirror names.

If we still want to use tags, we could move deleted tags to hidden refs, something like `refs/archive/develop-*`. The other option is for each prefix listed in the root, match the name to a regex `<prefix>/develop-*/` check for the `index.json` to see what to delete.

I am not sure which method to prefer for this, maybe use both, one for listing tags to remove, the other for deleting mirrors.

kwryankrattiger (commented):
I added my thought from the other day here.

Thanks for starting on this!

kwryankrattiger:
Possible change which would change how snapshot refs are queried

https://github.com/spack/spack-infrastructure/pull/867

PR #861, opened 2024-06-03
Use a persistent volume to store data for Grafana

zackgalbreath:
Documented here:
https://github.com/grafana/helm-charts/tree/main/charts/grafana


zackgalbreath:
I tested this out on the staging cluster and it appears to work as intended. Starring & modifying the "home dashboard" seems to stick across pod restarts.

mvandenburgh (approved):

PR #865, opened 2024-06-06
Add job spec hash as pod annotation

jjnesbitt:
This is needed for the upcoming analytics DB re-work and figured it made sense to split it out.


jjnesbitt:
For windows runners, this PR is technically adding this data as a pod label instead of a pod annotation, since other data that would be an annotation is also a label already on those runners (and would cause the job processor to fail anyway). I think it makes sense to create a follow up PR to bring those runners fully in line with the other runners.

zackgalbreath (approved):

PR #867, opened 2024-06-10
Snapshot hidden ref

kwryankrattiger:
Spack has some problems around snapshots

1. This script is pointing at the wrong repo after repo split
    - Github: `spack/spack` -> `spack/spack-packages`
    - Gitlab: 2 -> 57
2. The current method of pinning snapshots uses tags, which creates extra noise around the GH releases feature
    - Tags imply (or should imply) immutability
3. Spack Core Gitlab CI is unstable as it points at the head of packages develop, which may or may not be passing in packages
    - Needs to either manually update packages ref
    - Need a moving ref. (shouldn't be a tag)

Spack snapshots have historically been pushed as tags which can be noisy in Github since tags and releases are tied together. Alternatives to tag refs are either branches or hidden refs. While most people are more familiar with branches, a hidden ref will not inundate the branch search. For that reason, we decided to switch to hidden refs.

The Spack core repo can now point at `snapshots/develop-latest` for packages, which is guaranteed to have at least passed the packages CI check. The downside to this is extended periods of broken packages develop will end up delaying updates in core, this is worth it to ensure gitlab checks can be made required in core.

### DIFF
--- a/images/gh-gl-sync/SpackCIBridge.py
+++ b/images/gh-gl-sync/SpackCIBridge.py
@@ -353,6 +353,21 @@ class SpackCIBridge(object):
             print("    {0}".format(tag))
         return tags
 
+    def list_github_hidden_refs(self, hidden_refs):
+        if not hidden_refs:
+            return []
+
+        ref_list = []
+        for ref in hidden_refs:
+            ref_list.extend(self.py_gh_repo.get_matching_refs(ref))
+            print("Rate limit after get_matching_refs({}): {}".format(ref, self.py_github.rate_limiting[0]))
+
+        refs = sorted(set([ref.ref for ref in ref_list]))
+        print("Refs:")
+        for ref in refs:
+            print("    {0}".format(ref))
+        return refs
+
     def setup_git_repo(self):
         """Initialize a bare git repository with two remotes:
         one for GitHub and one for GitLab.
@@ -408,6 +423,12 @@ class SpackCIBridge(object):
         for tag in tags:
             fetch_refspecs.append("+refs/tags/{0}:refs/tags/{0}".format(tag))
             open_refspecs.append("refs/tags/{0}:refs/tags/{0}".format(tag))
+        return open_refspecs, fetch_refspecs
+
+    def update_refspecs_for_hidden_refs(self, hidden_refs, open_refspecs, fetch_refspecs):
+        for ref in hidden_refs:
+            fetch_refspecs.append("+{0}:{0}".format(ref))
+            open_refspecs.append("{0}:{0}".format(ref))
         return open_refspecs, fetch_refspecs
 
     def fetch_github_branches(self, fetch_refspecs):
@@ -651,7 +672,7 @@ class SpackCIBridge(object):
             print(e_inst)
         print("  {0} -> {1}".format(branch, sha))
 
-    def sync(self):
+    def sync(self, hidden_refs=None):
         """Synchronize pull requests from GitHub as branches on GitLab."""
 
         print("Initial rate limit: {}".format(self.py_github.rate_limiting[0]))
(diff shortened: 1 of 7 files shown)


images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -767,4 +795,4 @@ def sync(self):
                            prereq_checks=args.prereq_check,
                            required_label=args.required_label)
     bridge.setup_ssh(ssh_key_base64)
-    bridge.sync()
+    bridge.sync(hidden_refs=args.hidde_ref)
```
zackgalbreath:
```suggestion
    bridge.sync(hidden_refs=args.hidden_ref)
```

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -36,27 +40,39 @@
...
+
+    # Check if this sha is already the latest snapshot
+    py_gh_repo = py_github.get_repo("spack/spack-packages", lazy=True)
+    try:
+        latest_ref = py_gh_repo.get_git_ref("snapshots/develop-latest")
+        if latest_ref.object.sha == sha:
+            print("Latest ref is already latest snapshot")
+            sys.exit(0)
+    except GithubException:
+        # Failure to get the latest_ref means it doesn't exist and needs to be recreated.
```
zackgalbreath:
```suggestion
    except GithubException as e:
        if e.status != 404:
            raise
        # 404 means the ref doesn't exist yet and needs to be created.
```

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -36,27 +40,39 @@
...
+        latest_ref = None
+        pass
```
zackgalbreath:
```suggestion
```

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -36,27 +40,39 @@
...
-        sha=tag.sha)
+        ref=f"refs/snapshots/{ref_name}",
+        sha=sha)
+
+    # Create a ref for this sha using the `develop-latest` tag for the GH-GL sync script
+    print(f"snapshots/develop-latest: {sha}")
+    if latest_ref is not None:
+        latest_ref.edit(sha, force=True)
+    else:
+        py_gh_repo.create_git_ref(
```
zackgalbreath: It might make sense to catch "ref already exists" here (rather than letting it propagate to sentry) but reading [the docs](https://docs.github.com/en/rest/git/refs?apiVersion=2026-03-10#create-a-reference--status-codes) I'm not sure if that would be 409 or 422 (or if we would even be able to cleanly distinguish this type of error from others). WDYT?

zackgalbreath (changes requested):

zackgalbreath:
Would you please also update the description for this PR to describe the problem we're solving here?

images/snapshot-release-tags/snapshot_release_tags.py
kwryankrattiger: For this one we are updating the `snapshots/develop-latest` which, in most cases, we expect to exist. The issue is you need a different API call to update the ref vs push a new one.

I did miss above, where I am blindly updating the date ref, that we may snapshot multiple refs on the same day now. That I can fix by only running the cron once a day, but probably a good idea to make it a little more robust than that.

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -36,27 +74,15 @@
...
+    print(f"Gitlab pipeline: {sha}")
 
-    date_str = datetime.now(timezone.utc).strftime("%Y-%m-%d")
-    tag_name = f"develop-{date_str}"
-    tag_msg = f"Snapshot release {date_str}"
+    repo = py_github.get_repo("spack/spack-packages", lazy=True)
+    # If the latest develop snapshot is updated, also push a new date snapshot
+    if create_or_update_ref(repo, "snapshots/develop-latest", sha):
+        date_str = datetime.now(timezone.utc).strftime("%Y-%m-%d")
+        ref_name = f"develop-{date_str}"
```
zackgalbreath:
```suggestion
        ref_name = f"snapshots/develop-{date_str}"
```

zackgalbreath (changes requested):

zackgalbreath (approved):

PR #869, opened 2024-06-12
New cron job to restart gitlab-webservice once a day

zackgalbreath:
This will help mitigate the issue we've seen where the GitLab webservice consumes more and more memory as time goes on, which eventually results in spurious CI job failures.

The relevant upstream issue appears to be:
https://gitlab.com/gitlab-org/gitlab/-/issues/432847


zackgalbreath:
I manually verified that this seems to work as intended by `kubectl apply`'ing these new YAML files and then creating a one-off job using `kubectl create job --from=cronjob...`.

jjnesbitt (approved):
LGTM. I checked grafana and saw the restart which lines up with your testing of it, so I think we're good to go.

PR #870, opened 2024-06-12
Restructure analytics DB to use dimensional modeling

jjnesbitt:
This PR fundamentally changes the structure of our anayltics database to a dimensional modeling approach (optimized for OLAP), rather than a standard normalized structure (optimized for OLTP). [Here](https://www.holistics.io/books/setup-analytics/kimball-s-dimensional-data-modeling/) is an article that covers the basics of dimensional modeleing.

In this new schema, there are the following fact tables:

1. The `JobFact` table, which represents a single successful build job
2. The `TimerFact` table, which represents a single timer within a job (a single package dependency being installed)
3. The `TimerPhaseFact` table, which represents a single phase of a package being installed (e.g. fetch, install, etc.)

and the following dimensional tables:

1. The `DateDimension` table, which describes a date
2. The `TimeDimension` table, which describes a time of the day (to second precision)
3. The `JobDataDimension` table, which describes the details of a job (ref, commit, etc.)
4. The `NodeDimension` table, which describes the node a job ran on (empty if not run in the cluster)
5. The `RunnerDimension` table, which describes the gitlab runner a job ran on
6. The `PackageSpecDimension` table, which describes a specific package spec
7. The `PackageDimension` table, which describes a package by name only (used in places where the full package spec isn't available)
8. The `TimerPhaseDimension` table, which describes the various phases of a timer
9. The `TimerDataDimension` table, which at the moment only describes if a package was installed from cache or not.

The relation between these fact and dimensional tables is shown in the diagram below

spack_dimensional_model-Facts-To-Dimensions drawio

The existing models are not removed, but rather renamed with a `legacy` prefix. This is a precautionary measure, and once we feel confident enough, these can be permanently deleted.

This change touches a lot of the analytics db code, and a substantial re-work w.r.t the Prometheus and general job information gathering. There's also two large migrations (in raw SQL) that are required. I've tested these multiple times and confirmed everything works correctly. I've also tested the job processor itself on some real webhook data, confirming it also behaves correctly.


jjnesbitt:
Also, I know the commits are messy. I plan to squash and merge, to clean all of that up.

analytics/analytics/core/models/facts.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Pod and node info (null if not run in cluster)
+    pod_node_occupancy = models.FloatField(null=True, default=None)
+    pod_cpu_usage_seconds = models.FloatField(null=True, default=None)
+    pod_max_mem = models.PositiveBigIntegerField(null=True, default=None)
+    pod_avg_mem = models.PositiveBigIntegerField(null=True, default=None)
+    node_price_per_second = models.DecimalField(
+        max_digits=9,
+        decimal_places=8,
+        help_text="The price per second (in USD) of the spot instance this job ran on, at the time of running.",
```
danlamanna, line 44: :+1: 

FWIW, I think [Field.db_comment](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.db_comment) will add it into postgres and *might* translate to a metabase description field somehow (https://github.com/metabase/metabase/pull/6974).

analytics/dev/django.Dockerfile
```diff
@@ -16,5 +16,13 @@ ENV PYTHONUNBUFFERED 1
 COPY ./setup.py /opt/django-project/setup.py
 RUN pip install --editable /opt/django-project[dev]
 
+
+# Install spack
+RUN git clone -c feature.manyFiles=true https://github.com/spack/spack.git /opt/spack
```
mvandenburgh, line 21: Just wanted to point out, the version of `spack` used in the job processor will be whatever the status of `develop` is at build time of the docker image. Not sure if that matters or not, but figured it was worth mentioning.

mvandenburgh (approved):
Nothing objectionable from a high-level review of all this. I just made one comment on the potential for a state version of `spack` being set at image build time.

analytics/analytics/core/models/facts.py
jjnesbitt: I can confirm this is the case, thanks!

analytics/dev/django.Dockerfile
```diff
@@ -3,8 +3,17 @@ FROM python:3.10-slim
...
-    libpq-dev gcc libc6-dev git && \
-    rm -rf /var/lib/apt/lists/*
+    libpq-dev gcc libc6-dev git
+
+# Install spack dependencies
+RUN apt-get update && \
+    apt-get install --no-install-recommends --yes \
+    build-essential ca-certificates coreutils curl \
+    environment-modules gfortran git gpg lsb-release \
+    python3 python3-distutils python3-venv unzip zip
```
mvandenburgh, line 13: I think you can switch the base image to just `python:3.10` instead of `python:3.10-slim`, and `python3-distutils`/`python3-venv` will be automatically installed.

mvandenburgh (approved):

PR #875, opened 2024-06-18
Manage spackbot gitlab PAT with Terraform

mvandenburgh:
This PR encodes the gitlab personal access token used by spackbot in our Terraform configuration. It also uses a Terraform `lifecycle` hook to provide an alert when the token expires.


jjnesbitt (approved):
This all makes sense to me, with an added benefit of splitting up the github and gitlab credentials. Nice!

PR #886, opened 2024-06-24
Encode `delete-stale-branches` PAT in Terraform

mvandenburgh:
Similar to #877, this fixes the failing `delete-stale-branches` cronjob due to the token expiring.


zackgalbreath (approved):
LGTM if you're okay with my latest push

PR #892, opened 2024-06-26
Require at least two replicas for nginx controller

mvandenburgh:
Since there's some evidence that the nginx controller pod is being killed with a `SIGTERM` periodically, requiring at least two to be up at all times could help mitigate this issue.


jjnesbitt (approved):

mvandenburgh:
Will hold off on merging for now, as this might obscure the ssl timeout issue

PR #896, opened 2024-06-26
Speed up `spack` git clone in docker build

mvandenburgh:
Instead of cloning the full repo and then checking out a specific tag, we can use `--depth 1` to only clone the desired tag.


mvandenburgh:
I also tossed https://github.com/spack/spack-infrastructure/pull/896/commits/066c7166f28627906d5237e39156a6a8171e2b17 in to avoid another PR bumping the image agiain, I missed that in #895

jjnesbitt (approved):

PR #898, opened 2024-06-26
Retry celery tasks on gitlab timeouts

mvandenburgh:
Adds retries to error classification and process_job celery tasks on `requests.exceptions.ReadTimeout` exceptions. These happen when a request to gitlab times out.

### DIFF
--- a/analytics/analytics/job_processor/__init__.py
+++ b/analytics/analytics/job_processor/__init__.py
@@ -113,6 +113,7 @@ def create_job_fact(
 @shared_task(
     name="process_job",
     autoretry_for=(ReadTimeout,),
+    max_retries=3,
 )
 def process_job(job_input_data_json: str):
     # Read input data and extract params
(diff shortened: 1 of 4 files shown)


analytics/analytics/job_processor/__init__.py
```diff
@@ -109,7 +110,10 @@ def create_job_fact(
     )
 
 
-@shared_task(name="process_job")
+@shared_task(
+    name="process_job",
+    autoretry_for=(ReadTimeout,),
+)
```
jjnesbitt, line 116: While i think you're aware the default value is 3 for `max_retries`, it might be good to encode that here so everyone is aware:

```suggestion
@shared_task(
    name="process_job",
    autoretry_for=(ReadTimeout,),
    max_retries=3,
)
```

jjnesbitt (approved):

PR #901, opened 2024-06-27
Don't use _durable_subprocess_run for commands that are allowed to fail

zackgalbreath:


zackgalbreath:
I verified this change manually by running the sync script locally

mvandenburgh (approved):

PR #902, opened 2024-06-27
Re-adjust resource requests/limits for webhook handler

mvandenburgh:
Adjusts resource requests/limits for webhook Django app based on usage reported by Prometheus (https://grafana.spack.io/d/d249cc23-6930-403e-91f6-c9c30389b88d/kubernetes-deployment-cpu-and-memory-metrics?orgId=1&var-namespace=custom&var-deployment=webhook-handler-worker&var-pod=All&from=now-24h&to=now)

Underprovisioning seems to be responsible for some of the sentry errors we're seeing.

### DIFF
--- a/k8s/production/custom/webhook-handler/deployments.yaml
+++ b/k8s/production/custom/webhook-handler/deployments.yaml
@@ -27,11 +27,11 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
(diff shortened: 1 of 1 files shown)


jjnesbitt (approved):
We'll just want to keep an eye on costs as this will increase our base resource load in the cluster.

PR #903, opened 2024-06-27
Fix error handling in pruning script

kwryankrattiger:
Add option to fail fast for local testing/usage. By default the script will attempt all mirrors and write errors/failures to a file

Fix bug where error/failures files were not being opened for write.

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -42,7 +42,7 @@ jobs:
           - docker-image: ./analytics
             image-tags: ghcr.io/spack/django:0.3.5
           - docker-image: ./images/ci-prune-buildcache
-            image-tags: ghcr.io/spack/ci-prune-buildcache:0.0.3
+            image-tags: ghcr.io/spack/ci-prune-buildcache:0.0.4
           - docker-image: ./images/protected-publish
             image-tags: ghcr.io/spack/protected-publish:0.0.1
     steps:
--- a/images/ci-prune-buildcache/ci_buildcache_prune.py
+++ b/images/ci-prune-buildcache/ci_buildcache_prune.py
@@ -229,6 +229,12 @@ def configure_parser():
         help="use the buildcache index to check for buildcache hashes",
         action="store_true",
     )
+    parser.add_argument(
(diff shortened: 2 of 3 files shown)


images/ci-prune-buildcache/ci_buildcache_prune.py
```diff
@@ -495,7 +501,8 @@ def configure_parser():
         except Exception as e:
             print(f"Error -- Skipping pruning of {stack}")
             print(str(e))
-            raise "ff mode" from e
+            if args.fail_fast:
+                raise "fail fast" from e
```
scottwittenburg, line 505: This `raise` statement was in the error stack trace as well, it came up in handling of the first exception.  As mentioned there, you can't `raise` a string because it doesn't derive from `BaseException`.

kwryankrattiger: interesting, I was not having issue with this locally, maybe the from makes it work?

scottwittenburg: The `from` does not make it work. You must not have hit this code path locally, or else didn't notice the "During handling of the above exception..." bit.

.github/workflows/custom_docker_builds.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           - docker-image: ./analytics
             image-tags: ghcr.io/spack/django:0.3.5
           - docker-image: ./images/ci-prune-buildcache
-            image-tags: ghcr.io/spack/ci-prune-buildcache:0.0.3
+            image-tags: ghcr.io/spack/ci-prune-buildcache:0.0.4
```
scottwittenburg, line 45: Doesn't changing the tag here mean you need to change it in the `cronjobs.yaml` as well?

kwryankrattiger: oh shoot, yeah, thanks!

scottwittenburg (approved):
Seems like this could allow it to proceed after encountering the same error we saw recently, but I haven't tested.  So maybe check the kube logs shortly after the next time it runs.  Thanks @kwryankrattiger!

PR #911, opened 2024-07-08
Encode CI job retries in Terraform

mvandenburgh:
We recently set `GET_SOURCES_ATTEMPTS` to 3 in GitLab so that GitLab retries a failed clone of the `spack/spack` repo on gitlab.spack.io. This change encodes it so that it is managed by Terraform, and also enables some additional retrying with some other environment variables.


jjnesbitt (approved):
LGTM, only thing I might add is a link to the docs could be useful https://docs.gitlab.com/ee/ci/runners/configure_runners.html#job-stages-attempts

mvandenburgh:
Superseded by #954 (was included there)

PR #914, opened 2024-07-09
Add CI workflow to check for missing image version bumps

jjnesbitt:
(I accidentally closed #912 when playing around with git commands locally)

This PR adds a CI workflow that fails if it finds one of the following conditions:

1. A directory was modified, but the image version in `.github/workflows/custom_docker_builds.yml` was not incremented
2. A directory was modified, and the image version in the workflow file was bumped, but uses of the old image version still remain in the code base.

In cases where this check is either undesired or unhelpful, the `no-image-bump` tag can be applied to the pull request and the job will be skipped.

### DIFF
--- a/.github/scripts/check_docker_image_versions.sh
+++ b/.github/scripts/check_docker_image_versions.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+echoerr() { printf "%s\n" "$*" >&2; }
+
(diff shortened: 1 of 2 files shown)


jjnesbitt:
I've tested this to the best of my ability and it seems to do what we want. Technically it's not guaranteed to work if master is _ahead_ of this branch, but since we have branch protection rules on main that require PRs to be up to date before merging, I don't think this is a huge issue.

mvandenburgh (approved):

PR #955, opened 2024-09-20
Increase Gantry DB Backup Interval

cmelone:
Take a snapshot every 12h and keep for 5 days.

https://litestream.io/tips/#increase-snapshots-frequency-to-improve-restore-performance


cmelone:
@mvandenburgh

mvandenburgh (approved):

PR #957, opened 2024-10-03
Allow service tagged runners to define `CI_OIDC_REQUIRED`

kwryankrattiger:
This allows not erroring on OIDC setup for runners with service tag. Noop jobs use the service runners, but do not need access to S3 buckets. Jobs on these runners should not fail trying to find services they don't need.

If the variable `CI_OIDC_REQUIRED` is set in the job running on the runner, the value will be used. This will likely only be set on the `noop` jobs with a value of `0`.

If the variable `CI_OIDC_REQUIRED` is not set, the exit code will be `1`

ref. https://github.com/spack/spack/pull/46732

### DIFF
--- a/k8s/production/runners/public/graviton/2/release.yaml
+++ b/k8s/production/runners/public/graviton/2/release.yaml
@@ -61,7 +61,7 @@ spec:
 
           if [ -z "${PY3:-}" ]; then
             echo "Unable to find python3 executable"
-            exit 1
+            exit ${CI_OIDC_REQUIRED:-1}
           fi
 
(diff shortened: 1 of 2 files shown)


mvandenburgh:
@kwryankrattiger this will stop these jobs from getting reported as "Failed", right? https://gitlab.spack.io/spack/spack/-/jobs/12825706

kwryankrattiger:
It still requires a PR to Spack, but yes the goal is to stop failing noop jobs.

I still need to verify if this doesn't just abort the job with status 0 or if this exits the script and still runs the rest of CI. Honestly either is probably ok.

This change _should_ be safe to test in production as it is only changing the path of failure to being configurable to not fail. I don't think it is abusable since it affects runner types we don't really use for most of CI (they need v3).

mvandenburgh (approved):

PR #973, opened 2024-10-11
Change how `CI_OIDC_REQUIRED` works

kwryankrattiger:
Add runner and variable documentation in README

This also fixes a problem with the previous approach where if you set `CI_OIDC_REQUIRED: 0` in your job config, then the `exit 0` here prevented any of your jobs `before_script` or `script` from running at all.


scottwittenburg (approved):
I think this is a better approach than the previous one, thanks @kwryankrattiger!

PR #977, opened 2024-10-18
Enable stesachs@amazon.com to run `s3 sync` on a sub-bucket

stephenmsachs:
I want to add to the existing buildcache location and run `spack buildcache
update-index` in order to merge the content. Spack needs to be able to list the
content (or use `aws s3 sync`) to run `buildcache update-index`.

This is an extension of #731 to enable `aws s3 sync` additionally to `aws s3 cp`.

PR #978, opened 2024-10-21
 Enable stesachs@amazon.com to run `s3 sync` on a `pcluster/` sub-bucket

stephenmsachs:
`s3:CopyObjects` and `s3:ListObjects` are not registered as IAM s3 actions. Fixes #977

### DIFF
--- a/terraform/modules/spack_aws_k8s/bootstrap_s3.tf
+++ b/terraform/modules/spack_aws_k8s/bootstrap_s3.tf
@@ -36,9 +36,10 @@ resource "aws_s3_bucket_policy" "bootstrap" {
           "AWS" : "arn:aws:iam::679174810898:root"
         },
         "Action" : [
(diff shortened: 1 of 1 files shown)


mvandenburgh:
@stephenmsachs I'm getting `error MalformedPolicy: Action does not apply to any resource(s) in statement` back from AWS when I try to apply this.

stephenmsachs:
Thanks for testing. I am out of the office for a week now and will have to go look at this after that.

stephenmsachs:
I found the error causing `Action does not apply to any resource(s) in statement`. `s3:ListBucket` needs to be applied to a bucket, not a key. This setup should work now. I have tested it locally.

mvandenburgh (approved):

mvandenburgh:
Applied :heavy_check_mark:

PR #980, opened 2024-10-22
Add new regex for `concretization_error`

mvandenburgh:
Examples:
https://gitlab.spack.io/spack/spack/-/jobs/13067136 
https://gitlab.spack.io/spack/spack/-/jobs/13067292

### DIFF
--- a/analytics/analytics/core/job_failure_classifier/taxonomy.yaml
+++ b/analytics/analytics/core/job_failure_classifier/taxonomy.yaml
@@ -27,6 +27,8 @@ taxonomy:
         - "failed to concretize .+ for the following reasons"
         - "variant .+ not found in package"
         - "trying to set variant .+ in package .+, but the package has no such variant"
+        - "No such variant .+ for spec"
+        - "UnknownVariantError"
 
     job_log_missing:
         grep_for:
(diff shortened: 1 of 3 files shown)


analytics/analytics/core/job_failure_classifier/taxonomy.yaml
```diff
@@ -27,6 +27,7 @@ taxonomy:
         - "failed to concretize .+ for the following reasons"
         - "variant .+ not found in package"
         - "trying to set variant .+ in package .+, but the package has no such variant"
+        - "UnknownVariantError: No such variant .+ for spec"
```
scottwittenburg, line 30: Just wondering if reducing this to `"No such variant"` would catch a wider variety of concretization issues without risking false positives.

Or maybe adding `"UnknownVariantError"` and `"No such variant"` as two separate regexes or perhaps a single regex with a `|`?

I'm actually surprised we don't have such an expression already, I'll check.

scottwittenburg: I guess they're all right there under `concretization_error`.  Maybe the ones above correspond to older variant-related concretization errors, or maybe they're still valid, not sure.

mvandenburgh: False negatives seem to be a larger issue than false positives when it comes to concretization errors, so splitting them up is fine with me :+1:

scottwittenburg (approved):
Looks good, thanks @mvandenburgh!

PR #983, opened 2024-10-23
Add new regexes to error taxonomy

mvandenburgh:
Example jobs (these are just from today, but I've seen these errors come up multiple times the last couple days):
https://gitlab.spack.io/spack/spack/-/jobs/13088400 
https://gitlab.spack.io/spack/spack/-/jobs/13088128
https://gitlab.spack.io/spack/spack/-/jobs/13088255 
https://gitlab.spack.io/spack/spack/-/jobs/13088333
https://gitlab.spack.io/spack/spack/-/jobs/13088178 
https://gitlab.spack.io/spack/spack/-/jobs/13088078 
https://gitlab.spack.io/spack/spack/-/jobs/13088077 
https://gitlab.spack.io/spack/spack/-/jobs/13088006 
https://gitlab.spack.io/spack/spack/-/jobs/13085121 
https://gitlab.spack.io/spack/spack/-/jobs/13084926
https://gitlab.spack.io/spack/spack/-/jobs/13084862

### DIFF
--- a/analytics/analytics/core/job_failure_classifier/taxonomy.yaml
+++ b/analytics/analytics/core/job_failure_classifier/taxonomy.yaml
@@ -29,6 +29,9 @@ taxonomy:
         - "trying to set variant .+ in package .+, but the package has no such variant"
         - "No such variant .+ for spec"
         - "UnknownVariantError"
+        - "Spec version is not concrete"
+        - "UnsatisfiableSpecError"
+        - "cannot have a dependency on"
 
     job_log_missing:
         grep_for:
@@ -165,6 +168,7 @@ taxonomy:
         - 'Error: errors occurred during concretization of the environment'
         - 'cannot load package .+ from the .builtin. repository'
         - 'must have a default provider in /builds/spack/spack/etc/spack/defaults/packages.yaml'
+        - 'RuntimeError: .+ object has no attribute'
 
     invalid_pipeline_yaml:
       grep_for:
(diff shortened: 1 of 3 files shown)


analytics/analytics/core/job_failure_classifier/taxonomy.yaml
```diff
@@ -165,6 +168,7 @@ taxonomy:
         - 'Error: errors occurred during concretization of the environment'
         - 'cannot load package .+ from the .builtin. repository'
         - 'must have a default provider in /builds/spack/spack/etc/spack/defaults/packages.yaml'
+        - 'RuntimeError: .+ object has no attribute'
```
scottwittenburg, line 171: The first job you linked, [this](https://gitlab.spack.io/spack/spack/-/jobs/13088400) one, seems its close to this, but the line starts with `AttributeError:` instead of `RuntimeError:`.

scottwittenburg (changes requested):
Thanks for submitting this @mvandenburgh 👍 

As is, I think this PR would result in missing a couple of the jobs linked in the description, due to sometime `object has no attribute` comes after `RuntimeError:`, while other times it comes after `AttributeError:`.  So maybe change that regex to just `Error: .+ object has no attribute`?  Or drop the exception name altogether?  What do you think?

analytics/analytics/core/job_failure_classifier/taxonomy.yaml
mvandenburgh: I just removed `Runtime`. My thinking is its better to be more conservative, and we can adjust the regex to be more general if needed.

PR #988, opened 2024-10-31
Remove opensearch error taxonomy upload

mvandenburgh:
The error taxonomy has been moved to the analytics DB, so we no longer need to upload it to opensearch. This PR removes the entire `job_failure_classifier` package. Any functions from that package that were used elsewhere have been moved out of it.

Hopefully this also has the secondary effect of lowering the stress on the opensearch cluster and causing job log upload failures.


analytics/analytics/job_processor/dimensions.py
```diff
@@ -29,6 +32,121 @@ def __init__(self, job_id: int, name: str) -> None:
...
 
+@dataclass(frozen=True)
+class RetryInfo:
+    is_retry: bool
+    is_manual_retry: bool
+    attempt_number: int
+    final_attempt: bool
+
+
+def _job_retry_data(
```
jjnesbitt, line 43: I think this function is a bit misplaced here, since it's used in multiple files, and it feels off to lump it into "dimensions". Maybe it could go in `utils.py`, or even a new file.

jjnesbitt (changes requested):
Looks good to me, just one suggestion.

jjnesbitt (approved):

PR #989, opened 2024-10-31
Add retries to prometheus request

mvandenburgh:
Prometheus still occasionally times out _inside_ the cluster. I'm not sure why yet, but I think it's good to add retries here regardless.

### DIFF
--- a/analytics/analytics/job_processor/__init__.py
+++ b/analytics/analytics/job_processor/__init__.py
@@ -7,7 +7,7 @@ import gitlab.exceptions
 from celery import shared_task
 from django.db import transaction
 from gitlab.v4.objects import ProjectJob
-from requests.exceptions import ReadTimeout
+from requests.exceptions import ConnectionError, ReadTimeout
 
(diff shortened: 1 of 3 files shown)


analytics/analytics/job_processor/__init__.py
```diff
@@ -121,7 +121,10 @@ def create_job_fact(
 
 @shared_task(
     name="process_job",
-    autoretry_for=(ReadTimeout,),
+    autoretry_for=(
+        ConnectionError,
```
jjnesbitt, line 125: If we're already retrying the prometheus requests, what does adding an autoretry for `ConnectionError` here do?

mvandenburgh: You're right, they both do the same thing. I thought the Prometheus DNS issue was a separate exception, but it's not; I think using the celery autoretry is simpler than building in retry logic with `requests`, so I backed the latter out.

jjnesbitt (approved):

PR #990, opened 2024-10-31
Handle jobs built with compilers as nodes

jjnesbitt:
[This](https://github.com/spack/spack/pull/45189) PR changes how compilers are represented in the specs json file, in a way that breaks the current build timing processor. For the time being, the solution is to not store these specs.


.github/workflows/custom_docker_builds.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           - docker-image: ./images/cache-indexer
             image-tags: ghcr.io/spack/cache-indexer:0.0.3
           - docker-image: ./analytics
-            image-tags: ghcr.io/spack/django:0.3.22
+            image-tags: ghcr.io/spack/django:0.3.23
```
mvandenburgh, line 45: This needs to be rebased due to #989

mvandenburgh (approved):
LGTM, just need to update the image tags due to #989 merging

jjnesbitt:
On second thought, there's some other sentry errors that will still occur due to that PR/possible future implementation, so I'll address those changes before merging.

mvandenburgh (approved):

PR #991, opened 2024-10-31
Limit load balancer targets to Linux nodes

mvandenburgh:
`kube-proxy` seems to have some issues on Windows nodes. Any requests proxied to a Windows node by the load balancer will fail, resulting in the constant network timeout errors we've been seeing.


zackgalbreath (dismissed):

k8s/production/ingress-nginx/release.yaml
```diff
@@ -70,6 +70,10 @@ spec:
           # NOTE(opadron): Set the idle timeout for the ingress LB
           "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "3600"
 
+          # Force the load balancer to only route traffic to Linux nodes.
+          # kube-proxy doesn't seem to work right on Windows nodes.
+          service.beta.kubernetes.io/aws-load-balancer-target-node-labels: kubernetes.io/os=linux
```
jjnesbitt, line 75: Should this be in quotes? The above annotations are.

mvandenburgh: It doesn't matter for the purposes of YAML, but I updated it to be consistent with the other ones.

jjnesbitt (approved):

PR #992, opened 2024-11-01
Autoretry process_job for any `RequestException`

mvandenburgh:
@jjnesbitt instead of trying to enumerate all the possible ways `requests` can fail transiently (such as the new `ChunkedEncodingError` [here](https://kitware-data.sentry.io/issues/6034682867/?alert_rule_id=15115735&alert_timestamp=1730415288676&alert_type=email&environment=production&notification_uuid=d144ae5f-0d71-4496-a4a0-b2787ebb8132&project=4507069616685056&referrer=alert_email)), what do you think about just retrying all `RequestException` exceptions? That's the parent class that all these exceptions inherit from.

The only downside I see is that we may end up retrying legitimate request failures, but in that case it will just retry it 3 times and propagate the exception to Sentry. That seems more reasonable to me, any thoughts?

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -42,7 +42,7 @@ jobs:
           - docker-image: ./images/cache-indexer
             image-tags: ghcr.io/spack/cache-indexer:0.0.3
           - docker-image: ./analytics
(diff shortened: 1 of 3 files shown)


jjnesbitt (approved):
I think this is fine

PR #993, opened 2024-11-06
sync script: block on "all-prechecks"

zackgalbreath:
Update our sync script configuration to not push Spack PR branches until the "all-prechecks" check has succeeded.

### DIFF
--- a/k8s/production/custom/gh-gl-sync/cron-jobs.yaml
+++ b/k8s/production/custom/gh-gl-sync/cron-jobs.yaml
@@ -47,6 +47,6 @@ spec:
               - "--main-branch"
               - "develop"
               - "--prereq-check"
-              - "prechecks / style"
+              - "all-prechecks"
           nodeSelector:
             spack.io/node-pool: base


scottwittenburg:
The change looks right to me, based on the names of checks I see.  Just need to bump the image version number I think.  Or is there more you plan to add?

zackgalbreath:
> The change looks right to me, based on the names of checks I see. Just need to bump the image version number I think. Or is there more you plan to add?

We're not changing the source code of the sync script, just how it gets called from the cron job. So I don't think we need a version bump this time?

mvandenburgh:
> > The change looks right to me, based on the names of checks I see. Just need to bump the image version number I think. Or is there more you plan to add?
> 
> We're not changing the source code of the sync script, just how it gets called from the cron job. So I don't think we need a version bump this time?

That's correct, no version bump should be needed here

mvandenburgh (approved):

kwryankrattiger (approved):
LGTM, status name matches.

PR #994, opened 2024-11-08
Reapply Gantry DB Backup

cmelone:
I think the cluster upgrade may have wiped out the change made in #955, so I'm reapplying it here.

@mvandenburgh


mvandenburgh (approved):
Ah, you're right. Looks like it got lost during the refactoring of some of the Terraform code.

cmelone:
thanks!

PR #1011, opened 2024-12-13
Re-enable Windows Karpenter nodes

mvandenburgh:
This re-enables Karpenter support for Windows. This will get us back to autoscaling based on load, instead of just running a couple static instances 24/7.

The issue turned out to be an IAM policy that needed to be attached to the Karpenter controller node, in order for Windows nodes to have the necessary permissions to create EC2 instance profiles. Previously, Karpenter created a static instance profile for all nodes, but starting in v1, they changed the behavior. Since we're using Karpenter for both Linux and Windows nodes, there's some complications to setting up the Terraform modules; I've expanded on the details in a code comment.

### DIFF
--- a/terraform/modules/spack_aws_k8s/eks.tf
+++ b/terraform/modules/spack_aws_k8s/eks.tf
@@ -105,43 +105,6 @@ module "eks" {
         }
       }
     }
-    windows_mng = {
-      platform = "windows"
-      ami_type = "WINDOWS_FULL_2022_x86_64"
-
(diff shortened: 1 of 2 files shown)


mvandenburgh:
Job scheduled on a Karpenter windows node in staging - https://gitlab.staging.spack.io/spack/spack/-/jobs/108126

jjnesbitt (approved):
Awesome! Guess we'll want to keep an eye on windows jobs for a little while to make sure things are going according to plan.

PR #1014, opened 2024-12-18
Drop graviton2 (neoverse_n1). Add Neoverse_v1 tags to graviton3

kwryankrattiger:


zackgalbreath:
What do you think about getting rid of them (and their Karpenter ~provisioner~ node pool) altogether?

kwryankrattiger:
Until we get confirmation on the pcluster stacks I didn't want to pull the rug out entirely. This will at least stop using them for all of the stacks not explicitly requesting `graviton2`.

If we don't hear by the EOD today, I think I am going to PR not requiring pcluster to pass CI so that at very least we don't block on sporadic failed jobs there. But it will at very least keep running on populating the cache for them to the extent that it can.

kwryankrattiger:
Intentionally opting not to rename `graviton` -> `neoverse` to avoid missing something. Simply dropping the graviton2 references and adding the `neoverse_v1` tag for now.

This requires moving all of the current stacks reference `graviton2` to move to `graviton3` runners. A follow up should change `graviton3` -> `neoverse_v1`

k8s/production/runners/protected/graviton/2/release.yaml
```diff
@@ -185,7 +185,7 @@ spec:
       protected: true
       runUntagged: false
 
-      tags: "arm,aarch64,graviton,graviton2,small,medium,large,huge,protected,aws,spack"
+      tags: "graviton2,small,medium,large,huge,protected,aws,spack"
```
zackgalbreath, line 188: shouldn't we remove this runner (and the corresponding public one) since we're removing its node pool?

kwryankrattiger: bleh, it got undeleted in a rebase, repushing

zackgalbreath (approved):
Do you have any thoughts on when this should be safe to merge? Should we wait a bit for open spack PRs to get rebased on https://github.com/spack/spack/pull/48186 ?

kwryankrattiger:
Once this pipeline finishes I think we are good to go https://gitlab.spack.io/spack/spack/-/commit/9719220e8a70b68b75f71c3aa855a661fb7ee318

PR #1015, opened 2024-12-18
webhook-handler: ignore install times for failed jobs

mvandenburgh:
If a build job fails, it won't have an install_times.json.

@jjnesbitt this accounts for some, but not all, of the sentry errors complaining about missing install_times.json.

### DIFF
--- a/analytics/analytics/job_processor/artifacts.py
+++ b/analytics/analytics/job_processor/artifacts.py
@@ -8,6 +8,12 @@ from gitlab.exceptions import GitlabGetError
 from gitlab.v4.objects import ProjectJob
 
 
+class JobArtifactDownloadFailed(Exception):
+    def __init__(self, job: ProjectJob) -> None:
+        message = f"Job {job.id} artifact download failed"
+        super().__init__(message)
+
+
 class JobArtifactFileNotFound(Exception):
     def __init__(self, job: ProjectJob, filename: str):
         message = f"File {filename} not found in job artifacts of job {job.id}"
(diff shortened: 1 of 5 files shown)


analytics/analytics/job_processor/artifacts.py
```diff
@@ -37,7 +43,7 @@ def get_job_artifacts_file(job: ProjectJob, filename: str):
             with open(artifacts_file, "wb") as f:
                 job.artifacts(streamed=True, action=f.write)
         except GitlabGetError:
-            raise JobArtifactFileNotFound(job, filename)
+            raise JobHasNoArtifacts(job)
```
jjnesbitt, line 46: How can we confidently raise this error here, as opposed to the first one? We don't really know that the job has no artifacts based on this exception, do we?

mvandenburgh: Good point. I renamed it to `JobArtifactDownloadFailed`

jjnesbitt: Also, the existing `JobArtifactFileNotFound` exception is caught in https://github.com/spack/spack-infrastructure/blob/d5df819cdf1a7beef5e0e0e7cd0fd37276404681/analytics/analytics/job_processor/metadata.py#L183

which will re-raise the error if the job status isn't `failed`. So we'd also need to update that.

mvandenburgh: Updated

jjnesbitt (approved):

PR #1018, opened 2024-12-21
webhook-handler: Don't crash when `install_times.json` is missing

mvandenburgh:
Since the `install_times.json` file is generated using `spack` itself, it is possible that the file is missing in a PR pipeline. This PR accounts for this case and gracefully exits when the file is missing.

We should set up alternative alerting mechanisms (like a CI test in`spack/spack` for the install_times generator, or a metabase alert that lets us know when a large amount of timing data is missing from the analytics DB) to handle this, instead of blowing up Sentry with errors when a spack PR breaks it.

### DIFF
--- a/analytics/analytics/job_processor/build_timings.py
+++ b/analytics/analytics/job_processor/build_timings.py
@@ -9,7 +9,7 @@ from analytics.core.models.dimensions import (
     TimerPhaseDimension,
 )
 from analytics.core.models.facts import JobFact, TimerFact, TimerPhaseFact
-from analytics.job_processor.artifacts import get_job_artifacts_file
+from analytics.job_processor.artifacts import JobArtifactFileNotFound, get_job_artifacts_file
 
 
 def get_timings_json(job: ProjectJob) -> list[dict]:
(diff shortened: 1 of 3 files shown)


analytics/analytics/job_processor/build_timings.py
```diff
@@ -103,7 +103,14 @@ def get_package_mapping(timings: list[dict]) -> dict[str, PackageDimension]:
...
 
 def create_build_timing_facts(job_fact: JobFact, gljob: ProjectJob):
-    timings = [t for t in get_timings_json(gljob) if t.get("name") and t.get("hash")]
+    # Sometimes the timing file isn't present. This could be due to a number of reasons;
+    # since `install_times.json` is created using `spack` itself, there's no guarantee
+    # that the file will be present.
+    try:
+        data = get_timings_json(gljob)
+        timings = [t for t in data if t.get("name") and t.get("hash")]
+    except JobArtifactFileNotFound:
```
jjnesbitt, line 112: Just to be clear, we _don't_ want to catch `JobArtifactDownloadFailed` here, right? As in, we want to consider missing all artifacts a failure, but missing only the timings file isn't a failure?

mvandenburgh: Right. The only time a job should be missing artifacts is when it failed. If a successful job doesn't have any artifacts, I think a sentry alert would be appropriate

jjnesbitt (approved):

PR #1021, opened 2025-01-03
webhook handler: Add computed DB column for infrastructure errors

mvandenburgh:
This will allow us to just rely on this one column in our metabase dashboards, instead of adding a filtering clause for the different error taxonomy types that imply an infrastructure error and having to keep each dashboard up to date if we decide to add more.

Note, `GeneratedField` was introduced to the ORM in Django 5, so that's why I also did that upgrade here.


mvandenburgh:
I noticed we're including dev dependencies in `requirements.txt`. I think we should remove those so the production server doesn't install them, and just install directly from `setup.py` when dev dependencies are needed. To keep this PR scoped, I just left it as is.

.github/workflows/custom_docker_builds.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           - docker-image: ./images/cache-indexer
             image-tags: ghcr.io/spack/cache-indexer:0.0.3
           - docker-image: ./analytics
-            image-tags: ghcr.io/spack/django:0.3.30
+            image-tags: ghcr.io/spack/django:0.3.31
```
jjnesbitt, line 45: This is pretty minor but, since we're upgrading Django to 5.0, should we make this a minor version change?

jjnesbitt (commented):
One suggestion but otherwise LGTM

.github/workflows/custom_docker_builds.yml
mvandenburgh: https://github.com/spack/spack-infrastructure/pull/1021/commits/8b66a8832f7ea74e2f2e407d4c4fd0b3fd827840

mvandenburgh: > This is pretty minor

:drum:

jjnesbitt: Completely unintentional believe it or not :laughing:

jjnesbitt (approved):

PR #1025, opened 2025-01-06
Ignore all extra cache=False install times

jjnesbitt:
This should fix all the `KeyError` errors we've been seeing in sentry.

This was being caused by install time entries that didn't exist in the spec json. This seems to mainly occur with bootstrapped packages.

### DIFF
--- a/analytics/analytics/job_processor/build_timings.py
+++ b/analytics/analytics/job_processor/build_timings.py
@@ -9,7 +9,10 @@ from analytics.core.models.dimensions import (
     TimerPhaseDimension,
 )
 from analytics.core.models.facts import JobFact, TimerFact, TimerPhaseFact
-from analytics.job_processor.artifacts import JobArtifactFileNotFound, get_job_artifacts_file
+from analytics.job_processor.artifacts import (
+    JobArtifactFileNotFound,
+    get_job_artifacts_file,
+)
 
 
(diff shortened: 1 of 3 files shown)


jjnesbitt:
I've tested the `process_job` function on a job that was known to trigger this issue, and can confirm it's resolved with this change.

analytics/analytics/job_processor/build_timings.py
```diff
@@ -107,11 +110,16 @@ def create_build_timing_facts(job_fact: JobFact, gljob: ProjectJob):
     # since `install_times.json` is created using `spack` itself, there's no guarantee
     # that the file will be present.
     try:
-        data = get_timings_json(gljob)
-        timings = [t for t in data if t.get("name") and t.get("hash")]
+        timing_data = get_timings_json(gljob)
     except JobArtifactFileNotFound:
         return
 
+    # For boostrapped packages, install times with cache=False can be present, which don't have a
```
mvandenburgh, line 117:
```suggestion
    # For bootstrapped packages, install times with cache=False can be present, which don't have a
```

mvandenburgh (approved):

mvandenburgh (approved):
@jjnesbitt I'll just go ahead and merge this so we can re-enable Sentry

PR #1026, opened 2025-01-07
Misc. improvements to analytics Django app

mvandenburgh:
Follow up from [this comment](https://github.com/spack/spack-infrastructure/pull/1021#issuecomment-2569483297), plus fixes some other minor issues I noticed with the analytics/"webhook-handler" code.


.github/workflows/analytics_ci.yml
```diff
@@ -1,11 +1,31 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+
+      - name: Install requirements
+        run: pip install .[dev]
+
+      - name: Format code
+        run: tox -e format
```
jjnesbitt, line 27: Wouldn't we want to run `tox -e lint` instead of `format`? Since `format` won't actually contribute the formatted code back to the repo.

mvandenburgh: `lint` runs `flake8` on the code, while `format` runs `black` and `isort`. The codebase isn't compliant with `flake8` rules yet, so I figured just running `black` and `isort` (which is a subset of the formatting rules enforced by `flake8`) is at least a good start. 

`format` will still fail with a non-zero exit code if the code needs to be reformatted.

jjnesbitt: Should we switch `lint` to simply run `black` and `isort`? If the codebase is already compliant with those two anyway, it would largely result in no change.

mvandenburgh: I updated `lint` to run `black` and `isort`

jjnesbitt (approved):

PR #1027, opened 2025-01-07
webhook-handler: increase gitlab API timeout

mvandenburgh:
GitLab itself won't timeout until after 60 seconds https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/bulk_imports/clients/graphql.rb#L7, and we've seen cases where fetching a job trace takes more than 15 seconds. So, I've bumped the timeout for the GitLab Python client to 30 seconds to account for this; we can always increase it further if needed, but I think this should be a good start while avoiding potentially introducing head-of-line blocking to the celery queue.


mvandenburgh:
(For the record, @jjnesbitt approved this outside of github 🙂)

PR #1042, opened 2025-01-22
Store gitlab failure reason/exit code in analytics DB

mvandenburgh:
This PR updates the webhook-handler to store the `build_failure_reason` reported by gitlab (for example, `script_failure`) in the job data dimension table. It also attempts to parse out the exit code from the job log and store it if found.

A couple things to note-

1) ~From what I can tell, gitlab does not store the job exit code anywhere. So, I opted to parse it out of the job log with a regex. If there's a better way/it is in fact stored somewhere in the gitlab DB, I'd be happy to change this.~ Updated PR to use DB instead

2) I suspect that only `script-failure` jobs have an exit code, so we should be able to add a constraint that enforces exit code to be non-null for those jobs. I don't have any evidence for this yet, so I figured we can leave it as null-able for now and revisit it later when we have some data.

3) I've added a temporary management command to backfill the new fields; I didn't want to make it a migration, because long-running migrations don't work great with our current process for running them.


zackgalbreath:
> 1. From what I can tell, gitlab does not store the job exit code anywhere

I believe this is stored in the GitLab database as `ci_builds_metadata.exit_code`

mvandenburgh:
> > 1. From what I can tell, gitlab does not store the job exit code anywhere
> 
> I believe this is stored in the GitLab database as `ci_builds_metadata.exit_code`

I made a query in metabase that grabs all the `build_failure_reasons` from the webhook logs table, and joins it to the `ci_builds_metadata` table to check this https://metabase.spack.io/question/113-gitlab-db-script-failure-jobs-w-exit-codes-joined (note - that query takes ~20 seconds to run), and it looks like that column is `NULL` most of the time. And if you remove the `AND ("public"."ci_builds_metadata"."exit_code" IS NOT NULL)` line from that query, you'll see all the `script_failure` jobs don't have exit_codes. :man_shrugging:

mvandenburgh:
> > > 1. From what I can tell, gitlab does not store the job exit code anywhere
> > 
> > 
> > I believe this is stored in the GitLab database as `ci_builds_metadata.exit_code`
> 
> I made a query in metabase that grabs all the `build_failure_reasons` from the webhook logs table, and joins it to the `ci_builds_metadata` table to check this [metabase.spack.io/question/113-gitlab-db-script-failure-jobs-w-exit-codes-joined](https://metabase.spack.io/question/113-gitlab-db-script-failure-jobs-w-exit-codes-joined) (note - that query takes ~20 seconds to run), and it looks like that column is `NULL` most of the time. And if you remove the `AND ("public"."ci_builds_metadata"."exit_code" IS NOT NULL)` line from that query, you'll see all the `script_failure` jobs don't have exit_codes. 🤷‍♂️

This might be incorrect. Marking this as draft until I can confirm

mvandenburgh:
There must be something going on with the regex in that metabase query, because it does look like at least `script_failure` jobs always have a non-null exit code. I updated the PR to use that instead.

analytics/analytics/core/management/commands/migrate_exit_code_and_failure_reasons.py
```diff
@@ -0,0 +1,123 @@
...
+            JobDataDimension.objects.filter(job_id__in=job_id_to_exit_code_mapping.keys())
+        )
+        for job in jobs_to_update:
+            job.job_exit_code = job_id_to_exit_code_mapping[job.job_id]
+
+        JobDataDimension.objects.bulk_update(jobs_to_update, ["job_exit_code"])
+
+
+# From https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/concerns/enums/ci/commit_status.rb#L7
+FAILURE_REASONS = {
```
jjnesbitt, line 40: I wonder if you could use this in the opposite direction to eliminate the need for `get_job_exit_code`? If the mapping is consistent, it could remove the need to query the gitlab DB.

mvandenburgh: This enum mapping is unrelated to the exit code, it's just used to map the `ci_builds.failure_reason` column to a string.

mvandenburgh:
For posterity - @jjnesbitt and I discussed this offline and decided to move the backfilling of exit codes/failure reasons to a Django migration instead of a management command.

analytics/analytics/core/migrations/0003_jobdatadimension_gitlab_failure_reason_and_more.py
```diff
@@ -0,0 +1,145 @@
...
+    JobDataDimension = apps.get_model("core", "JobDataDimension")
+
+    # Only failed jobs have an exit code in the database
+    failed_jobs = JobDataDimension.objects.filter(status="failed")
+
+    # Batch the job ids to avoid hitting the max query length/memory limit
+    BATCH_SIZE = 5_000
+    for job_ids in batched(
+        failed_jobs.values_list("job_id", flat=True).iterator(chunk_size=BATCH_SIZE), BATCH_SIZE
+    ):
```
jjnesbitt, line 20: Is there a reason to call `batched` alongside `.iterator`? I don't see the benefit.

jjnesbitt (commented):
Just had one remaining question, but otherwise LGTM. We'll just need to make sure to disable flux before we merge this, so we can manually run `migrate`.

analytics/analytics/core/migrations/0003_jobdatadimension_gitlab_failure_reason_and_more.py
mvandenburgh: It's so the gitlab DB queries can be done in batches, instead of querying for one job at a time.

jjnesbitt: Ahh, gotcha. Makes sense.

jjnesbitt (approved):

analytics/analytics/core/migrations/0003_jobdatadimension_gitlab_failure_reason_and_more.py
mvandenburgh: To clarify in case the .iterator() usage is misleading - the `chunk_size` argument of `.iterator()` doesn't do any batching at the Python-level. It controls the size of the batch of rows the cursor grabs at the DB driver level, but from Python's perspective, it's always just an `Iterator` to be exhausted.

PR #1047, opened 2025-01-30
Don't explicitly log out in docker build job

mvandenburgh:
It seems there is the possibility of a race condition here - https://news.ycombinator.com/item?id=28607735

Thanks @johnwparent !


mvandenburgh:
I guess there's not really a way to test that this works without merging it. :man_facepalming:

PR #1049, opened 2025-01-30
Add script for retrying all missed job webhooks

mvandenburgh:
This adds a Django management script that
1) Queries the `web_hook_logs` table of the GitLab DB for all webhook events in the past _n_ seconds (_n_ is an argument the script)
2) Cross checks each of those rows with the Analytics DB, to figure out if any jobs had webhooks fired but do not exist in the analytics DB (i.e. jobs that were "missed" for any reason)
3) Uses the GitLab API to resend those webhook events.

I've ran variations of this script in the past to retry missed job webhooks after fixing bugs in the webhook handler, downtime of the webhook handler, etc. and it works.


mvandenburgh:
@jjnesbitt with the analytics webhook being down the last couple days, it seems like now would be a good time to test this out and get it merged. IIRC I ran it a while back and it appeared to work, but I don't think we've needed to retry this main job events before.

PR #1050, opened 2025-01-31
Split up job data dimension

jjnesbitt:
This PR accomplishes the following:

1. Split up `JobDataDimension` into the following dimensions:
	* `SpackJobDataDimension`
	* `GitlabJobDataDimension`
	* `JobResultDimension`
	* `JobRetryDimension`

	This data was all bundled into the `JobDataDimension`, which necessitated an entry in that dimensional table for every job. This meant that the `JobDataDimension` table was as large as the `JobFact` table, which is undesirable, and results in poor join performance. With this dimension split up, each table is much smaller, which results in more re-used rows, as well as faster joins.

2. The storage of gitlab section timers on the `JobFact` table. Since this is numeric, and will have aggregations performed on it, it's the correct place to store it.

3. Setting the `job_id`  as the primary key for the `JobFact` table. This is really how we expect these facts to exist anyway (only one job fact per job). As a result of the _old_ primary key for this table, there were actually several duplicate job entries, which this PR cleans up.
4. The removal of the `JobDataDimension` table. This is due to all of its constituent data being split into smaller dimension tables. However, this also results in the removal of that foreign key from the `TimerFact` and `TimerPhaseFact` tables, although they still store the `job_id`. I plan to evaluate if and what new relations should be introduced, and make those changes in the future.

I have run these migrations locally and can confirm there are no errors that arise. I also have tested some webhook payloads and it seems the job processor runs without error. The migration downtime for this should be ~6 hours in production.


analytics/analytics/core/models/dimensions.py
```diff
@@ -174,6 +206,11 @@ class Meta:
         db_comment="Whether this job has 'No need to rebuild' in its trace.",
     )  # type: ignore
 
+    # This field is duplicated here (from the gitlab job data dimension)
+    # to allow for the generated `infrastructure_error` field
+    job_type = models.CharField(
+        max_length=max(len(c) for c, _ in JobType.choices), choices=JobType.choices
+    )
```
mvandenburgh, line 213: We could consider using a database trigger to set `infrastructure_error` instead of duplicating this column. Something to think about for a follow up at least; there's already substantial DB changes happening in this PR so I don't think it's worth adding more complexity unless you feel otherwise

mvandenburgh (approved):
Haven't tested this yet, but it makes sense to me from a design perspective :+1:

analytics/analytics/core/models/dimensions.py
jjnesbitt: We could do that as a follow up, although I'll just point out that duplicate data isn't a problem at all in this type of database. Since these records aren't subject to modification like they would be in a normal web application (or if they are it's a system wide modification applied consistenly), it's not an issue.

PR #1052, opened 2025-02-06
Add option to skip deferral check for urgent PRs

zackgalbreath:
Skip the "deferred pipelines" check for PRs labeled "pipelines:urgent"


images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -202,7 +202,8 @@ def list_github_prs(self):
                         print("Skip pushing {0} because of {1}".format(pr_string, backlogged))
 
                 if not backlogged:
-                    if self.main_branch and pull.base.ref == self.main_branch:
+                    check_for_deferral = not any(label.name == 'pipelines:urgent' for label in pull.labels)
+                    if check_for_deferral and self.main_branch and pull.base.ref == self.main_branch:
```
kwryankrattiger, line 206: Is the "else" to this still going to push the bench to Gitlab?

zackgalbreath: Exactly. It's the same "my base isn't develop" case we were looking at earlier this week.

kwryankrattiger: Right. Some branches in PRs are named "develop" or may have other problematic names that could run CI we didn't expect. We also don't clean up non-PR branches pushed to Gitlab.

zackgalbreath: The context of this change is only inspecting PR branches from GitHub. It's also worth pointing out that the names of the branches that we push to GitLab follow a particular format. For example, if the topic branch for PR 12345 is named `develop`, then we would push `pr12345_develop` to GitLab.

kwryankrattiger: Oh okay. So the context now, this is pushing the PR branch even if develop is still running.

In that case this makes sense to me!

scottwittenburg (approved):
I think this is safe to do, but I hope everyone doesn't start using it, or we may be back to rebuilding the same hashes in every PR because the current develop pipeline hasn't quite finished pushing them yet.

PR #1058, opened 2025-02-12
Support content addressable buildcache

scottwittenburg:
Until spack has buildcache views, pipelines must be run pointing at independent mirrors, and `protected-publish` is responsible for making stack-specific binaries available at the mirror root.

Add support for "publishing" both old style as well as content addressable (`v2` vs `v3`) buildcaches from stacks to the root. By default the script assumes `v2` layout, but it can be changed by adding `--version <version>` on the command line.  The only accepted options are 2 and 3.

Add an endpoint in the container that can migrate a binary mirror from the previous version to the content-addressed layout.  Migration consists of, for each spec in the mirror:

- validate the current signature on the metadata
- add timestamp, size, and compression type attributes, and update layout version attribute from 2 to 3
- re-sign the updated spec with the original signing key
- upload the metadata and copy the compressed tarball to content-addressed locations


images/protected-publish/TESTING.txt
```diff
@@ -0,0 +1,92 @@
+There is some curated testing data located in:
```
scottwittenburg, line 1: This file doesn't have to be part of it, it's just here in case it's helpful during evaluation of the PR.

scottwittenburg: Or it could be fixed up however is needed and incorporated into the README.

images/protected-publish/oneshot_pod.yaml
```diff
@@ -0,0 +1,52 @@
...
+    metadata:
+      labels:
+        app: key-access
+    spec:
+      serviceAccountName: naccess
+      containers:
+      - name: access
+        image: ghcr.io/scottwittenburg/protected-publish:0.0.4
+        command: ["/bin/bash", "-c"]
+        args: ["cat /srcs/test_stacks.txt | /srcs/migrate.sh"]
```
scottwittenburg, line 23: I'd rather specify the the list of stacks to publish when I launch the pod (i.e. when I `kubectl apply -f` these yamls files), but I haven't yet found a way to do that.

scottwittenburg: @mvandenburgh in case you have an idea how to do this.

kwryankrattiger: Talked about making the args a list of mirror names.

images/protected-publish/README.md
```diff
@@ -189,10 +231,13 @@ To examine a local or remote (S3 only) index for any missing specs:
...
     -ti protected-publish:latest \
-    validate_index.py --url s3://spack-binaries/develop-2024-01-07
+    -m pkg.validate_index \
+    --url s3://spack-binaries/develop-2024-01-07 --version 2
 ```
 
-Optionally you can specify a local file instead with `--file` and an absolute path.
+The version is used to select whether v2 or v3 index is sought.
+
+Optionally, instead of specifying `--url` and `--version`, you can specify a local file with `--file` followed by an absolute path.
```
zackgalbreath, line 240: is this `--file` argument what `develop_stacks.txt` and `test_stacks.txt` is meant to be used for?

images/protected-publish/pkg/migrate.py
```diff
@@ -0,0 +1,412 @@
...
+    )
+
+    print(f"Uploading {signed_specfile_path} to s3://{bucket}/{new_layout_meta_prefix}")
+
+    try:
+        s3_upload_file(signed_specfile_path, bucket, new_layout_meta_prefix)
+    except Exception as e:
+        error_msg = getattr(e, "message", e)
+        error_msg = (
+            f"Migrattion failed: unable to upload re-signed metadata for "
```
zackgalbreath, line 223:
```suggestion
            f"Migration failed: unable to upload re-signed metadata for "
```

images/protected-publish/pkg/migrate.py
```diff
@@ -0,0 +1,412 @@
...
+        return MigrationResult(False, error_msg)
+
+    if os.path.exists(verified_specfile_path) and force:
+        os.remove(verified_specfile_path)
+
+    if not os.path.exists(verified_specfile_path):
+        # Verify the signature of the locally downloaded metadata file, it seems
+        # when you let gpg deduce what to do from the arguments, it not only
+        # verifies, but also strips the signature material and writes the file
+        # back to disck with the .sig extension removed.
```
zackgalbreath, line 116:
```suggestion
        # back to disk with the .sig extension removed.
```

zackgalbreath (changes requested):
a couple questions and a couple typos, otherwise LGTM!

images/protected-publish/README.md
scottwittenburg: No, looks like this is is the `validate` functionality, so `--file` is supposed to be a buildcache index.  I'll update the text.

Those files are meant to be pasted in to the `migrate_job.yaml` before running it to migrate a particular swath of mirrors.  Maybe they don't belong in the repo?

images/protected-publish/pkg/migrate.py
```diff
@@ -0,0 +1,412 @@
...
+    s3_copy_file,
+    s3_download_file,
+    s3_upload_file,
+    spec_catalogs_from_listing_v2,
+)
+
+sentry_sdk.init(traces_sample_rate=1.0)
+
+CONTENT_ADDRESSABLE_TARBALLS_REPO = "https://github.com/scottwittenburg/spack.git"
+CONTENT_ADDRESSABLE_TARBALLS_REF = "content-addressable-tarballs-2"
```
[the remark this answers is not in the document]
scottwittenburg: I just created a "todo after merging" section in the description of the spack PR.

zackgalbreath (approved):

PR #1060, opened 2025-02-19
Update GitLab RDS instance to `gp3` storage

mvandenburgh:
We are currently on gp2, and encountering IOPS issues. Updating to gp3 will increase the IOPS and also save us some money.

Screenshot from 2025-02-19 12-27-57

This will require some downtime; if no one has any objections to this, I'll apply it to staging and then coordinate the deploy to production.

### DIFF
--- a/terraform/modules/spack_aws_k8s/gitlab_db.tf
+++ b/terraform/modules/spack_aws_k8s/gitlab_db.tf
@@ -45,6 +45,7 @@ module "gitlab_db" {
   max_allocated_storage = 1000
   storage_type          = "gp3"
   iops                  = 12000 # 12,000 is the minimum IOPs for gp3 storage. We can increase this as needed.
+  storage_throughput    = 500   # 500 is the minimum throughput for gp3 storage. We can increase this as needed.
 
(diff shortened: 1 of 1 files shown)


mvandenburgh:
Some evidence that this may be related to the reliability issues we've seen with gitlab.spack.io lately. The current RDS instance with `gp2` storage supports a max of 3,000 IOPS, assuming I am reading https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html properly.

Screenshot from 2025-02-19 16-54-23

jjnesbitt (approved):
This LGTM, although I'm curious how this migration will play out. I'm assuming terraform is able to replace the storage type on its own, but would that take all of our data with it? Or is that migration taken care of?

mvandenburgh:
> This LGTM, although I'm curious how this migration will play out. I'm assuming terraform is able to replace the storage type on its own, but would that take all of our data with it? Or is that migration taken care of?

I believe replacing the storage type doesn't destroy any data, it just incurs some downtime while the switchover happens. I'm running it on staging now to confirm.

PR #1061, opened 2025-02-20
Update analytics, CDash RDS instances to gp3

mvandenburgh:
Similar to #1060. We don't have any pressing issues with the performance of `gp2` on these instances, but `gp3` is cheaper in addition to more performant, so there's only upsides to doing this.


mvandenburgh:
This is applied to both production and staging :heavy_check_mark:

PR #1069, opened 2025-03-06
migration: update develop stack list

scottwittenburg:
This is just to record the stacks we're migrating, so it should not require any image version bumps.


zackgalbreath (approved):
This diff matches the list of stacks that we decided to migrate.

scottwittenburg:
Whoops, I forget we have [this](https://github.com/spack/spack-infrastructure/actions/runs/13708599934/job/38339826290?pr=1069#step:4:5) now, hopefully it won't prevent me from merging.

PR #1072, opened 2025-03-11
Update error taxonomy with new regexes

mvandenburgh:
Hopefully this will help cut down the number of failed jobs that are labelled as `other` recently.


mvandenburgh:
This PR fell off my radar for a bit :sweat_smile:

These types of errors haven't appeared lately, so I think we can just close this and address them later if they re-emerge

PR #1073, opened 2025-03-11
Update analytics python dependencies

mvandenburgh:
This will resolve some dependabot alerts and allows PRs like #1070 to be closed.

To generate this, I just created a new venv and ran a `pip install --no-cache-dir --upgrade . && pip freeze`.


mvandenburgh:
I think it's unwise to merge this at this point, due to Sentry being disabled because of https://github.com/spack/spack/pull/49547/ causing all webhook tasks to fail currently. I think we can just close this and regenerate the `requirements.txt` in a new PR when it's safe to do so

PR #1074, opened 2025-03-11
Remove obselete github actions workflow

mvandenburgh:
@kwryankrattiger this github action hasn't been ran in a long time, so I assume we are no longer using the AMIs it builds. Is that correct?


kwryankrattiger (approved):

kwryankrattiger:
Yeah, I think these are dead. @stephenmsachs is using a custom image instead iirc

PR #1077, opened 2025-03-20
Switch GitLab DB to t4g instance type

mvandenburgh:
This switches our GitLab RDS instances from `db.t3.*` to `db.t4.*`. The `t4g` instance family has equivalent specs to the `t3` instance family, but is cheaper since they are graviton-based.


mvandenburgh:
I applied this to the production DBs.

PR #1079, opened 2025-03-25
gh-gl-sync: Add handling to post status to for merge queue

kwryankrattiger:
Enable posting "skipped" statuses to merge queue branches CI status for gitlab pipelines.

Placeholder for forwarding statuses from gitlab for these pipelines.


images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -318,6 +333,19 @@ def list_github_protected_branches(self):
             print("    {0}".format(protected_branch))
         return protected_branches
 
+    def list_github_mq_branches(self):
+        """ Return a list of branch names associated with a merge queue"""
+        branches = self.py_gh_repo.get_branches()
```
zackgalbreath, line 338: To preserve our precious GitHub API rate limit, we should cache the results of `get_branches()` so we don't have to call it twice (here and and in `list_github_protected_branches()`

kwryankrattiger: Good call :)

PR #252, opened 2022-05-13
[WIP] Signing infrastructure

kotfic:
Add infrastructure to manage signing packages

PR #1090, opened 2025-04-09
notary: Use separate input and output directories

scottwittenburg:
After content-addressable tarballs is merged, clearsigning will be performed on spec manifest files, without changing the name of the file. This change keeps the original file name when re-signing, and to prevent issues with gpg overwriting files, it puts re-signed files into a different directory from the original files.

This can be merged before the content-addressable tarballs spack [PR](https://github.com/spack/spack/pull/48713).

### DIFF
--- a/.github/workflows/custom_docker_builds.yml
+++ b/.github/workflows/custom_docker_builds.yml
@@ -37,7 +37,7 @@ jobs:
           - docker-image: ./images/gitlab-skipped-pipelines
             image-tags: ghcr.io/spack/gitlab-skipped-pipelines:0.0.2
           - docker-image: ./images/notary
-            image-tags: ghcr.io/spack/notary:0.0.1
+            image-tags: ghcr.io/spack/notary:0.0.2
           - docker-image: ./images/python-aws-bash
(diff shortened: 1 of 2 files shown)


zackgalbreath:
> The notary image appears to be the only one where we don't pinpoint a specific numeric version of the image to build and use, does anyone know if that was an oversight or was it deliberate?

FWIW, there's no discussion on #252 about why we did it this way at the time. At the very least, we are setting `imagePullPolicy: "always"` [here](https://github.com/spack/spack-infrastructure/blob/main/k8s/production/runners/signing/release.yaml#L184).

Off the top of my head, I can't think of any reason why pinning the notary runner to a specific tag would be a bad idea.

scottwittenburg:
> Off the top of my head, I can't think of any reason why pinning the notary runner to a specific tag would be a bad idea.

Ok, I'll start that by specifying `0.0.1` in this PR.

zackgalbreath (approved):

zackgalbreath:
Noting for posterity that the lack of image bump in `k8s/production/runners/signing/release.yaml` is intentional.

PR #1092, opened 2025-04-10
notary: Fix another issue with changed signing procedure

scottwittenburg:
Before files had .sig extension, and gpg could guess the operation was verify. The operation had the result of writing the file (minus the clear signature) to disk without the .sig name.

By removing the .sig extension and specifying --verify, the problem was that the signed file was still sitting there (with the signature wrapped around it).  But the signing script then deleted that file in the loop, so there were no files to sign in the next loop.

Now verification puts the extracted files (no more signature) into an intermediate directory, deleting the files in the input directory as it goes.  The next loop iterates files in the intermediate dir, writing the resigned files into the output directory, and deleting the itermediate files in the loop.


scottwittenburg:
@zackgalbreath I've left the tagged/pushed image at `0.0.2` for this, but let me know if you're aware of any reasons I should bump it.

zackgalbreath (approved):

PR #1095, opened 2025-04-17
Search through artifacts zip for files of interest

jjnesbitt:
Fixes an issue where the webhook handler would crash if processing jobs from https://github.com/spack/spack/pull/49547.

Previously we simply hardcoded paths like `jobs_scratch_dir/reproduction/cloud-ci-pipeline.yml`. Since this occasionally changes, searching through the artifacts file list fixes the problem.


jjnesbitt:
Holding off merging until https://github.com/spack/spack/pull/49547 has fixed the issue with missing `cloud-ci-pipeline.yml` file.

PR #1097, opened 2025-04-21
Update cdash.spack.io to be deployed by Helm

zackgalbreath:
Also update our terraform config to:
* Use postgres instead of mysql
* Provision an S3 bucket for CDash to read/write files to
* Create a new ServiceAccount to allow CDash to access this S3 bucket


mvandenburgh:
@zackgalbreath I rebased this on #1136 and updated the target branch accordingly. So we should be able to merge #1136 into `main`, and we can leave this PR open until we are ready to migrate the cdash.spack.io database.

PR #1099, opened 2025-04-22
Fix protected publish for v3 layout buildcaches

scottwittenburg:
Updates the v3 version of protected publish to correctly look for manifests, generating the meta and archive urls from those.

I tested this by publishing a single test stack (build_systems) to the root under both v2 and v3 layouts.  We should be able to merge it now, and publishing v2 stacks to the root will continue to work.  Once the content-addressable tarballs PR is merged to spack `develop`, we will need a tiny PR to the cron job arguments to swtich from `--version 2` to `--version 3`.


scottwittenburg:
Not sure it makes sense to update the image version in the migrate job, since we're not planning to update that for the newest v3 layout.

zackgalbreath (approved):

PR #1103, opened 2025-04-25
Fix migration to create most recent layout

scottwittenburg:
Update migration logic to latest layout structure:

- For all specs:
  * Download v2 spec, compress it, and compute checksum
  * Push compressed spec as a blob
  * Create and push manifest for buildcache entry (including archive and spec file)
- For all public keys:
  * Download key, compute checksum and size
  * Push uncompressed key as blob
  * Create and push key manifest

Buildcache index and public key index are still handled by `spack buildcache update-index --keys <mirror-url>` after completing the steps above.  The `update-index` is currently using my branch, but can be updated to use mainline spack `develop` after the content-addressable tarballs [PR](https://github.com/spack/spack/pull/48713) is merged.


kwryankrattiger (approved):
LGTM

PR #1104, opened 2025-04-25
Refactor management of gitlab access tokens

mvandenburgh:
This PR does two things:
1) Refactors gitlab personal access tokens into a reusable Terraform module
2) Increases the time we will get warned about expiring tokens from 1 day before the expiration date to 1 month.

Currently we use the GitLab Terraform provider's [`gitlab_personal_access_token`](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/personal_access_token) resource to manage the access tokens we need to create and store in k8s secrets for use by our various cron jobs and other services to interact with the GitLab API. I set up some [Terraform `precondition` rules](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions) to fail TF plans if any token is set to expire in the next day.
 
This approach had two issues - (1) one day isn't enough, because we don't run `terraform {plan|apply}` everyday, and (2) the expiration dates for the various tokens were each defined separately throughout different files and required a dev to change each one when the time for token renewal comes (see #1102 for example)

This PR refactors the token creation into a reusable module so that common attributes like the expiration date are only located in one place. This means that next time we need to renew the tokens, we just need to modify the one timestamp in this module.

### DIFF
--- a/terraform/modules/spack_gitlab/modules/spackbot_personal_access_token/main.tf
+++ b/terraform/modules/spack_gitlab/modules/spackbot_personal_access_token/main.tf
@@ -1,0 +1,33 @@
+locals {
+  # GitLab no longer allows tokens to be created with an indefinite expiration date,
+  # so we set a long expiration date and a warning date to remind us to update it.
+  token_expiration_date = "2026-04-25"
+}
+
+locals {
+  # Set the token expiration warning date to one month before the expiration date
+  # so that `terraform apply` will start failing a month before the expiration date.
+  # This should give us plenty of time to notice and update the token before it actually expires.
+
+  # Note: the largest time unit terraform's timeadd function supports is hours.
+  token_expiration_warning_date = timeadd("${local.token_expiration_date}T00:00:00Z", "-730h")
+}
+
+resource "gitlab_personal_access_token" "this" {
+  user_id    = data.gitlab_user.this.id
+  name       = var.name
+  expires_at = local.token_expiration_date
+
+  scopes = var.scopes
+
+  lifecycle {
+    precondition {
+      condition     = timecmp(timestamp(), local.token_expiration_warning_date) == -1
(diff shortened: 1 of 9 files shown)


terraform/modules/spack_gitlab/modules/spackbot_personal_access_token/main.tf
```diff
@@ -0,0 +1,33 @@
...
+  name       = var.name
+  expires_at = local.token_expiration_date
+
+  scopes = var.scopes
+
+  lifecycle {
+    precondition {
+      condition     = timecmp(timestamp(), local.token_expiration_warning_date) == -1
+      error_message = "The ${var.name} GitLab PAT will expire in less than a month. Update the expiration date as soon as possible."
+    }
```
jjnesbitt, line 27: Could we combine this with a targeted terraform command in github ci workflow, so that we can check this value automatically without needing everything in terraform to be perfect?

jjnesbitt: We discussed offline and decided this is not worth doing.

terraform/modules/spack_gitlab/modules/spackbot_personal_access_token/main.tf
```diff
@@ -0,0 +1,33 @@
+locals {
+  # GitLab no longer allows tokens to be created with an indefinite expiration date,
+  # so we set a long expiration date and a warning date to remind us to update it.
+  token_expiration_date = "2026-04-25"
```
jjnesbitt, line 4: Just to be clear, this means that all PATs created with this module expire on the same date?

mvandenburgh: Yes. That was already the case, but the date was replicated for each token. Now, all we need to do is update this one string to renew every token.

jjnesbitt (approved):
Looks great! Seems like a big improvement

PR #1112, opened 2025-05-07
protected-publish: Speed up publishing specs to the root

scottwittenburg:
This PR does two things to speed up protected publish:

1. Use `aws s3 sync ...` to download the manifests, instead of the current approach of `s3_client.download_fileobj()` for each one using multithreading
2. When copying the manifest, metadata, and archive for a spec, create only one s3 session rather than three


mvandenburgh (approved):
LGTM. It's a shame we can do this with `boto3`, but apparently that has been a requested feature for a while https://github.com/boto/boto/issues/3343 🤷‍♂️

scottwittenburg:
> It's a shame we can't do this with boto3

Yeah, that might have been nice. On the other hand, this way it will be quite easy to swap out for another cli tool like `s5cmd`, if we're still looking for performance improvements.

PR #1117, opened 2025-05-16
Add step to update spack-packages repo during snapshot

kwryankrattiger:
Add spack-package update to snapshot cron.

### DIFF
--- a/images/snapshot-release-tags/snapshot_release_tags.py
+++ b/images/snapshot-release-tags/snapshot_release_tags.py
@@ -2,12 +2,17 @@
 
 from datetime import datetime, timezone
 from github import Github, InputGitAuthor
+from typing import Optional
+import contextlib as ctx
 import json
 import os
 import re
+import sys
 import sentry_sdk
+import shutil
 import subprocess
 import tempfile
+import time
 import urllib.request
 
 sentry_sdk.init(
@@ -15,6 +20,28 @@ sentry_sdk.init(
     # so just record all transactions.
     traces_sample_rate=1.0,
 )
+
+
+GITHUB_REPO = os.environ.get("GITHUB_REPO", "spack/spack-packages")
+spackbot_author = InputGitAuthor("spackbot", "noreply@spack.io")
+
+
+def _durable_subprocess_run(*args, **kwargs):
+    """
+    Calls subprocess.run with retries/exponential backoff on failure.
+    """
+    max_attempts = 5
+    for attempt_num in range(max_attempts):
+        try:
+            return subprocess.run(*args, **kwargs, check=True)
+        except subprocess.CalledProcessError as e:
+            if attempt_num == max_attempts - 1:
+                raise e
+            print(
+                f"Subprocess failed ({e}), retrying ({attempt_num+1}/{max_attempts})",
+                file=sys.stderr,
+            )
+            time.sleep(2 ** (1 + attempt_num))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


kwryankrattiger:
I am not sure if this is the correct place to do this, but we can relocate the logic as needed.

A few questions I had while looking at this.

- Will spack core still get weekly tags
- If there are weekly tags, will there be a gitlab status to check or will we move the repo testing into github with a very very small stack to build on PRs
- If no weekly tags, should this cron just grab the latest head of `develop` for the update?
- Should we run the `spack/spack-packages:develop` snapshot on acceptance of the updated spack ref, or should it run on the spack ref. of the previous week that the packages had been developed on.

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -59,4 +87,64 @@
...
+                    old_sha_match = re.search(shasum_re, line)
+                    if old_sha_match:
+                        old_sha = old_sha_match[2]
+                        # The if shas match, don't update
+                        if old_sha == tag.sha:
+                            raise RuntimeError("sha has not changed")
+                        old_date = py_gh_repo.get_commit(old_sha).commit.committer.date
+                        new_date = py_gh_repo.get_commit(tag.sha).commit.committer.date
+                        # The if new sha is older than the old sha, don't update
+                        if old_date > new_date:
```
kwryankrattiger, line 122: I think this is safe since all commits going into develop are going to be merge commits.

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -59,4 +87,64 @@
...
+                    ofd.write(line)
+
+            with ctx.chdir(spackages):
+                commit_msg = f"""Snapshot Sync: {tag.name} {tag.sha}"""
+                _durable_subprocess_run(["git", "commit", "-a", "-m", commit_msg])
+                _durable_subprocess_run(["git", "push", "-u", "github", ])
+
+            spackages_repo = py_github.get_repo("spack/spack-packages", lazy=True)
+            # Create the PR
+            pr = spackages_repo.create_pull(
```
kwryankrattiger, line 134: Should this get an urgent tag?

images/snapshot-release-tags/snapshot_release_tags.py
```diff
@@ -59,4 +87,64 @@
...
+                head=branch_id,
+                body="""
+Automated sync commit created by the snapshot release tags cron.
+
+@spack/maintainers
+"""
+            )
+            last_commit = pr.get_commits()[pr.commits - 1]
+            # Trigger a rebuild everything on the PR to be reviewed by maintainers
+            pr.create_comment("@spackbot rebuild everything", last_commit, "", 0)
```
kwryankrattiger, line 145: I think this should work with the urgent tag, but I am not sure about timing of the events sent to spackbot and the periodicity of the sync script.

kwryankrattiger: Use a label to denote a different branch name scheme to use for this type of CI.

images/buildcache-tools/pkg/publish.py
```diff
@@ -669,7 +465,7 @@ def main():
         "-v",
         "--version",
         type=int,
-        default=2,
+        default=3,
         help=("Target layout version to publish (either 2 or 3, defaults to 2)"),
```
zackgalbreath:
```suggestion
        help=("Target layout version to publish (either 2 or 3, defaults to 3)"),
```

PR #1120, opened 2025-05-16
Always assume terraform role in aws providers

jjnesbitt:
This will ensure that all of the current `aws` provider always use a special role in AWS specifically for terraform. This eliminates the issue we were running into, where the last user that attempted to run a plan or apply would be set as the cluster creator, introducing unwanted changes.


jjnesbitt:
@mvandenburgh I wonder if there's some way to further standardize the different `aws` providers, so we don't need to specify the role explicitly in multiple places? In the future if we add another `aws` provider block and forget to include `assume_role`, this could regress.

mvandenburgh:
> @mvandenburgh I wonder if there's some way to further standardize the different `aws` providers, so we don't need to specify the role explicitly in multiple places? In the future if we add another `aws` provider block and forget to include `assume_role`, this could regress.

Good point. `AWS_ROLE_ARN` seems like it might do the trick https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html

jjnesbitt:
> Good point. `AWS_ROLE_ARN` seems like it might do the trick [docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html)

Trying this locally, it appears it does not act in the same way this PR does (still tries to change cluster owner). The [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume-role-configuration-reference) seem to state the following:

> Note that environment variables are not supported for assuming IAM roles.

It does seems that variable will work when using `assume_role_with_web_identity`, but not for the regular `assume_role` block.

PR #1122, opened 2025-05-19
Enable `FF_WAIT_FOR_POD_TO_BE_REACHABLE` on all k8s runners

mvandenburgh:
This will enable the `FF_WAIT_FOR_POD_TO_BE_REACHABLE` feature flag on all of our AWS/k8s-based GitLab runners. This feature flag makes it so a runner executor pod is only "Ready" in k8s (and thus can accept new jobs) when the underlying node has its TLS certificates all ready.

This helps avoid an issue where a runner executor pod gets scheduled on a brand new node that has no certs yet, resulting in new jobs failing with errors like these that @tgamblin saw over the weekend -
- https://gitlab.spack.io/spack/spack/-/jobs/16676859
- https://gitlab.spack.io/spack/spack/-/jobs/16676867
- https://gitlab.spack.io/spack/spack/-/jobs/16676872

See https://docs.gitlab.com/runner/configuration/feature-flags/ (ctrl-F for `FF_WAIT_FOR_POD_TO_BE_REACHABLE`)

As part of this PR, I've also moved the runner feature flags into the `runners.feature_flags` section of the TOML config. The GitLab docs mention this as an option that will prevent downstream jobs from overriding these env vars themselves; for example, with the old way, a `spack/spack` PR could set `FF_WAIT_FOR_POD_TO_BE_REACHABLE=false` in its environment and disable that feature flag.

### DIFF
--- a/k8s/production/runners/protected/graviton/3/release.yaml
+++ b/k8s/production/runners/protected/graviton/3/release.yaml
@@ -77,7 +77,13 @@ spec:
           """
 
           output_limit = 20480
-          environment = ["FF_GITLAB_REGISTRY_HELPER_IMAGE=1"]
+          environment = [
+            "FF_GITLAB_REGISTRY_HELPER_IMAGE=1",
+
+            # Avoid TLS errors when scheduling jobs on a new node
+            # https://gitlab.com/gitlab-org/gitlab-runner/-/issues/37244
+            "FF_WAIT_FOR_POD_TO_BE_REACHABLE=1",
+          ]
           [runners.kubernetes]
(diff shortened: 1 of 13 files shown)


zackgalbreath:
any reason not to enable this feature flag for our Windows runners too?

mvandenburgh:
> any reason not to enable this feature flag for our Windows runners too?

nope, my search and replace didn't catch those and I didn't factor in windows when i was double checking the runner types :sweat_smile: I made that change too

zackgalbreath (approved):

PR #1207, opened 2025-09-26
Ignore VSCode and PyCharm config files

williamjallen:
Adds common IDE config files to the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,7 @@ __pycache__/
 
 # Spack secret files
 private.key
+
+# Editor config files
+.vscode
+.idea


jjnesbitt (approved):

danlamanna:
FWIW, this is ignoring files that already exist in the repo. In general I think `.vscode` should probably be ignored though, maybe with the exception of `extensions.json`. [This post](https://stackoverflow.com/questions/32964920/what-kinds-of-source-control-settings-data-does-a-projects-vscode-folder-conta/57749909#57749909) goes into the quagmire that is trying to version control things in `.vscode`.

PR #1215, opened 2025-10-06
Update sequences on DB restore

williamjallen:
Postgres does not update sequences on `COPY` commands, meaning that our sequences remain at 1 after running `restore-db.sh`.  This PR sets the next sequence to the maximum ID for each table.

### DIFF
--- a/analytics/dev/db-backup-scripts/restore-db.sh
+++ b/analytics/dev/db-backup-scripts/restore-db.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
+set -e
+
 export PGUSER=postgres
 export PGPORT=5432
 export PGHOST=localhost
 export PGDATABASE=django
 export PGPASSWORD=postgres
-
-NULL_VALUE="<null>"
 
 
 function load_from_file {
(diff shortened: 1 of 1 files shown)


analytics/dev/db-backup-scripts/restore-db.sh
```diff
@@ -20,6 +20,18 @@ function load_from_file {
...
+        DO
+        \$$
+        BEGIN
+            IF EXISTS (SELECT 1 FROM pg_class where relname = '${table_name}_id_seq') THEN
+                PERFORM setval('${table_name}_id_seq', max(id)) FROM ${table_name};
+            END IF;
+        END;
+        \$$
+        LANGUAGE plpgsql;
+    "
```
jjnesbitt, line 34:
```suggestion

    # Fix the ID sequence after COPY FROM, if it exists
    # This has to be wrapped in an anonymous function to work around errors if the sequence does not exist.
    seq_table_name=${table_name}_id_seq
    psql -c "
        DO
        \$$
        BEGIN
            IF EXISTS (SELECT 1 FROM pg_class where relname = '$seq_table_name') THEN
                PERFORM setval('$seq_table_name', max(id)) FROM ${table_name};
            END IF;
        END;
        \$$
        LANGUAGE plpgsql;
    "
```

jjnesbitt (changes requested):
Just a minor suggestion, but looks good. Tested successfully locally.

jjnesbitt (approved):

Issue #1208, opened 2025-09-26
Track parent pipeline in `GitlabJobDataDimension`


jjnesbitt:
Right now, we only track the direct pipeline a job is associated with. However, since pipelines spin off child pipelines, we also want to know what parent pipeline a job was associated with. This should be tracked in a new `parent_pipeline_id` field on the `GitlabJobDataDimension` model. Also, we should probably rename the existing `commit_id` to `pipeline_id`.

In the future, they're may be a further `spack_pipeline_id` field added, which tracks one more level up. For now though, that's not the case.

PR #1217, opened 2025-10-07
Add parent pipeline ID to `GitlabJobDataDimension` model

williamjallen:
Closes https://github.com/spack/spack-infrastructure/issues/1208.


analytics/analytics/job_processor/dimensions.py
```diff
@@ -130,11 +130,20 @@ def create_gitlab_job_data_dimension(
     rvmatch = re.search(r"Running with gitlab-runner (\d+\.\d+\.\d+)", job_trace)
     runner_version = rvmatch.group(1) if rvmatch is not None else ""
 
+    # Defaults to this pipeline's ID if it doesn't have a parent pipeline.
+    # In the future, we may want to consider leaving the parent_pipeline_id
+    # null if there is no parent pipeline instead of setting it to its own ID.
```
jjnesbitt, line 135: I don't think we'll ever want this to be intentionally set to `null`, since we need this value to aggregate jobs by parent pipelines. The problem we're trying to solve is that when we group jobs by `pipeline_id`, it groups both top level pipelines as well as child pipelines into one metric, throwing off the value. If we set `parent_pipeline_id` to `null`, then jobs that exist directly in the parent pipeline won't get aggregated when grouping by `parent_pipeline_id`. 

I'd like to avoid speculating on how we'd change this approach until we've actually made use of it, as it can just cause more confusion otherwise.

```suggestion
    # If a job doesn't have source_pipeline, then it exists within the parent pipeline itself
```

jjnesbitt (changes requested):
We'll want to create a management command to update the existing values in the database, by making calls to the gitlab DB (specifically the `ci_sources_pipelines` table).

PR #1261, opened 2025-11-21
Improve infrastructure documentation

mvandenburgh:
Adds some documentation for the cloud infrastructure, particularly Terraform and how Flux works.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,57 @@
+# Spack Infrastructure Terraform Configuration
+
+This directory contains the Terraform configuration for the Spack infrastructure. All changes to the AWS cloud infrastructure should be made via this configuration.
+
+## Architecture Overview
+
+We use Terraform to manage the following aspects of the infrastructure:
+
+- **AWS Cloud Infrastructure**: Core AWS infrastructure, including VPC, subnets, NAT gateways, EKS cluster, RDS databases, S3 buckets, etc.
+    - For this we use the AWS provider (https://registry.terraform.io/providers/hashicorp/aws/latest/docs), as well as various commmunity modules from the [terraform-aws-modules](https://github.com/terraform-aws-modules) organization (https://registry.terraform.io/namespaces/terraform-aws-modules).
+- **Flux**: GitOps deployment tool for managing the Kubernetes cluster
+    - For this we use the flux provider (https://registry.terraform.io/providers/fluxcd/flux/latest/docs).
+    - The provider does the following:
+        - Creates a Flux installation in the EKS cluster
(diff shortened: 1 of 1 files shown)


terraform/README.md
```diff
@@ -0,0 +1,61 @@
...
+- **Kubernetes**: Kubernetes yaml manifests for the EKS cluster
+    - For this we use the kubectl provider (https://registry.terraform.io/providers/alekc/kubectl/latest/docs) and the helm provider (https://registry.terraform.io/providers/hashicorp/helm/latest/docs).
+    - We try to use Flux/GitOps to manage the Kubernetes cluster (i.e., by committing k8s yaml manifests to `k8s/`) as much as possible, but sometimes it is cleaner to manage some manifests with the Terraform kubectl provider. This is true mainly for manifests that have to reference resources that are managed by Terraform. A concrete examples of this is using [the `kubectl_manifest` resource to deploy a `Secret` resource for RDS credentials](./modules/spack_aws_k8s/gitlab_db.tf), which enables us to reference the RDS credentials using Terraform interpolation instead of hardcoding them in a manifest.
```
jjnesbitt: This link doesn't seem to resolve (may be because it's still a PR). Why not use a permalink? That way, even if the location changes, we can still refer to it? I also included a line highlight.
  
    
      
        Suggested change
      
    
    
      
          
            
                - We try to use Flux/GitOps to manage the Kubernetes cluster (i.e., by committing k8s yaml manifests to `k8s/`) as much as possible, but sometimes it is cleaner to manage some manifests with the Terraform kubectl provider. This is true mainly for manifests that have to reference resources that are managed by Terraform. A concrete examples of this is using [the `kubectl_manifest` resource to deploy a `Secret` resource for RDS credentials](./modules/spack_aws_k8s/gitlab_db.tf), which enables us to reference the RDS credentials using Terraform interpolation instead of hardcoding them in a manifest.
          
          
            
                - We try to use Flux/GitOps to manage the Kubernetes cluster (i.e., by committing k8s yaml manifests to `k8s/`) as much as possible, but sometimes it is cleaner to manage some manifests with the Terraform kubectl provider. This is true mainly for manifests that have to reference resources that are managed by Terraform. A concrete examples of this is using [the `kubectl_manifest` resource to deploy a `Secret` resource for RDS credentials](https://github.com/spack/spack-infrastructure/blob/70db5aca32f396a2be3c64bd12d61e3d32ab0c77/terraform/modules/spack_aws_k8s/gitlab_db.tf#L146-L162), which enables us to reference the RDS credentials using Terraform interpolation instead of hardcoding them in a manifest.

jjnesbitt (approved):

PR #1294, opened 2026-01-08
Update ci-prune-buildcache to also prune the stack-specific buildcaches

zackgalbreath:


zackgalbreath:
~Note that the relevant cron job is still intentionally disabled. My plan is to monitor a manual run of this job after this PR and https://github.com/spack/spack/pull/51818 are merged to verify that it is behaving as expected. If everything looks good then we'll re-enable automatic scheduled pruning.~

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -32,30 +30,54 @@ since_date=$(date --iso-8601 -d "${PRUNE_SINCE_DAYS} days ago")
...
+  --since-days "${PRUNE_SINCE_DAYS}"
+echo "Keep lists created with $(wc -l < develop_keeplist.txt) total hashes"
+
+prune_stack() {
+  local keeplist_file=$1
+  stack="${keeplist_file%_keeplist.txt}"
+  echo "Start pruning process for $stack buildcache"
+
+  echo "Contents of $keeplist_file:"
+  cat $keeplist_file
```
kwryankrattiger: this will likely be long enough to make the log unreadable. Is it possible to archive the keep lists to s3?

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -32,30 +30,54 @@ since_date=$(date --iso-8601 -d "${PRUNE_SINCE_DAYS} days ago")
...
+  if [[ "$stack" == "develop" ]]; then
+    spack mirror add "${stack}" "${BUILDCACHE_URL}"
+  else
+    spack mirror add "${stack}" "${BUILDCACHE_URL}/${stack}"
+  fi
+  echo "Adding mirror $stack... done!"
+
+  # Run pruning with keeplist
+  echo "Running buildcache prune for $stack..."
+  spack --debug python prune.py --mirror "${stack}" --keeplist "${keeplist_file}"
```
kwryankrattiger, line 55: Is there a reason to keep this as `spack python prune.py` vs `spack buildcache prune --keeplist <list_file> <mirror>`

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -32,30 +30,54 @@ since_date=$(date --iso-8601 -d "${PRUNE_SINCE_DAYS} days ago")
...
+  echo "Running buildcache prune for $stack... done!"
+
+  # Update the mirror index
+  echo "Updating mirror index for $stack..."
+  spack buildcache update-index "${stack}"
```
kwryankrattiger, line 60: Doesn't have to be here, but a while back we had briefly discussed doing this as part of the pruning command optionally so that the index would be update _before_ the pruning to exclude the pruned specs. Looking at the pruning function in spack it seems that is still not there, just something to remember for later.

k8s/production/custom/prune-buildcache/cron-jobs.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
   name: prune-buildcache-v3
   namespace: custom
 spec:
-  suspend: true
   schedule: "0 0 * * 6"
```
kwryankrattiger, line 8: This could maybe be more frequent, mostly for testing initially.

images/ci-prune-buildcache/entrypoint.sh
zackgalbreath: IIRC the only reason we do it this way is for [sentry integration](https://github.com/spack/spack-infrastructure/blob/a5b0bb5c2b3942c709c53bd0f2b20e1327ea24bb/images/ci-prune-buildcache/prune.py#L15)

images/ci-prune-buildcache/entrypoint.sh
zackgalbreath: `containerLogMaxSize` is [10 Mi by default](https://kubernetes.io/docs/concepts/cluster-administration/logging/#log-rotation) . I'll kick off another local prune to see how we currently measure up against this limit.

images/ci-prune-buildcache/entrypoint.sh
kwryankrattiger: ah that makes sense, I am fine with that then

images/ci-prune-buildcache/entrypoint.sh
```diff
@@ -3,8 +3,6 @@ set -euo pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-echo "Input: $@"
-
 # Clone spack
 echo "Cloning spack..."
 git clone --depth 1 https://github.com/spack/spack.git
```
kwryankrattiger: We should probably move this to the image creation and pin the spack commit/version.

kwryankrattiger (approved):
LGTM, thanks!

PR #1305, opened 2026-02-06
Separate NodePool for notary pods

zackgalbreath:


k8s/production/karpenter/provisioners/runners/x86_64/notary/provisioners.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+  # Terminate nodes after 5 minutes of idle time
+  disruption:
+    consolidationPolicy: WhenEmpty
+    consolidateAfter: 5m
+
+  # Resource limits for this provisioner only
+  limits:
+    cpu: 40
+    memory: 128G
```
jjnesbitt: These numbers don't seem _wrong_ to me, but I also have no way to evaluate if they're accurate. Maybe we should just keep an eye out.

PR #1311, opened 2026-02-26
Use kustomize to de-duplicate runner provisioner YAML

zackgalbreath:


jjnesbitt:
I feel like the additional complexity introduced by this change isn't worth the benefit we get. Now instead of looking at the full configuration of a provisioner, we navigate to a `patch.yaml` and just see the kustomization patches, which are difficult to understand without context.

I agree the existing approach of duplicating certain lines of yaml isn't ideal, but it doesn't really seem to hurt us much, aside from that.

Thoughts @mvandenburgh?

PR #1315, opened 2026-03-09
Windows: Stop using PSScriptRoot

johnwparent:
In some contexts PSScriptRoot is not set. So far given usage it appears to be safe, but in upcoming Gitlab updates, and for other runner configurations, this value isn't set during pre-script, and will break the pre-build behavior and prevent proper upload.

Use HOME which is always set and writable


johnwparent:
Can't tag reviewers so @zackgalbreath @kwryankrattiger

PR #1333, opened 2026-03-25
Move pipeline experiments to private repo

kwryankrattiger:
The scott/pipeline-experiments project was great for the days of single repo, but with split repo there are different workflows that may need testing. To make things consistent pipeline experiments is now a group, and then each project is made available under the top level group to support their own testing variables.


terraform/modules/spack_gitlab/runner_iam.tf
```diff
@@ -23,10 +24,14 @@ locals {
         "project_path:${data.gitlab_project.spack_packages.path_with_namespace}:ref_type:branch:ref:releases/v*",
         "project_path:${data.gitlab_project.spack_packages.path_with_namespace}:ref_type:tag:ref:develop-*",
         "project_path:${data.gitlab_project.spack_packages.path_with_namespace}:ref_type:tag:ref:v*",
-        "project_path:${data.gitlab_project.scott_sidecar.path_with_namespace}:ref_type:branch:ref:develop",
-        "project_path:${data.gitlab_project.scott_sidecar.path_with_namespace}:ref_type:branch:ref:releases/v*",
-        "project_path:${data.gitlab_project.scott_sidecar.path_with_namespace}:ref_type:tag:ref:develop-*",
-        "project_path:${data.gitlab_project.scott_sidecar.path_with_namespace}:ref_type:tag:ref:v*"
+        "project_path:${data.gitlab_project.spack_experiments.path_with_namespace}:ref_type:branch:ref:develop",
```
jjnesbitt: I know this is just updating what was already there, but I don't like that we're sharing permissions to push to both the PR and protected binary mirrors with these experimental repos. It would be pretty easy for someone to push garbage directly to develop, and then push that to the binary mirror. I'm not sure what complications that introduces, but just thought I'd point out my observation.

This doesn't need to block this PR.

jjnesbitt (approved):
Aside from my other comment regarding the existing structure, I don't see anything wrong with this.

PR #1337, opened 2026-03-31
Enable consolidation on Gitlab nodes

jjnesbitt:
Since gitlab pods can be scaled in response to load, we're in a situation where new nodes can be spun up to handle this load, but it's difficult to fully drain the node, to allow it to terminate.


jjnesbitt:
@mvandenburgh After looking into it, the runners are placed on the `base` nodepool, while the gitlab pods are placed on the `gitlab` nodepool. So this change will only affect the nodes that core gitlab is running on.

jjnesbitt:
This is applied in both prod and staging

PR #1339, opened 2026-04-10
Update clear-admin-keys service account

zackgalbreath:
Update this service account to be able to list members of the 'Custodians' group. This change was originally overlooked in #1321 .


jjnesbitt (approved):
LGTM, ran a terraform plan and nothing unexpected showed up.

PR #1341, opened 2026-04-15
Disable snapshot cron

kwryankrattiger:
This cron is wrong and the rest of the infra for it does not currently work. Disable to remove noise.

### DIFF
--- a/k8s/production/custom/snapshot-release-tags/cron-jobs.yaml
+++ b/k8s/production/custom/snapshot-release-tags/cron-jobs.yaml
@@ -5,6 +5,7 @@ metadata:
   name: snapshot-release-tags
   namespace: custom
 spec:
+  suspend: true
   schedule: "0 1 * * 0" # 1am on Sunday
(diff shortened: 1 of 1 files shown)


k8s/production/custom/snapshot-release-tags/cron-jobs.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
   name: snapshot-release-tags
   namespace: custom
 spec:
-  schedule: "0 1 * * 0" # 1am on Sunday
+  # schedule: "0 1 * * 0" # 1am on Sunday
+  schedule: "0 5 31 2 *" # Disabled - No such date as Feb 31
```
mvandenburgh: You should be able to accomplish the same thing by setting `suspend: true` here. https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-suspension

jjnesbitt (approved):

PR #1342, opened 2026-04-20
SpackCIBridge: improve error handling

zackgalbreath:
This commit address a couple error conditions we encountered while setting up the SpackCIBridge for papi. In particular:

* More gracefully handle UTF-8 decoding errors when attempting to parse the merge commit message. We also attempt to avoid the issue entirely by only parsing the first 128 characters of the message.

* Better handling for the case where the GitHub PR has a merge commit when we're not using the `--main-branch` option.


kwryankrattiger (approved):
lgtm

PR #1345, opened 2026-04-23
Remove old users and policies for source mirror access

jjnesbitt:
Follow-on  to #1344


jjnesbitt:
@alecbcs Once you've confirmed that the new auth workflow is working, I think we can get rid of these users and policies (the existing auth workflow).

alecbcs:
spack/spack-packages#4373 has now merged so we should be good to go ahead and remove the old users

PR #1369, opened 2026-06-05
Restrict gitlab nodepool deprovisioning

### LINKED ISSUE #1366: Adjust Karpenter Disruption for Gitlab Provisioner
We've been seeing an uptick in `503 Service Unavailable` response from Gitlab. This may be due to consolidation being now enabled. We should limit the consolidation events to happen during non-peak hours (10pm to 2am Eastern Time).

Karpenter Docs
```yaml
  # Disruption section which describes the ways in which Karpenter can disrupt and replace Nodes
  # Configuration in this section constrains how aggressive Karpenter can be with performing operations
  # like rolling Nodes due to them hitting their maximum lifetime (expiry) or scaling down nodes to reduce cluster cost
  disruption:
    # Describes which types of Nodes Karpenter should consider for consolidation
    # If using 'WhenEmptyOrUnderutilized', Karpenter will consider all nodes for consolidation and attempt to remove or replace Nodes when it discovers that the Node is empty or underutilized and could be changed to reduce cost
    # If using `WhenEmpty`, Karpenter will only consider nodes for consolidation that contain no workload pods
    consolidationPolicy: WhenEmptyOrUnderutilized | WhenEmpty

    # The amount of time Karpenter should wait to consolidate a node after a pod has been added or removed from the node.
    # You can choose to disable consolidation entirely by setting the string value 'Never' here
    consolidateAfter: 1m | Never # Added to allow additional control over consolidation aggressiveness

    # Budgets control the speed Karpenter can scale down nodes.
    # Karpenter will respect the minimum of the currently active budgets, and will round up
    # when considering percentages. Duration and Schedule must be set together.
    budgets:
    - nodes: 10%
    # On Weekdays during business hours, don't do any deprovisioning.
    - schedule: "0 9 * * mon-fri"
      duration: 8h
      nodes: "0"
```

jjnesbitt:
Closes #1366

### DIFF
--- a/k8s/production/karpenter/provisioners/gitlab/provisioner.yaml
+++ b/k8s/production/karpenter/provisioners/gitlab/provisioner.yaml
@@ -37,3 +37,19 @@ spec:
   disruption:
     consolidationPolicy: WhenEmptyOrUnderutilized
     consolidateAfter: 5m
+    budgets:
+      # ┌───────────── minute (0 - 59)
+      # │ ┌───────────── hour (0 - 23)
+      # │ │ ┌───────────── day of the month (1 - 31)
+      # │ │ │ ┌───────────── month (1 - 12)
+      # │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday;
+      # │ │ │ │ │                                   7 is also Sunday on some systems)
+      # │ │ │ │ │                                   OR sun, mon, tue, wed, thu, fri, sat
+      # │ │ │ │ │
+      # * * * * *
+      #
+      # Only deprovision between 4AM and 8AM UTC
+      # (apply this in reverse by disallowing deprovisioning from 8AM to 4AM UTC)
+      - schedule: "0 8 * * *"
+        duration: 20h
+        nodes: "0"


k8s/production/karpenter/provisioners/gitlab/provisioner.yaml
```diff
@@ -37,3 +37,19 @@ spec:
...
+      # │ │ ┌───────────── day of the month (1 - 31)
+      # │ │ │ ┌───────────── month (1 - 12)
+      # │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday;
+      # │ │ │ │ │                                   7 is also Sunday on some systems)
+      # │ │ │ │ │                                   OR sun, mon, tue, wed, thu, fri, sat
+      # │ │ │ │ │
+      # * * * * *
+      #
+      # Only deprovision between 4AM and 7AM UTC
+      # (apply this in reverse by disallowing deprovisioning from 8AM to 3AM UTC)
```
zackgalbreath: This might be quibbling over an inclusive vs. exclusive time range, but these comments seem like they're off-by-one?

This schedule defines a 20 hour window starting at 8am UTC during which deprovisioning is disabled.

This means that deprovisioning can only occur between 4am and 8am UTC.

jjnesbitt: Good catch, you're right. It's important to make sure these comments reflect reality. I've fixed it.

zackgalbreath (approved):

PR #1374, opened 2026-06-19
gitlab: add startupProbe and loosen readinessProbe

zackgalbreath:
We've recently noticed webservice pods getting killed by kubelet before Puma finishes booting.

This commits add a startupProbe to give newly created webservice pods five minutes to complete the Puma bootup sequence before enabling the readiness and liveness probes.

This commit also increases the webservice readiness timeout from 10s to 20s.

### DIFF
--- a/k8s/production/gitlab/release.yaml
+++ b/k8s/production/gitlab/release.yaml
@@ -154,6 +154,23 @@ spec:
             secretName: tls-gitlab-webservice
         minReplicas: 4
         maxReplicas: 32
+        deployment:
+          # Use a startupProbe to gate liveness/readiness until Puma finishes
+          # booting.
+          startupProbe:
+            httpGet:
+              path: /-/readiness
+              port: 8080
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            failureThreshold: 30  # Give Puma 5 minutes (30 * 10s) to boot
+            timeoutSeconds: 5
+          readinessProbe:
+            initialDelaySeconds: 0
+            periodSeconds: 5
+            timeoutSeconds: 5
+            successThreshold: 1
+            failureThreshold: 4   # Consider pod unready after 20 seconds (5 * 4)
         resources:
           # Prometheus queries to check for usage values
           # cpu:


zackgalbreath:
Here's an example pod that restarted recently:
```
gitlab-webservice-default-674c65ffd7-j5cbc                    2/2     Running     1 (74m ago)     78m
```

Describing this pod, we see the previous webservice container was killed due to "Error" after ~3 minutes.
```
    Last State:     Terminated
      Reason:       Error
      Exit Code:    137
      Started:      Fri, 19 Jun 2026 10:06:27 -0400
      Finished:     Fri, 19 Jun 2026 10:09:40 -0400
```

Looking at the logs from this pod's previous webservice container, we can see when Puma started booting up:
```
{"timestamp":"2026-06-19T14:06:34.479Z","pid":25,"message":"Puma starting in cluster mode..."}
```

The last log from this container shows that it was still checking postgres partitions 160 seconds later.
```
{"component": "gitlab","subcomponent":"application_json","severity":"INFO","time":"2026-06-19T14:09:10.471Z","message":"Checking state of dynamic postgres partitions","table_name":"backup_vulnerability_reads","connection_name":"main"}
```

As above, the container was killed 30 seconds after this log entry.

zackgalbreath:
This change applied cleanly on our staging cluster.

```
Containers:
  webservice:
    State:          Running
      Started:      Fri, 19 Jun 2026 11:39:46 -0400
    Ready:          True
    Liveness:   http-get http://:8080/-/liveness delay=20s timeout=30s period=60s #success=1 #failure=3
    Readiness:  http-get http://:8080/-/readiness delay=0s timeout=5s period=5s #success=1 #failure=4
    Startup:    http-get http://:8080/-/readiness delay=30s timeout=5s period=10s #success=1 #failure=30 
```

jjnesbitt (approved):
This looks good! I'm still trying to understand why exactly the startup takes so much time, and if we can improve it, but this should improve the startup behavior. If we are ever able to improve startup time we can revisit.

PR #1376, opened 2026-06-22
SpackCIBridge: don't post status when required label is missing

zackgalbreath:
Prior to this commit, the sync script would post a pending status to PRs on GitHub that are missing the project-specific label that's required for GitLab CI testing to start.

Users found this confusing because it makes it appear as if CI has not finished running yet. Instead we now opt not to post any status at all in this situation A skipped status would be better, but that is only supported from "checks" (GitHub Actions), not "statuses".

### DIFF
--- a/.github/images.yml
+++ b/.github/images.yml
@@ -1,6 +1,6 @@
 images:
   - path: ./images/gh-gl-sync
-    image: ghcr.io/spack/ci-bridge:0.0.50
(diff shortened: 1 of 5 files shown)


kwryankrattiger (approved):
LGTM

ax3l:
Thank you, Zack!!

PR #1378, opened 2026-06-29
Restrict gitlab OIDC role assume via source IP

jjnesbitt:


terraform/modules/spack_gitlab/runner_iam.tf
```diff
@@ -82,6 +82,15 @@ data "aws_iam_policy_document" "gitlab_oidc_assume_role" {
...
     }
+
+    condition {
+      test     = "IpAddress"
+      variable = "aws:SourceIp"
+      values = concat(
+        [for ip in var.nat_public_ips : "${ip}/32"],
+        ["128.223.202.0/24"] # UO's IP block
+      )
+    }
```
zackgalbreath, line 94: This would further restrict who can assume this role (based on their IP address), but if the resulting temporary credentials were "shared" afterwards, the IP address restriction would no longer apply, right? This isn't necessarily a bad thing, I just want to make sure I understand what's going on here 😅 .

jjnesbitt: That's correct, in order to fully prevent the action via IP we'd need to add the condition to the policy itself. Luckily the policies in question are self managed (not off the shelf AWS policies), so it should be possible, but it'll take some terraform rework.

jjnesbitt: After trying this out, I'm not sure restricting the `PutObject`/`DeleteObject` actions via `SourceIP` is a sensible condition we can restrict by. It may be somehow possible, but I think for now, restricting the role assumption is sufficient.

jjnesbitt:
This has been applied in staging and prod successfully

PR #1379, opened 2026-06-29
Update runners to get secrets after build

kwryankrattiger:
`post_build_script` runs in the same shell context as the main script, but after the main script is completed. Run the OIDC script and trust GPG secrets after build.


kwryankrattiger:
It would be nice to have https://github.com/spack/spack/pull/52615 merged and updated in packages before this lands.

PR #1381, opened 2026-07-01
Install missing PyGithub dependency for prune_pr_mirrors.py

zackgalbreath:


zackgalbreath:
This should address the following failure we're seeing in our pruning job:

```
Running buildcache prune for develop... done!
Updating mirror index for develop...
Updating mirror index for develop... done!

Pruning PR Mirrors
Traceback (most recent call last):
  File "/app/prune_pr_mirrors.py", line 9, in <module>
    import github
ModuleNotFoundError: No module named 'github'
```

kwryankrattiger (approved):

PR #1382, opened 2026-07-01
Configure our RDS proxy to avoid session pinning

zackgalbreath:
By default, RDS Proxy pins a client session to a dedicated backend connection whenever a SET statement is executed.  Rails' pg gem issues SET statements on every connection, thwarting our attempts to multiplex connections using RDS proxy. As a result, the proxy uses a separate backend connection for each Puma/Sidekiq thread instead of pooling them.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy-pinning.html#rds-proxy-pinning.postgres

This commit aims to improve this situation by using a *session pinning filter* to instruct the proxy to ignore SET statements when determining whether a session should be pinned or not.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy-best-practices.configuration.html

### DIFF
--- a/terraform/modules/spack_aws_k8s/gitlab_db.tf
+++ b/terraform/modules/spack_aws_k8s/gitlab_db.tf
@@ -111,6 +111,13 @@ module "gitlab_db_proxy" {
   engine_family = "POSTGRESQL"
   debug_logging = false
 
+  # Without this, RDS Proxy pins a client session to a dedicated backend
+  # connection as soon as it sees a SET statement, which Rails' pg gem issues
+  # on every connection checkout (e.g. statement_timeout). That defeats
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Closing as I don't think this is the right fix. In particular, I don't think either of these two conditions apply for us:

> This option is only safe in the following situations:
> 
> 1. The SET statements are no-ops, such as setting a system variable to a value identical with the server default.
> 
> 2. The SET statements and subsequent queries are part of the same transaction, and every transaction sets up its own state fully independently so that it's not affected by variables set by other transactions.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy-best-practices.configuration.html

PR #1383, opened 2026-07-01
Enable WAF in count mode

jjnesbitt:
Related to #1372, #1377


jjnesbitt:
FYI I've tested this in staging and successfully run a job.

jjnesbitt:
This is also applied in prod

PR #1387, opened 2026-07-06
gh-gl-sync: Add option for using a gitlab oauth token

kwryankrattiger:
Passing private base64 encoded ssh keys doesn't make sense in all context. When running the sync script as a cron from Gitlab itself using the temporary job token is better practice.

### DIFF
--- a/images/gh-gl-sync/SpackCIBridge.py
+++ b/images/gh-gl-sync/SpackCIBridge.py
@@ -796,7 +796,7 @@ on a commit of the main branch that is newer than the latest commit tested by Gi
 
     args = parser.parse_args()
 
-    gl_url = urllib.parser.urlparse(args.gitlab_repo)
+    gl_url = urllib.parse.urlparse(args.gitlab_repo)
 
     gitlab_ssh_key_base64 = os.getenv("GITLAB_SSH_KEY_BASE64")
     if gl_url.scheme == "ssh" and gitlab_ssh_key_base64 is None:
--- a/images/gh-gl-sync/SpackCIBridge.py
+++ b/images/gh-gl-sync/SpackCIBridge.py
@@ -41,6 +41,15 @@ def _durable_subprocess_run(*args, **kwargs):
             time.sleep(2 ** (1 + attempt_num))
 
 
+def _write_secret(file: str, content: str):
+    # Create the file with the oatuh token
+    with open(file, "wb") as fd:
+        fd.write(b"")
+    os.chmod(file, 0o600)
+    with open(file, "wb") as fd:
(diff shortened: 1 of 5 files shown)


images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -751,9 +796,15 @@ def sync(self):
 
     args = parser.parse_args()
 
-    ssh_key_base64 = os.getenv("GITLAB_SSH_KEY_BASE64")
-    if ssh_key_base64 is None:
-        raise Exception("GITLAB_SSH_KEY_BASE64 environment is not set")
+    gl_url = urllib.parser.urlparse(args.gitlab_repo)
```
hainest: The python lib on `ghcr.io/spack/ci-bridge:0.0.44` says this is


```suggestion
    gl_url = urllib.parse.urlparse(args.gitlab_repo)
```

hainest:
When using `ghcr.io/spack/ci-bridge:0.0.44`, I get an error that `Auth` is not a member of `github`. Switching to 0.0.51 of the container works fine. I'm assuming we just need to update the default value for `sync_image` in dav-sdk/gh-gl-sync/github-sync?

hainest:
@kwryankrattiger Is https://gitlab.spack.io/dav-sdk/gh-gl-sync the official repo for the github-sync component?

kwryankrattiger:
> Is https://gitlab.spack.io/dav-sdk/gh-gl-sync the official repo for the github-sync component?

yes, afaik that is the only place it is published now.

images/gh-gl-sync/SpackCIBridge.py
```diff
@@ -101,8 +110,44 @@ def cleanup():
...
+        if not url.scheme.startswith("http"):
+            raise Exception(f"Cannot configure oauth for {url.scheme}")
+
+        # Inject the user name and OAuth token into the URL
+        if ":" in token:
+            url = url._replace(netloc=f"{token}@" + url.netloc)
+        else:
+            url = url._replace(netloc=f"spackbot:{token}@" + url.netloc)
+
+        _write_secret(".git-credentials", urllib.parse.urlunparse(url) + "\n")
```
hainest, line 140: We need to update the gitlab remote repo name name to match what's in the credentials file. I'm not entirely certain why this is needed because everything I've seen indicates that the `https://gitlab.spack.io` should be enough to match. :shrug:

```suggestion
        self.gitlab_repo = urllib.parse.urlunparse(url)

        _write_secret(".git-credentials", self.gitlab_repo + "\n")
```

hainest (changes requested):

PR #1415, opened 2026-07-22
Add gitlab-orphaned-job-canceller cronjob

zackgalbreath:
We recently noticed some GitLab CI jobs get stuck in a "Running" state beyond their normal job timeout. We traced this back to jobs running on our Kubernetes runners that lose their backing pod without GitLab ever finding out. In at least one instance, this was caused by Karpenter wrongly deleting a node it thinks is empty: kubernetes-sigs/karpenter#2916.

This commit adds a new cronjob to mitigate this issue. It correlates running jobs against Kubernetes pods and cancels those whose pod is missing or already terminal. This new cronjob only cancels jobs (it doesn't restart them), and it only ever touches jobs from Spack group runners (not instance runners). This corresponds to our Kubernetes runners but not our on-premises runners at the University of Oregon.


kwryankrattiger:
It would also be good to include canceling any jobs in general that exceed the job timeout, regardless of where they are started from.

images/gitlab-orphaned-job-canceller/main.py
```diff
@@ -0,0 +1,194 @@
...
+        # (e.g. ~/.kube/config) so this can be run locally too.
+        config.load_kube_config()
+    v1 = client.CoreV1Api()
+
+    projects = [p.strip() for p in args.projects.split(",") if p.strip()]
+
+    failed_projects = []
+    for project in projects:
+        try:
+            cancel_orphaned_jobs(v1, project, args.grace_period_minutes)
```
kwryankrattiger: Is it worth collecting these jobs and raising an error with them for sentry to report?

zackgalbreath:
> It would also be good to include canceling any jobs in general that exceed the job timeout, regardless of where they are started from.

I like this idea, but I don't think the GitLab API actually tells us when a job would timeout 🫠 
https://docs.gitlab.com/api/jobs/#retrieve-a-job-by-job-id

images/gitlab-orphaned-job-canceller/main.py
```diff
@@ -0,0 +1,194 @@
...
+    return runner.get("runner_type") == KUBERNETES_RUNNER_TYPE
+
+
+def is_orphaned(pod_index, job, grace_period):
+    """A running job is orphaned if it's running on one of our own
+    Kubernetes-executor runners, has been running longer than the grace
+    period (to avoid racing normal pod-scheduling delays), and its backing
+    pod either no longer exists, or has already reached a terminal phase
+    without GitLab having found out."""
+    if not is_kubernetes_executor_job(job):
```
kwryankrattiger, line 160: I think we can also check here if the runtime is longer than the expected max timeout of 12h?

Then, jobs orphaned by any non-k8s runners can be cancelled. I don't think I have seen these, but it seems to be possible during updates/unexpected downtime this type of thing can happen.

images/gitlab-orphaned-job-canceller/main.py
```diff
@@ -0,0 +1,194 @@
...
+    return index
+
+
+def is_kubernetes_executor_job(job):
+    """Only jobs picked up by one of our own group-level Kubernetes-executor
+    runners can ever have a backing pod in this cluster."""
+    runner = job.get("runner")
+    if not runner:
+        return False
+    return runner.get("runner_type") == KUBERNETES_RUNNER_TYPE
```
kwryankrattiger: I don't love this check. Discussing offline to figure out if there is a better way to do this...

jjnesbitt:
Since this doesn't retry any jobs, if a job that _would_ have been retried on failure is cancelled due to overrun, it won't be retried at all by any other service, including gitlab. Correct?

Are we just assuming these jobs will be retried manually? It may be hard to clearly identify which jobs require this, since they won't be obviously overrunning their timeout.

images/gitlab-orphaned-job-canceller/main.py
zackgalbreath: The concern here is that we may add on-prem group runners in the future and this check will suddenly become invalid. The alternative we came up with is to identify our cloud runners by their description.

zackgalbreath:
> Since this doesn't retry any jobs, if a job that _would_ have been retried on failure is cancelled due to overrun, it won't be retried at all by any other service, including gitlab. Correct?

Yes, that is my understanding as well.

> Are we just assuming these jobs will be retried manually? It may be hard to clearly identify which jobs require this, since they won't be obviously overrunning their timeout.

The issue that this PR is mitigating is that jobs are stuck in a running state indefinitely. Because of our use of deferred pipelines, when this happens on a develop pipeline (like it did this morning) it has the affect of preventing any new PRs from getting tested/merged until we manually intervene.

At the risk of stating the obvious, jobs that run forever don't get automatically retried. That being said, I can look into adding auto-retry logic to this new cronjob if you think that's a good idea.

kwryankrattiger:
> I can look into adding auto-retry logic to this new cronjob if you think that's a good idea.

The 30 minute grace period probably handles most of these cases where gitlab would have retried the job itself, but since this is going to catch issues much earlier the pipeline is probably still useful the the developer so retrying would be good.

jjnesbitt:
> The issue that this PR is mitigating is that jobs are stuck in a running state indefinitely. Because of our use of deferred pipelines, when this happens on a develop pipeline (like it did this morning) it has the affect of preventing any new PRs from getting tested/merged until we manually intervene.
> 
> At the risk of stating the obvious, jobs that run forever don't get automatically retried. That being said, I can look into adding auto-retry logic to this new cronjob if you think that's a good idea.

I was mainly just trying to make sure we weren't taking a half-measure w.r.t the mitigation. Since the goal is to allow deferred pipelines to run, I think limiting the scope as you've done is perfectly fine.

images/gitlab-orphaned-job-canceller/main.py
zackgalbreath: Good idea, I added a sentry warning for orphaned jobs that get canceled & retried, and made it an error if we exhaust our retry cap. @jjnesbitt FYI. I think this is a rare enough defect that it won't flood sentry but I wanted to make you aware.

zackgalbreath:
> I can look into adding auto-retry logic to this new cronjob if you think that's a good idea.

Our f2f conversation about canceled jobs being confusing to users was enough to convince me that this is probably a good idea. This script now attempts to retry canceled jobs up to a limit of 3 total retries per job.

PR #1434, opened 2026-08-13
Encode Gitlab group/projects and variables into Terraform

jjnesbitt:


terraform/staging/versions.tf
```diff
@@ -40,3 +40,8 @@ provider "aws" {
     role_arn = "arn:aws:iam::588562868276:role/terraform-role"
   }
 }
+
+provider "gitlab" {
+  base_url = "https://gitlab.staging.spack.io"
+  token    = var.gitlab_token
+}
```
mvandenburgh, line 47: Is this needed? It wasn't there before, and it production doesn't have it either.

terraform/staging/gitlab.tf
```diff
@@ -0,0 +1,140 @@
...
+#
+# These are set on the project rather than committed to testing-branch so that
+# they also cover pipelines on develop, which would otherwise push to the
+# production buildcache.
+resource "gitlab_project_variable" "spack_packages_binary_mirrors" {
+  for_each = {
+    PR_MIRROR_FETCH_DOMAIN        = "s3://spack-binaries-prs-staging"
+    PR_MIRROR_PUSH_DOMAIN         = "s3://spack-binaries-prs-staging"
+    PROTECTED_MIRROR_FETCH_DOMAIN = "s3://spack-binaries-staging"
+    PROTECTED_MIRROR_PUSH_DOMAIN  = "s3://spack-binaries-staging"
```
mvandenburgh, line 74: We could add these buckets to https://github.com/spack/spack-infrastructure/blob/main/terraform/modules/spack_aws_k8s/outputs.tf, and then reference them using variables instead of hardcoding.

mvandenburgh (commented):
Some suggestions, though I don't think any are necessarily blocking for merging this - as long as these are accurately reflecting the state of staging, it does what we need it to for recreating the staging cluster. 
Plus, I'd like to revisit this entire configuration after we recreate staging, to see if there's opportunities to consolidate these resouces into the `spack_gitlab` module for cleaner reuse between production and staging.

mvandenburgh (dismissed):

terraform/staging/versions.tf
jjnesbitt: We discussed offline but just for posterity, it's needed because we define new gitlab resources in the root staging terraform config, which production doesn't have.

mvandenburgh (dismissed):

mvandenburgh (approved):

PR #1435, opened 2026-08-13
Grant prune-buildcache role access to spack-binaries-prs bucket

zackgalbreath:
This fixes AccessDenied errors we were seeing in the prune-buildcache-v3 CronJob when attempting to s3:DeleteObject against the spack-binaries-prs bucket.

This was caused by the fact that the relevant IAM policy only granted this action on the develop/* prefix of the protected mirror.

Resolve this error by adding ListBucket/PutObject/DeleteObject on the PR binary mirror bucket to the same statement.

### DIFF
--- a/terraform/modules/spack_aws_k8s/iam_service_accounts.tf
+++ b/terraform/modules/spack_aws_k8s/iam_service_accounts.tf
@@ -15,6 +15,16 @@ module "build_cache_pruner" {
             "s3:DeleteObject"
           ],
           "Resource" : "${module.protected_binary_mirror.bucket_arn}/develop/*"
+        },
+        {
+          "Effect" : "Allow",
+          "Action" : "s3:ListBucket",
+          "Resource" : module.pr_binary_mirror.bucket_arn
+        },
+        {
+          "Effect" : "Allow",
+          "Action" : "s3:DeleteObject",
+          "Resource" : "${module.pr_binary_mirror.bucket_arn}/*"
         }
       ]
     }),


terraform/modules/spack_aws_k8s/iam_service_accounts.tf
```diff
@@ -12,9 +12,14 @@ module "build_cache_pruner" {
...
+            "s3:ListBucket"
           ],
-          "Resource" : "${module.protected_binary_mirror.bucket_arn}/develop/*"
+          "Resource" : [
+            "${module.protected_binary_mirror.bucket_arn}/develop/*",
+            module.pr_binary_mirror.bucket_arn,
+            "${module.pr_binary_mirror.bucket_arn}/*"
+          ]
         }
       ]
```
mvandenburgh, line 29: I think these can be scoped down - the pruning job doesn't need `PutObject` permissions on the PR bucket, right?

So something more exact like:
```
{
  "Effect" : "Allow",
  "Action" : ["s3:PutObject", "s3:DeleteObject"],
  "Resource" : "${module.protected_binary_mirror.bucket_arn}/develop/*"
},
{
  "Effect" : "Allow",
  "Action" : "s3:ListBucket",
  "Resource" : module.pr_binary_mirror.bucket_arn
},
{
  "Effect" : "Allow",
  "Action" : "s3:DeleteObject",
  "Resource" : "${module.pr_binary_mirror.bucket_arn}/*"
}
```

mvandenburgh (changes requested):

mvandenburgh (approved):

PR #1439, opened 2026-08-14
Encode production GitLab `ApplicationSettings` in Terraform

mvandenburgh:
Adds a gitlab_application_settings resource to the spack_gitlab module with all instance settings (at least, those supported by the current GitLab TF provider) captured from production GitLab.

This PR is intentionally verbose and includes all application settings that exist in our database that are supported by the GitLab TF provider. The idea is to capture these now, use them to stand up the new staging instance, and then we can pare away the settings we don't care about tracking in Terraform.


mvandenburgh:
Note, some of the settings that are really important for spack, `max_yaml_size_bytes` and `max_yaml_depth`, aren't supported by the Terraform provider. I made [an MR](https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/merge_requests/3231) to add them, and it was merged and will be released in the next version. So, I'll make an issue to remember to set those when it's out (looks like it could be next Thursday)

terraform/modules/spack_gitlab/application_settings.tf
```diff
@@ -0,0 +1,721 @@
...
+
+  # Indicates whether the GitLab Duo features enabled setting is enforced for all subgroups.
+  lock_duo_features_enabled = false
+
+  # Enable Mailgun event receiver.
+  mailgun_events_enabled = false
+
+  # When instance is in maintenance mode, non-administrative users can sign in with read-only access and make read-only API requests.
+  maintenance_mode = false
+
```
jjnesbitt, line 376:
```suggestion

maintenance_mode_message = "Gitlab is undergoing scheduled maintenance. If you are experiencing issues, please try again later."

```

jjnesbitt:
FYI @mvandenburgh I was getting an error that seemed to be due to new resources. I bumped the gitlab provider version to match our prod gitlab version.

jjnesbitt (approved):

PR #1441, opened 2026-08-14
Bundle UO runner configs

jjnesbitt:
Built on top of #1420

@eugeneswalker Let me know if this is useful for you. I bundled the core runner configs so that all of the commonality is templated out in a new `generate.py` file, and specific configuration is placed in `fleet.toml`. So all you'd need to do to update your configs is pull down `main` and run `python generate.py`.

I placed the existing configs you committed into a `reference/` sub-directory.


eugeneswalker:
This is great! Thank you!

PR #1456, opened 2026-08-26
Terraform improvements

mvandenburgh:
Various improvements to terraform structure/resources/dependency resolution that we came up with while recreating the staging cluster.


terraform/modules/spack_aws_k8s/aws_load_balancer_controller.tf
```diff
@@ -309,7 +309,7 @@ resource "helm_release" "aws_load_balancer_controller" {
       requests:
         cpu: 100m
         memory: 128Mi
-    enableCertManager: true
+    enableCertManager: false  # We're using AWS ACM for TLS
```
jjnesbitt: We should set this to false, but the reason has nothing to do with TLS traffic _into_ the cluster. This setting controls how the cert is provisioned for internal cluster use with [admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks). Admission webhooks must be served over TLS.

I'm not deeply familiar with the concept, and I don't think we (or anyone reading this file) need to be, so I think no comment here is fine.


```suggestion
    enableCertManager: false
```

terraform/modules/spack_aws_k8s/gitlab_db.tf
```diff
@@ -79,7 +81,7 @@ data "aws_kms_alias" "secretsmanager" {
 }
 
 resource "aws_secretsmanager_secret" "gitlab_db" {
-  name        = "spack-gitlab${local.suffix}-db-proxy-credentials"
+  name        = "spack-gitlab${local.suffix}-db-proxy-credentials-v2"
```
jjnesbitt: Why is this changed?

terraform/modules/spack_aws_k8s/aws_load_balancer_controller.tf
mvandenburgh: Removed

terraform/modules/spack_aws_k8s/gitlab_db.tf
mvandenburgh: Secrets Manager secrets can't be deleted immediately; they instead stick around for a mandatory 30-day period during which they can be recovered. So attempting to recreate it causes a collision.

Although now that I'm thinking about it, we could also just update the existing secret. That probably makes more sense. Terraform wanted to recreate it, but we can do it manually and adjust state accordingly.

jjnesbitt (approved):

PR #1457, opened 2026-08-26
Prune the public buildcache more frequently

zackgalbreath:
Update our pruning cron to run every other day, rather than once a week


jjnesbitt (approved):

jjnesbitt:
Flux is still disabled in prod but I've manually applied this.

PR #1464, opened 2026-09-04
Deploy the analytics stack to staging

mvandenburgh:
Provisions the staging analytics RDS instance (with a cheaper `db.t4g.small` instance type) and deploys the webhook-handler (web + celery worker) to the staging cluster as a kustomize overlay of the production manifests, scaled down to single replicas with small resource requests. The RDS instance adds approximately $23/month to our monthly costs, and the two additional pods for the webhook-handler should add virtually zero costs as they will fit on existing nodes.

This gives staging an end-to-end CI-analytics path, enabling development/testing of the new OpenSearch setup without affecting production (since the webhook-handler will likely be involved with GitLab job log ingestion).

### DIFF
--- a/k8s/staging/custom/webhook-handler/kustomization.yaml
+++ b/k8s/staging/custom/webhook-handler/kustomization.yaml
@@ -1,0 +1,46 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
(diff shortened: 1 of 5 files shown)


jjnesbitt (approved):
Nothing stands out to me. It cleaning up the terraform a bit is a nice side effect!

PR #1473, opened 2026-09-18
updates to terraform SES configuration

zackgalbreath:
* Enable [Virtual Deliverability Manager](https://docs.aws.amazon.com/ses/latest/dg/vdm.html) to give us more insights into the emails we send
* Restrict email sending to instances inside our VPC


mvandenburgh (approved):
This is applied to both prod and staging as of last week, I just forgot to merge this PR

